PR #314, opened 2021-02-19
Import bma

jeswanke:


frontend/src/components/BulkActionModel.tsx
```diff
@@ -180,37 +181,44 @@ export function BulkActionModel<T = unknown>(props: IBulkActionModelProps<T> | {
...
-                                                  let validError = true
-                                                  if (props.isValidError) {
-                                                      validError = props.isValidError(promiseResult.reason)
-                                                  }
-                                                  if (validError) {
-                                                      errors.push({
-                                                          item: props.resources[index],
-                                                          error: promiseResult.reason,
-                                                      })
+                                          if (props.customFn) {
```
showeimer, line 191: I'm not sure that this component should support custom functions; seems like it defeats the purpose of using it in the first place?  Is it possible to just leverage the existing code?

frontend/src/lib/bare-metal-assets.ts
```diff
@@ -40,35 +40,7 @@ export async function syncBMAs(hosts: JsonArray, resources: JsonArray) {
         }
```
showeimer, line 40: There's a lot of code in this file that is not strongly typed.  When I load this branch locally, in vs code I see a lot of errors about types.  We are in the process of working on improving the linting; but we decided to use TypeScript and not JavaScript for a reason, so I would ask that we conform to the TypeScript way of doing things.

frontend/src/lib/bare-metal-assets.ts
```diff
@@ -133,3 +105,101 @@ export async function attachBMAs(assets: JsonArray, hosts: JsonArray, clusterNam
...
+                        }
+                    }
+                }, 100)
+            }
+        }
+        input.click()
+    })
+}
+
+export async function createBMAs(bmas: JsonArray, assets = [], errors = []) {
```
showeimer, line 172: JsonArray is not a valid type

frontend/src/lib/bare-metal-assets.ts
```diff
@@ -133,3 +105,101 @@ export async function attachBMAs(assets: JsonArray, hosts: JsonArray, clusterNam
...
+    let results
+    let response
+    // determine the namespaces required and create them
+    const namespaces = Object.keys(keyBy(bmas, 'namespace'))
```
showeimer, line 176: We don't feel like using lodash functions is necessary because modern Javascript has most of the lodash features now.  Lodash also frequently arises in security scans.  I think temptifly is the only reason lodash is even in this repo now?

frontend/src/lib/bare-metal-assets.ts
```diff
@@ -133,3 +105,101 @@ export async function attachBMAs(assets: JsonArray, hosts: JsonArray, clusterNam
...
+        input.onchange = (e) => {
+            input.remove()
+            const file = e.target.files[0]
+            const reader = new FileReader()
+            reader.readAsText(file, 'UTF-8')
+            reader.onload = async (readerEvent) => {
+                const content = readerEvent.target.result
+
+                // parse csv
+                setTimeout(() => {
```
showeimer, line 127: How come a setTimeout is needed?

showeimer (changes requested):

frontend/src/lib/bare-metal-assets.ts
```diff
@@ -90,7 +63,7 @@ export async function syncBMAs(hosts: JsonArray, resources: JsonArray) {
             const namespace = get(secret, 'value.metadata.namespace')
             return `${name}-${namespace}`
         })
-        const setSecrets = (fhosts) => {
+        const setSecrets = (fhosts: any[]) => {
```
showeimer, line 66: Setting to any is pretty much like not setting a type at all; you should type the structure of `fhosts`

frontend/src/lib/bare-metal-assets.ts
```diff
@@ -112,7 +85,7 @@ export async function syncBMAs(hosts: JsonArray, resources: JsonArray) {
     return { assets, errors }
 }
 
-export async function attachBMAs(assets: JsonArray, hosts: JsonArray, clusterName: string, errors: JsonArray) {
+export async function attachBMAs(assets: any[], hosts: any[], clusterName: string, errors: any[]) {
```
showeimer, line 88: Same comment here; we shouldn't be using `any`, you should type the structure of `assets` and `hosts`, and `errors`.  There's probably already types for these if you didn't transform the data into something else.

frontend/src/components/BulkActionModel.tsx
jeswanke: removed the customFn, but I still needed a preActionFn to run before looping through the actionFn in order to create the namespace that the bare metal assets use--we could create that namespace in each actionFn but it wouldn't be performant if there are a hundred assets imported that only needed one namespace to be created

frontend/src/lib/bare-metal-assets.ts
jeswanke: fixed the typing

frontend/src/lib/bare-metal-assets.ts
jeswanke: modern js has a spec that adds lodash functions like map, etc (which babel/typscript lib actually add), however there is no modern js spec for keyBy

also the way we include lodash, we're only pulling in that function, not the whole lodash library

also it looks like James has a function that keeps lodash up to the current version

frontend/src/lib/bare-metal-assets.ts
jeswanke: replaced it with   
  await new Promise((r) => setTimeout(r, 5000));        
which I see is used in other places in console to pause the next functionality to let the ui thread finish

frontend/src/lib/bare-metal-assets.ts
jeswanke: fixed

frontend/src/lib/bare-metal-assets.ts
jeswanke: fixed

frontend/src/lib/bare-metal-assets.ts
jeswanke: fixed

jeswanke: fixed

PR #323, opened 2021-02-23
PolicyReport drawer component in cluster details page

zlayne:
fyi @RobDolares 
Related issue: https://github.com/open-cluster-management/backlog/issues/8651

Screen Shot 2021-03-05 at 3 11 07 PM


frontend/src/lib/search.ts
```diff
@@ -51,3 +52,21 @@ export function queryStatusCount(cluster: string): IRequestResult<ISearchResult>
...
+
+export function queryCCXReports(cluster: string): IRequestResult<ISearchResult> {
+    return postRequest<SearchQuery, ISearchResult>(backendUrl + apiSearchUrl, {
+        operationName: 'searchResult',
+        variables: {
+            input: [
+                {
+                    filters: [
+                        { property: 'kind', values: ['policyreport'] },
+                        { property: 'namespace', values: [cluster] },
```
showeimer, line 64: The policy report is a hub side resource?

zlayne: Yes as of now we are creating a PolicyReport CR on the hub for every violation per cluster.

showeimer: Cool, if that model stays the same, I think we should not use search then; it would be better to just make kube calls.

jamestalton (approved):
lgtm

PR #325, opened 2021-02-23
Cahl azure regions

chrisahl:
Needs to for in main/2.3 but also will be for issue https://github.com/open-cluster-management/backlog/issues/9700 and 2.2


jamestalton (approved):
lgtm

PR #340, opened 2021-03-05
Server side events

jamestalton:


chrisahl (approved):

backend/src/lib/server-side-events.ts
```diff
@@ -0,0 +1,277 @@
...
+            events,
+            namespaces,
+            writableStream,
+            compressionStream,
+            eventQueue: [],
+        }
+        this.clients[clientID] = eventClient
+
+        res.setTimeout(2147483647)
+        req.setTimeout(2147483647)
```
hanqiuzh, line 209: super large timeout. I assume it's for sse & http2? do we need to modify the ingress? 

I remember for websocket if 30 seconds no messages, will force close the connection. Not sure if the SSE having the problem if we are not polling anymore?

hanqiuzh: nvm, I saw the keep alive, but seems it's 60 seconds?

PR #343, opened 2021-03-08
Fix capitalization error in import command

rh-dluong:
Addresses:
https://bugzilla.redhat.com/show_bug.cgi?id=1936642

Should be --decode.  Mac's base64 uses -Dd and RHEL base64 uses -d.
Changed to -d throughout the whole command to be OS agnostic like how it is earlier in the import command.

### DIFF
--- a/frontend/src/routes/ClusterManagement/Clusters/components/ImportCommand.tsx
+++ b/frontend/src/routes/ClusterManagement/Clusters/components/ImportCommand.tsx
@@ -165,7 +165,7 @@ export async function pollImportYamlSecret(clusterName: string): Promise<string>
                 const alreadyImported = i18next.t('cluster:import.command.alreadyimported')
                 const alreadyImported64 = Buffer.from(alreadyImported).toString('base64')
                 resolve(
-                    `echo ${klusterletCRD} | base64 --decode | kubectl create -f - || test $? -eq 0 && sleep 2 && echo ${importYaml} | base64 --decode | kubectl apply -f - || echo "${alreadyImported64}" | base64 -D`
+                    `echo ${klusterletCRD} | base64 -d | kubectl create -f - || test $? -eq 0 && sleep 2 && echo ${importYaml} | base64 -d | kubectl apply -f - || echo "${alreadyImported64}" | base64 -d`
                 )
             })
             .catch((err) => {


jamestalton:
great catch. i've been bitten by the fact that different os's handle base64 differently
unfortunately there are even more versions out there...

```
docker run --rm alpine base64 --decode
base64: unrecognized option: decode
BusyBox v1.32.1 () multi-call binary.

Usage: base64 [-d] [FILE]

Base64 encode or decode FILE to standard output
        -d      Decode data
```

rh-dluong:
Huh, strange.  You'd think they'd leave the full flag on there at least.  I still think `--decode` is the best way to go since `-d` only definitely leaves out Mac users and `-Dd` breaks everyone else.  Also, we use decode earlier on in that command so that would've never worked in Alpiine it sems.  And there's no way, AFAIK, to know what OS the user is using.  

The only other option would be to give users 2 options like:

For Linux clients, please use the following command to import your cluster:

`Option with base64 -d` # As far as I know, -d will work on all the linux distros I've seen thus far, but I could be mistaken

For Mac clients, please use the following command to import your cluster:

`Option with base64 -Dd instead`

rspagnola:
looking at the base64 --help on Big Sur it doesn't list -d instead it shows -Dd

`base64 --help
Usage:	base64 [-hvDd] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -Dd, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
`

jamestalton:
I believe that `base64 -Dd` means it supports decoding with either `base64 -D` or `base64 -d`
I tested on osx with `base64 -d` and it worked for me.

rspagnola:
Thanks James, I looked into that and found the same and deleted by comment.
Thanks for working on this.


Ryan Spagnola, RHCE

STSE Customer Experience & Engagement

Red Hat  <https://www.redhat.com/>
<https://red.ht/sig>

rh-dluong:
I've adjusted and changed all the `--decode` in that particular command string to `-d` to ensure the whole thing is compatible.

jamestalton:
This bug came in related to this - tagging here
https://github.com/open-cluster-management/backlog/issues/10278

jamestalton (approved):
lgtm

PR #350, opened 2021-03-09
Server stream for Bare Metal Asset Table

Randy424:


frontend/src/routes/BareMetalAssets/BareMetalAsset.test.tsx
```diff
@@ -139,46 +146,42 @@ describe('bare metal asset page', () => {
...
-        await waitFor(() => expect(getAllByText(mockBareMetalAssets[0].metadata.name!).length > 0)) // check for asset in doc
-        userEvent.click(getAllByLabelText('Actions')[0])
+        await waitForText(mockBareMetalAssets[0].metadata!.name!)
+        await clickByLabel('Actions', 0) // Click the action button on the first table row
         await waitFor(() => expect(nocksAreDone(rbacNocks)).toBeTruthy())
-        userEvent.click(getByText('bareMetalAsset.rowAction.deleteAsset.title')) // click the delete action
-        expect(getByText('common:delete')).toBeInTheDocument()
-        userEvent.click(getByText('common:delete')) // click confirm on the delete dialog
-        await waitFor(() => expect(deleteNock.isDone()).toBeTruthy()) // expect the delete api call to finish
-        expect(queryByText('test-bare-metal-asset-1')).toBeNull() // expect asset to no longer exist in doc
```
Randy424, line 157: I noticed that in the providerConnection code, assertions confirming the result of a delete were removed following the stream api change. I adopted that approach tentatively, to align my code with the changes that have already been established/PR'd. I do wonder if this is not a step backwards, as far as thorough testing goes.

Randy424: Also, I recognize that streaming does not seem to be totally compatible with our nock method. I would guess that's why this change was made in the first place.

Randy424:
Treating this as a **_work in progress_** until we get approval for wider use of stream api.

PR #358, opened 2021-03-10
Cahl disable travis use prow (#355)

chrisahl:
* move build, sonar, manifest update and ff to prow

* Doc husky pre commit hook

* Remove Doc husky pre commit hook - James changed function so it works without this


jamestalton (approved):
lgtm

PR #398, opened 2021-03-19
Clusters page - surface managed cluster addons status

showeimer:
Screen Shot 2021-03-19 at 10 33 49 AM


frontend/src/lib/get-cluster.ts
```diff
@@ -386,7 +399,14 @@ export function getClusterStatus(
                 activeCsr && !activeCsr?.status?.certificate ? ClusterStatus.needsapproval : ClusterStatus.pendingimport
         }
     } else {
-        mcStatus = clusterAvailable ? ClusterStatus.ready : ClusterStatus.offline
+        if (clusterAvailable) {
+            const hasDegradedAddons = !!managedClusterAddOns?.some((mca) =>
+                checkForCondition(AddonStatus.Degraded, mca.status?.conditions!)
```
leena-jawale, line 404: will this check all addon status and return true if any one of them has Degraded ?

showeimer: Yeah the `.some` in JS will return true if anything in the array returns true based on the condition

leena-jawale:
looks good. Adding lgtm. One suggestion, we have unit tests only around readystatus. we should consider adding some tests for other cluster statuses.

PR #409, opened 2021-03-22
Some bugs

jeswanke:


frontend/src/resources/resource.ts
```diff
@@ -77,7 +77,7 @@ export function getResourceApiPath(options: {
         path = join(path, options.kind.toLowerCase() + 's')
     }
 
-    return path
+    return path.replace(/\\/g, '/')
```
jamestalton, line 80: When would we have a case with paths that have "\".  Seems like that code should be fixed and not patched here.

jeswanke: I'll back it out--I only need it on Windows

jamestalton (approved):
lgtm

PR #412, opened 2021-03-22
add prow badge to README

chrisahl:


itdove (approved):
We will need to add these badges in each project ;-)

PR #421, opened 2021-03-23
Add open-cluster-management label to featuregate example

zkayyali812:
Ensure example of discovery feature gate in docs has label shown here - 
https://github.com/open-cluster-management/console/blob/main/backend/src/routes/watch.ts#L80

### DIFF
--- a/docs/featuregate.yaml
+++ b/docs/featuregate.yaml
@@ -6,6 +6,6 @@ kind: FeatureGate
 metadata:
   name: open-cluster-management-discovery
   labels:
-    open-cluster-management: ''
+    console.open-cluster-management.io: ''
 spec:
   featureSet: DiscoveryEnabled
--- a/docs/featuregate.yaml
+++ b/docs/featuregate.yaml
@@ -5,5 +5,7 @@ apiVersion: config.openshift.io/v1
 kind: FeatureGate
 metadata:
   name: open-cluster-management-discovery
+  labels:
+    open-cluster-management: ''
 spec:
   featureSet: DiscoveryEnabled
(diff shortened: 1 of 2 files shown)


docs/featuregate.yaml
```diff
@@ -5,5 +5,7 @@ apiVersion: config.openshift.io/v1
 kind: FeatureGate
 metadata:
   name: open-cluster-management-discovery
+  labels:
+    open-cluster-management: ''
```
showeimer, line 9: I think the label would be better as something like:

`console.open-cluster-management.io`

zkayyali812: Should the label be updated here as well? - https://github.com/open-cluster-management/console/blob/main/backend/src/routes/watch.ts#L80

showeimer: Yeah, but does the FG created currently for discovery add the label?  I know you guys are trying to GA discovery, so the FG won't be needed anyways

zkayyali812: Well the user would create it manually when enabling discovery as of now. Since the label in the `watch` is a new addition, we would like to get our documented feature gate up to speed first.

But you are right in the long term we do plan to get rid of the feature gate. This PR will also help ensure our examples are up to date to help with doc/QE validation in the meantime.

showeimer: @zkayyali812 sounds good, let's make the update then.  Thanks for contributing these updates

zkayyali812: @showeimer Np! I just pushed an update to contain this suggestion

showeimer (approved):

PR #430, opened 2021-03-26
Update adding/removing cluster to/from cluster set rbac

showeimer:
APIs have a webhook that is checking for a different set of permissions than just simply patching labels:

https://github.com/open-cluster-management/registration/blob/08d8569c94bedfbe355e4d8a3930fce1bcccac8b/pkg/webhook/cluster/validating_webhook.go#L264-L269

### DIFF
--- a/frontend/src/lib/rbac-util.ts
+++ b/frontend/src/lib/rbac-util.ts
@@ -105,17 +105,27 @@ export async function checkAdminAccess() {
 }
 
 type Verb = 'get' | 'patch' | 'create' | 'delete' | 'update'
+type SubResource = 'join'
 
-export function rbacResource(verb: Verb, resource: IResource, namespace?: string, name?: string) {
+export function rbacResource(
+    verb: Verb,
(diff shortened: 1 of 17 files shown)


frontend/src/lib/rbac-util.ts
```diff
@@ -105,17 +105,27 @@ export async function checkAdminAccess() {
 }
 
 type Verb = 'get' | 'patch' | 'create' | 'delete' | 'update'
-
-export function rbacResource(verb: Verb, resource: IResource, namespace?: string, name?: string) {
+type SubResource = 'join'
```
hanqiuzh, line 108: we will only need join?

showeimer: Currently this is the only subresource I know of; and that we currently will use I guess.  Are there more?

hanqiuzh: no, just feel this code is a little bit confusing, but I'm good with that 👍

PR #435, opened 2021-03-31
Bulk add/remove clusters to cluster set

showeimer:
Issue:  https://github.com/open-cluster-management/backlog/issues/10930


Randy424 (approved):

Randy424:
@showeimer Hey I did encounter this one bug during my review, where on the create cluster set page, an error from validation for the name produces two alerts, one at the top of the page and one at the bottom. I was going to open an issue for it but wondered if it was the right move, given that the issue seems pretty small and cluster sets is still in it's early stages of being developed (so let me know if you want me to open one, if only to keep track of the problem as you continue your work). Here's a screen shot:
Screen Shot 2021-03-29 at 2 53 59 PM

PR #463, opened 2021-04-12
spacing fix

Randy424:
regarding: https://github.com/open-cluster-management/backlog/issues/11398

Screen Shot 2021-04-12 at 5 22 41 PM

Screen Shot 2021-04-12 at 5 22 36 PM


showeimer (changes requested):
I don't like how large those buttons are 🤣

Randy424:
James was lecturing me on how we want to move away from explicitly defining styles and move towards relying on build in Pattern Fly styling...the reason one of these layouts was tighter than the other was because of explicit styling, and I though it would be regressive to just copy that styling after what James...Not sure how to constrain the button size without some explicit styling in this scenario though...so I guess I'll make an exception?

Screen Shot 2021-04-12 at 5 33 45 PM

Screen Shot 2021-04-12 at 5 35 28 PM

jamestalton (approved):
lgtm

PR #471, opened 2021-04-14
Fix machine pool status null pointer bug, remove launch console from cluster actions, change node links to anchor elements, update ui-components

showeimer:


jamestalton (approved):
lgtm

PR #472, opened 2021-04-14
Import claimed cluster,  disable console launch link while hibernating, add claim cluster button to pool table

showeimer:


jamestalton (approved):
lgtm

PR #490, opened 2021-04-20
make default flavor m1.xlarge so openshift has enough memory on opens…

chrisahl:
…tack

Signed-off-by: Chris Ahl <cahl@redhat.com>


jamestalton (approved):
lgtm

PR #497, opened 2021-04-24
Cluster pool updates

showeimer:
- Use ClusterClaim resource name as display name for cluster
- Add grey description with cluster's real resource name to cluster details page title (if claimed cluster)
- Add Name row to cluster details description list that indicates cluster's real resource name
  - Add tooltip to name row explaning this is the resource name for kube resources, and explain display name is different
- Update references to cluster name to use the display name
- Align closer to design, move columns, change fraction reference to sentence, borderless sub tables
- Remove unclaimed cluster table from cluster pool row expand
- Remove links to cluster details from unclaimed/available table
- Change update release image modal to only display ClusterImageSets with label `visible='true'`
- Remove lifetime feature
- Filter out unclaimed clusters from the clusters page


jamestalton (approved):
lgtm

PR #502, opened 2021-04-26
Cluster set updates

showeimer:
- Remove add/remove menu actions from clusters page
- Change Manage clusters page to Manage resource assignments
  - Add ability to add cluster pool
- Break cluster set details into Overview, Clusters, Cluster pools, Access management tabs
- Change create cluster set to a modal


jamestalton (approved):
lgtm

PR #518, opened 2021-04-28
harden server side events

jamestalton:


KevinFCormier (approved):
Looks good to me. What about the access of `writableStream` in the `dispose()` function. Is that one safe?

PR #547, opened 2021-05-03
Nls - product names

chrisahl:
For issue https://github.com/open-cluster-management/backlog/issues/12132


chrisahl:
Using a different PR

PR #558, opened 2021-05-05
change edit credential to patch operation

jamestalton:

### DIFF
--- a/frontend/src/routes/Credentials/CredentialsForm.tsx
+++ b/frontend/src/routes/Credentials/CredentialsForm.tsx
@@ -21,7 +21,7 @@ import { ErrorPage } from '../../components/ErrorPage'
 import { LoadingPage } from '../../components/LoadingPage'
 import { DOC_LINKS } from '../../lib/doc-util'
 import { getAuthorizedNamespaces, rbacCreate } from '../../lib/rbac-util'
-import { createResource, replaceResource } from '../../lib/resource-request'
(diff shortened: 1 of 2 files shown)


jnpacker:
Issue Ref: https://github.com/open-cluster-management/backlog/issues/12196

frontend/src/routes/Credentials/CredentialsForm.tsx
```diff
@@ -936,15 +936,19 @@ export function CredentialsForm(props: {
         ],
         submit: () => {
             if (isEditing) {
-                return replaceResource(stateToData() as IResource).promise.then(async () => {
-                    if (process.env.NODE_ENV === 'development')
-                        await new Promise((resolve) => setTimeout(resolve, 4000))
+                const secret = stateToData() as Secret
+                const patch: { op: 'replace'; path: string; value: unknown }[] = []
```
jnpacker, line 940: @jamestalton I think we need a little more code, there is a key in Spec.Data.`credential-hash` we need to include that from the current Spec.Data to the new Data set. Then update or patch works.

I think this flow will still remove the `credential-hash` because you overwrite (with patch) the whole data field.
Am I missing where the secret Spec.Data is aggregated?

jamestalton: we will have to handle `credential-hash`

PR #595, opened 2021-05-14
Auto fill server

cameronmwall:


frontend/src/routes/ClusterManagement/Clusters/ImportCluster/ImportCluster.tsx
```diff
@@ -210,6 +210,12 @@ export function ImportClusterPageContent() {
                     <AcmSubmit
                         id="submit"
                         variant="primary"
+                        // isDisabled={
```
zkayyali812, line 213: Why Is this commented out? Should it be removed?

cameronmwall: These should not have been commented out. Was left in there from fixing unit tests. Has been corrected

swopebe (approved):
I may be missing it, but I don't see any labels or messages to review so I will approve.

jamestalton (approved):
lgtm

PR #603, opened 2021-05-17
Cluster set - submariner features

showeimer:
Also fixes:

-- Add-ons not showing up on add-on page if no ClusterManagementAddOn exists
-- Resets state of cluster claim modal after claiming a cluster
-- Removes unclaimed clusters from cluster set clusters page
-- Launch out icon on nodes page to be on left for alignment
-- Vertically centers eye icon for cluster credentials button


jamestalton (approved):
lgtm

PR #617, opened 2021-05-20
Increase import secret timeout, remove clean up logic

showeimer:


hanqiuzh:
I think the change make sense for manual import, but will it cause any regression for not deleting?

PR #625, opened 2021-05-21
Update Curator status checks

jnpacker:
Signed-off-by: Joshua Packer <jpacker@redhat.com>

* Set default status as pending
* Detect curator-job failed, so that we still check for sub condition failures
* Check the hive job status only when activate-and-monitor job is present.


jnpacker:
Issue Ref: https://github.com/open-cluster-management/backlog/issues/12604

showeimer (approved):

PR #646, opened 2021-05-26
rename credentials fields

jamestalton:


leena-jawale:
There are these 3 fields in vmware need to change

```
"vcenter":              "vCenter",
"vmClusterName":        "cluster",
"datastore":            "defaultDatastore",
```

And also I am not seeing `osServicePrincipal.json` in PR

jeswanke (approved):

frontend/src/resources/provider-connection.ts
```diff
@@ -17,25 +17,22 @@ export const ProviderConnectionDefinition: IResourceDefinition = {
...
-    awsSecretAccessKeyID?: string
+    aws_access_key_id?: string
+    aws_secret_access_key?: string
 
     baseDomainResourceGroupName?: string
-    clientId?: string
-    clientSecret?: string
-    subscriptionId?: string
-    tenantId?: string
+    ['osServicePrincipal.json']?: string
```
chrisahl, line 24: What is this [' '] notation for?


Other general comment is I would prefer to keep prefixes on variables to make it easier to know which provider they belong to.  Plus it seems to easy to have name collision with this change.

frontend/public/locales/en/create.json
```diff
@@ -226,9 +226,9 @@
     "creation.ocp.network.interface": "Provisioning network interface",
     "creation.ocp.network.bridge": "Provisioning network bridge",
     "creation.ocp.external.bridge": "External network bridge",
-    "creation.ocp.vcenter": "vCenter",
+    "creation.ocp.vCenter": "vCenter",
     "creation.ocp.datacenter": "Datacenter",
-    "creation.ocp.datastore": "Default datastore",
+    "creation.ocp.defaultDatastore": "Default defaultDatastore",
```
chrisahl, line 231: Shouldn't the text stay `Default datastore`?

frontend/src/resources/provider-connection.ts
jamestalton: Any keys that have a '.' in them need to be quoted.
Renaming the keys to match the names used by hive.

chrisahl (approved):

PR #661, opened 2021-06-02
Use SectionGroups for Submariner installer

showeimer:
- Also fixes SubmarinerConfig name to `submariner`
- Fixes logic that prevented vSphere from showing a configuration page.


jamestalton (approved):
lgtm

PR #670, opened 2021-06-03
Support multi brower tab update streams

jamestalton:

### DIFF
--- a/backend/src/lib/server-side-events.ts
+++ b/backend/src/lib/server-side-events.ts
@@ -29,6 +29,7 @@ export interface ServerSideEvent<DataT = unknown> {
 }
 
 export interface ServerSideEventClient {
+    token: string
     events?: Record<string, boolean>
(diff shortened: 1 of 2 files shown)


backend/src/lib/server-side-events.ts
```diff
@@ -226,12 +227,14 @@ export class ServerSideEvents {
...
         }
         const eventClient: ServerSideEventClient = {
+            token,
             events,
             namespaces,
             writableStream,
             compressionStream,
             eventQueue: [],
         }
+        const clientID = randomString(8)
```
hanqiuzh, line 237: does it mean length with 8? not sure about the possibility of collision, or what about random+token/hash-of-token?

hanqiuzh: I guess 8 should be enough for alphanumeric. it's good enough 👍

hanqiuzh (approved):

PR #688, opened 2021-06-07
gafana links

jamestalton:


frontend/src/lib/get-addons.ts
```diff
@@ -147,17 +155,6 @@ function getLaunchLink(cma: ClusterManagementAddOn, mcas: ManagedClusterAddOn[])
...
-        if (cmaHasLaunchLink) {
-            return {
-                displayText: cma?.metadata?.annotations?.[textKey] ?? '',
-                href: cma?.metadata?.annotations?.[pathKey] ?? '',
-            }
-        } else {
-            return undefined
         }
     }
 }
```
showeimer, line 160: The code needs to support the use case that in an all clusters context it returns a link from the ClusterManagementAddOn and in a single cluster case it returns the link from the ManagedClusterAddOn.  From this code it appears that the Grafana launch link could appear in the case of an imported *ks cluster which Observability does not support, and would result in a new defect.

showeimer (changes requested):

PR #689, opened 2021-06-07
Fix release image select width and enable batch update

showeimer:
Signed-off-by: Sho Weimer <sweimer@redhat.com>


jamestalton (approved):
lgtm

PR #692, opened 2021-06-08
Fix scaling alert bug caused by Submariner scaling on AWS

showeimer:
Closes:  https://github.com/open-cluster-management/backlog/issues/13117


jamestalton (approved):
lgtm

PR #720, opened 2021-06-16
Feature gated template

Randy424:
Regarding: https://github.com/open-cluster-management/backlog/issues/13387
Template jobs for Scale and Destroy now hidden behind featureGate

Before*
Screen Shot 2021-06-15 at 8 42 41 PM

Screen Shot 2021-06-15 at 8 42 33 PM

After*
Screen Shot 2021-06-15 at 8 42 08 PM

Screen Shot 2021-06-15 at 8 42 22 PM


Randy424:
fyi @jnpacker

features/automation.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright Contributors to the Open Cluster Management project
+
+apiVersion: config.openshift.io/v1
+kind: FeatureGate
+metadata:
+    name: open-cluster-management-automation-template
```
showeimer, line 6: I think we can just drop `open-cluster-management` here because the label kind of already indicates that.  Maybe we can name this `ansible-automation-template`

features/automation.yaml
```diff
@@ -0,0 +1,10 @@
...
+    labels:
+        console.open-cluster-management.io: ""
+spec:
+    featureSet: AutomationStepsAllEnable
```
showeimer, line 10: Let's just value this the same as the resource name

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -92,6 +93,10 @@ export function AnsibleAutomationsForm(props: {
     const { t } = useTranslation(['cluster', 'common', 'credentials', 'create'])
     const { ansibleCredentials, clusterCurator, isEditing, isViewing } = props
 
+    const [featureGates] = useRecoilState(featureGatesState)
+    const automationFeatureGateFlag =
+        featureGates.filter((featureGate) => featureGate.spec?.featureSet === 'AutomationStepsAllEnable').length > 0
+
```
showeimer, line 99: You can just use a `.find` here, I think we can just move it down to where it's referenced, instead of making a new variable that we have to keep track of

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -313,92 +318,96 @@ export function AnsibleAutomationsForm(props: {
...
-                                    setEditAnsibleJob(ansibleJob)
-                                },
-                                onCreate: () => {
-                                    setEditAnsibleJobs({ jobs: destroyPostJobs, setJobs: setDestroyPostJobs })
-                                    setEditAnsibleJob({ name: '', extra_vars: {} })
-                                },
-                            },
-                        ],
-                    },
+                    ...(automationFeatureGateFlag
```
showeimer, line 321: Can `&&` work instead of doing a ternary?

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -313,92 +318,96 @@ export function AnsibleAutomationsForm(props: {
...
+                                          type: 'OrderedItems',
+                                          label: t('template.preScale.label'),
+                                          placeholder: t('template.job.placeholder'),
+                                          value: scalePreJobs,
+                                          onChange: setScalePreJobs,
+                                          keyFn: (ansibleJob: AnsibleJob) => ansibleJob.name,
+                                          cellsFn,
+                                          onEdit: (ansibleJob) => {
+                                              setEditAnsibleJobs({ jobs: scalePreJobs, setJobs: setScalePreJobs })
+                                              setEditAnsibleJob(ansibleJob)
```
showeimer, line 339: Not really related to these changes, but I would find this plural distinction hard to understand..

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
showeimer: and let's reference the resource name too, instead of the featureSet value

showeimer (commented):
I think maybe we should create a top level file that keeps track of the FeatureGates, and make an enum kind of like we do with NavigationPath.

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
Randy424: i think the `...` operator does not respond well to the case `...(undefined)` so we need to be able to return an empty array.

showeimer:
@Randy424 you may have missed my earlier comment, but I think we should keep a list of the feature gates in a file to reference:

https://github.com/open-cluster-management/console/pull/720#pullrequestreview-684635711

Randy424:
> @Randy424 you may have missed my earlier comment, but I think we should keep a list of the feature gates in a file to reference:
> 
> [#720 (review)](https://github.com/open-cluster-management/console/pull/720#pullrequestreview-684635711)

https://github.com/open-cluster-management/console/pull/720/files#diff-d94d5a67d79f95669f0bd58a441a3b1b3481a5f40b3f3ecd5869f1b2074993f9R1-R7

frontend/src/FeatureGates.ts
```diff
@@ -0,0 +1,7 @@
+/* Copyright Contributors to the Open Cluster Management project */
+/* istanbul ignore file */
+
+export enum FeatureGates {
+    clusterSet = 'open-cluster-management-cluster-sets',
```
showeimer, line 5: There's no feature gate for cluster set

showeimer (approved):

PR #747, opened 2021-06-25
Update to ui-components@0.178.0

KevinFCormier:
https://github.com/open-cluster-management/backlog/issues/12617

Note that this PR reverts the `package-lock.json` to lockfileVersion 1. It looks like @jeswanke accidentally used npm 7 to update packages, and I'm worried that this may cause builds using npm 6 to ignore the `package-lock.json`.

image


frontend/src/App.css
```diff
@@ -1,7 +1,3 @@
 .pf-c-toolbar__content-section {
     row-gap: 16px !important;
 }
-
```
chrisahl, line 4: Wondering about this change?

KevinFCormier: Fix is available in version of PF console is using:
https://github.com/patternfly/patternfly/issues/4053

chrisahl (approved):

chrisahl:
For issue https://github.com/open-cluster-management/backlog/issues/12617

PR #750, opened 2021-06-28
Ansible status fix

Randy424:
https://github.com/open-cluster-management/backlog/issues/13374
https://github.com/open-cluster-management/backlog/issues/13402


frontend/src/lib/get-cluster.ts
```diff
@@ -612,6 +634,30 @@ export function getClusterStatus(
             }
 
             return { status: ccStatus, statusMessage }
+        } else if (clusterDeployment) {
```
chrisahl, line 637: Seems to be there is a lot of duplication in the two if blocks and you could rework it to be simpler.  Something like

                if (
                    checkCuratorConditionFailed(CuratorCondition.curatorjob, ccConditions) &&
                    checkCuratorLatestFailedOperation(CuratorCondition.install, ccConditions)
                ) {
                    if (clusterDeployment.spec?.installed) {
                       ccStatus = ClusterStatus.posthookfailed
                    } else {
                       ccStatus = ClusterStatus.posthookfailed
                    }
                    statusMessage = clusterCurator.status?.conditions[0].message

                    return { status: ccStatus, statusMessage }
                }
            }

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ClusterDetails/ClusterDetails.tsx
```diff
@@ -79,6 +84,7 @@ export default function ClusterDetailsPage({ match }: RouteComponentProps<{ id:
             clusterManagementAddonsState,
             clusterClaimsState,
             clusterCuratorsState,
+            ansibleJobState,
```
showeimer, line 87: Why not just read the ansibleJobs from the recoil cache in the DistributionInfo component instead of doing all the set up and context setting and prop passing down in all the other files?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx
```diff
@@ -1,29 +1,105 @@
...
+            >
+                {t('view.logs')}
+            </AcmButton>
+        )
+
+        // if pre/post failed
+        if (checkCuratorLatestFailedOperation(CuratorCondition.upgrade, ccConditions)) {
+            statusType = StatusType.warning
+            console.log('in failure cond')
+            console.log('check: ', upgradeInfo?.prehooks.inProgress)
```
showeimer, line 72: ^

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx
```diff
@@ -1,29 +1,105 @@
...
+export function DistributionField(props: {
+    cluster?: Cluster
+    clusterCurator?: ClusterCurator | undefined
+    ansibleJobs?: AnsibleJob[] | undefined
+}) {
     const { t } = useTranslation(['cluster'])
     const [open, toggleOpen] = useState<boolean>(false)
     const toggle = () => toggleOpen(!open)
+    const clusterCurator = props.clusterCurator
+    const upgradeInfo = props.cluster?.distribution?.upgradeInfo
```
showeimer, line 34: all the other conditional flows just reference the full object path, so we should have consistency with everything

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx
```diff
@@ -1,29 +1,105 @@
...
+                icon={<ExternalLinkAltIcon />}
+                iconPosition="right"
+                isDisabled={!latestAnsibleJob.prehook?.status?.ansibleJobResult?.url}
+            >
+                {t('view.logs')}
+            </AcmButton>
+        )
+
+        // if pre/post failed
+        if (checkCuratorLatestFailedOperation(CuratorCondition.upgrade, ccConditions)) {
```
showeimer, line 69: We still have to do this, when the get-cluster file is attempting to extract the failures?

Randy424: I don't believe we have to, but checking the common case first (if there is a failure in either hook) cuts down on redundant code in the inner conditions (prehook/posthook); statusType and the footerContent will be declared the same way in either case, made sense to do it once then check each hook for the specific stuff.

showeimer (approved):

PR #762, opened 2021-07-01
String fixes

Randy424:

https://github.com/open-cluster-management/backlog/issues/13886

Tooltip change:
Screen Shot 2021-07-01 at 12 50 19 PM

The former string in creation wizard, above the template selector, was also a bit misleading (jobs can be run in creation and upgrade). The old string only mentioned creation.

Screen Shot 2021-07-01 at 12 54 38 PM

There's one last change for the placeholder in Ansible credential creation.


imjoyjean (commented):
thanks for catching that - fix looks good to me

jamestalton (approved):
lgtm

PR #788, opened 2021-07-26
Workspaces

jamestalton:


KevinFCormier (changes requested):
Looks good! Can you please update the README to mention that npm 7 is required? And anything else developers need to know... for example, what is the process to update/install a dependency for a given workspace?

PR #796, opened 2021-08-04
Workspaces + breaking out resources as a library

jamestalton:


Dockerfile.prow
```diff
@@ -1,34 +1,23 @@
...
 RUN mkdir -p /app
 WORKDIR /app
 ENV NODE_ENV production
-COPY --from=backend /app/backend/node_modules ./node_modules
-COPY --from=backend /app/backend/build ./
-COPY --from=frontend /app/frontend/build ./public
+COPY --from=builder /app/node_modules ./node_modules
+COPY --from=builder /app/backend/node_modules ./backend/node_modules
+COPY --from=builder /app/backend/build ./backend
+COPY --from=builder /app/frontend/build ./backend/public
```
KevinFCormier, line 21: The destination path seems a little odd here.

README.md
```diff
@@ -26,6 +26,25 @@ Go to the [Contributing guide](CONTRIBUTING.md) to learn how to get involved.
 ## Prerequisites
 
 - Node.js v14.x
+  - `node -v`
+- NPM 7 (npm 6 ships with node 14)
+  - `npm install --global npm@7`
+  - `npm -v`
```
KevinFCormier, line 32: Can we add a note here, something like:

Caution: Before working with other nodejs 14 projects that are using npm 6, run `npm uninstall --global npm` to revert to npm 6 so that you do not modify the lockfile version in those projects.

KevinFCormier:
@jamestalton Are you seeing errors like this?
image

image

jamestalton:
i am not seeing errors. you have npm 7?

README.md
jamestalton: hmmm. thinking on that one

KevinFCormier:
> i am not seeing errors. you have npm 7?

Yes. Just tried again and no errors ¯\_(ツ)_/¯

README.md
KevinFCormier: Personally I use [asdf](https://asdf-vm.com/) for managing node versions. We could potentially add an `.nvmrc` file, which is compatible with both `nvm` and `asdf`, and point it to nodejs 16. As long as nothing in the build system actually looks at `.nvmrc`.

KevinFCormier (approved):

PR #799, opened 2021-08-06
npm-7-downstream-support

jamestalton:

Basically creating a npm-7 folder with just npm-7 installed locally so that there is a package-lock.json that cachito can reference for installing npm-7.


fxiang1 (approved):
Looks good!

PR #800, opened 2021-08-06
Update credential selection and validation

Randy424:
- Updated the resource-request method "listAnsibleTowerJobs" to properly cast json to AnsibleTowerJobTemplateList type (it was not doing this properly before.
- Made some changes to references of listAnsibleTowerJobs results in AnsibleAutomationForms.tsx, no longer using lodash, now referencing the AnsibleTowerJobTemplateList & AnsibleTowerJobTemplate member vars (sorry if you had trouble with this earlier, it was definitely my fault 😅  )
- Now passing Ansible jobs to EditAnsibleJobModal(), & making api call from AnsibleAutomationsForm(), to allow for validation in the select form. 
- Now using validation with the api call to check the credential.

### DIFF
--- a/backend/src/routes/ansibletower.ts
+++ b/backend/src/routes/ansibletower.ts
@@ -19,8 +19,8 @@ export function ansibleTower(req: Http2ServerRequest, res: Http2ServerResponse):
         chucks.push(chuck)
     })
     req.on('end', () => {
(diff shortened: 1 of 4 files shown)


chenz4027 (approved):
lgtm!

PR #806, opened 2021-08-11
Added podman variants to docker: scripts

jelkosz:
Added podman variants to enable fedora users where docker has been
replaced by podman.


jelkosz:
@rawagner I suspect this might be done more elegantly?

jamestalton (approved):
lgtm

PR #807, opened 2021-08-11
url not found error case

Randy424:

Hey @chenz4027, I found that we actually were missing an error case, and added it to the branch. 
When the url itself is bad, a ENOTFOUND & 404 is returned, but fetchRetry was not catching 404's.


chenz4027 (approved):
lgtm, thanks for the fix!

PR #808, opened 2021-08-11
Changes for our test

Randy424:
Looks like we can just add another job in the result array with better naming, then reference it in the test. 

Here I'm now using "pre-job", and "post-job", instead ob just using "job_1".


chenz4027 (approved):
thanks Randy!

PR #812, opened 2021-08-16
try catch for bad url in backend

Randy424:

regarding: https://github.com/open-cluster-management/backlog/issues/15202


KevinFCormier (approved):
Looks good! This is the type of check I would propose doing in the frontend as well.

PR #813, opened 2021-08-16
Added validation for a basic URL + remove trailing '/' from ansible host input

Randy424:
regarding: https://github.com/open-cluster-management/backlog/issues/15299
https://github.com/open-cluster-management/backlog/issues/15202

### DIFF
--- a/frontend/src/lib/validation.ts
+++ b/frontend/src/lib/validation.ts
@@ -213,7 +213,7 @@ export function validateBareMetalOSImageURL(value: string, t: TFunction) {
     return undefined
 }
 
-export function validateURL(url: string, t: TFunction) {
+export function validateWebURL(url: string, t: TFunction) {
     if (
         validator.isURL(url, {
             require_protocol: true,
--- a/frontend/src/lib/validation.ts
+++ b/frontend/src/lib/validation.ts
@@ -213,7 +213,7 @@ export function validateBareMetalOSImageURL(value: string, t: TFunction) {
 }
 
 export function validateURL(url: string, t: TFunction) {
-    const regex = /^(?:\w+:)?\/\/([^\s.]+\.\S{2}|localhost[:?\d]*)\S*$/
+    const regex = /(http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-/]))\S*$/
     if (!regex.test(url)) {
         return t('validate.ansible.url.not.valid')
     } else {
--- a/frontend/src/lib/validation.ts
+++ b/frontend/src/lib/validation.ts
@@ -213,7 +213,7 @@ export function validateBareMetalOSImageURL(value: string, t: TFunction) {
 }
 
 export function validateURL(url: string, t: TFunction) {
-    const regex = /(http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-/]))\S*$/
+    const regex = /(http|https):\/\/(www.)?(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?((\/|\/)?([\w#!:.?+=&%@!\-/]))\S*$/
     if (!regex.test(url)) {
         return t('validate.ansible.url.not.valid')
     } else {
--- a/frontend/src/lib/validation.ts
+++ b/frontend/src/lib/validation.ts
@@ -3,6 +3,7 @@
 import YAML from 'yaml'
 
 import { TFunction } from 'i18next'
+import validator from 'validator'
 
 const lowercaseAlphaNumericCharacters = 'abcdefghijklmnopqrstuvwxyz1234567890'
 export function validateKubernetesDnsName(value: string, t: TFunction) {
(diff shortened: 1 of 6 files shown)


frontend/src/routes/Credentials/CredentialsForm.tsx
```diff
@@ -250,6 +251,15 @@ export function CredentialsForm(props: {
...
 
+    function removeTrailingChar(url: string, char: string): string {
+        let newUrl = url
+        if (newUrl.charAt(newUrl.length - 1) === char) {
+            newUrl = newUrl.substr(0, newUrl.length - 1)
+            return removeTrailingChar(newUrl, char)
+        }
+        return newUrl
+    }
+
```
Randy424, line 262: When building the request url we assume the credential host does not contain a trailing '/', this function ensures that it is saved without one.

KevinFCormier: We already depend on `lodash` so should we just use https://lodash.com/docs/4.17.15#trimEnd ?

frontend/src/lib/validation.ts
```diff
@@ -211,3 +211,12 @@ export function validateBareMetalOSImageURL(value: string, t: TFunction) {
 
     return undefined
 }
+
+export function validateURL(url: string, t: TFunction) {
+    const regex = /(http|https):\/\/(www.)?(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?((\/|\/)?([\w#!:.?+=&%@!\-/]))\S*$/
```
KevinFCormier, line 216: This regex looks very complex and it seems to be checking more than that it's just a syntactically correct URL. For example, it only allows http or https protocol... so if this is specific to what an Ansible URL should like, let's not use a generic name.

Other concerns with it:
1. Does not begin with `^` - can invalid stuff be inserted at the beginning?
2. Why do we check for optional `www.`? I think we can just leave that out.
3. The character class `\w` includes underscore, which is not valid in the domain part of a URL.

I tried with `!3kfj   http://hello_you.com` and it was accepted.

KevinFCormier: Maybe there's a library we can use to try parsing the URL, then if successful check the protocol and any other restrictions we need to place for an Ansible URL.

KevinFCormier (changes requested):
See comments, please.

frontend/src/routes/Credentials/CredentialsForm.tsx
KevinFCormier: Also, if you parse the URL, note that it allows whitespace before and after, so it would be a good idea to trim whitespace as well. This is handy if someone does a copy-paste and accidentally includes some space.

frontend/src/lib/validation.ts
```diff
@@ -211,3 +211,19 @@ export function validateBareMetalOSImageURL(value: string, t: TFunction) {
...
+        towerUrl = new URL(url)
+    } catch (err) {
+        return t('validate.ansible.url.not.valid')
+    }
+
+    if (towerUrl.protocol.toString() === 'https:' || towerUrl.protocol.toString() === 'http:')
+        if (!regex_injected_spaces.test(url)) return undefined
+
+    return t('validate.ansible.url.not.valid')
+}
```
Randy424, line 229: @KevinFCormier 
So I've kept some regex because ```new URL``` does not error when it encounters spaces within the url string.  It also accepts basically any string for a protocol, as long as it's ended correctly.
This new regex picks up on inner spaces, trailing, and leading spaces-making me wonder if the trimming method I've updated is still necessary?  I'm also explicitly checking the protocol. Thoughts?

KevinFCormier: Hey Randy,

Thanks for looking at this more.

1. I think my advice was bad. The URL parsing doesn't seem to do much validation at all. In addition to spaces, I can put lots of invalid characters in there, like underscores in the domain part, punctuation, etc. I guess we could go back to a regex, or maybe @jamestalton has a suggestion? There are some packages available like https://www.npmjs.com/package/validator that give an isURL function with a bunch of options.
1. On the question of whether the trimming method is needed - depends whether you want to mark spaces or trailing / as invalid, or whether you want to automatically remove them.

frontend/src/routes/Credentials/CredentialsForm.tsx
```diff
@@ -250,6 +252,18 @@ export function CredentialsForm(props: {
     // Red Hat Cloud
     const [ocmAPIToken, setOcmAPIToken] = useState(providerConnection?.stringData?.ocmAPIToken ?? '')
 
+    function removeTrailingChar(url: string, char: string): string {
+        return _.trimEnd(url, char)
+    }
+
+    function removeLeadingTrailingSpaces(url: string): string {
+        return _.trimEnd(_.trimStart(url))
```
KevinFCormier, line 260: You can just use `_.trim` function here.

frontend/src/routes/Credentials/CredentialsForm.tsx
```diff
@@ -250,6 +252,18 @@ export function CredentialsForm(props: {
...
+    }
+
+    function removeLeadingTrailingSpaces(url: string): string {
+        return _.trim(url)
+    }
+
+    function cleanAnsibleUrl(url: string): string {
+        return removeTrailingChar(removeLeadingTrailingSpaces(url), '/')
+    }
+
```
KevinFCormier, line 266: I guess the first 2 functions are just aliases... could just collapse them into `cleanAnsibleUrl`

KevinFCormier (approved):
Thanks, Randy!

PR #815, opened 2021-08-18
scroll fix

Randy424:
regarding: https://github.com/open-cluster-management/backlog/issues/15352

https://user-images.githubusercontent.com/21374229/129972261-eb4fc33a-7862-4aa5-83b6-4abe0fdc6f03.mov

### DIFF
--- a/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
+++ b/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
@@ -543,6 +543,8 @@ function EditAnsibleJobModal(props: {
             <AcmForm>
                 <AcmForm>
                     <AcmSelect
+                        maxHeight="12em"
+                        menuAppendTo="parent"
                         label={t('cluster:template.modal.name.label')}
                         id="job-name"
                         value={ansibleJob?.name}
--- a/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
+++ b/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
@@ -543,7 +543,7 @@ function EditAnsibleJobModal(props: {
             <AcmForm>
                 <AcmForm>
                     <AcmSelect
-                        maxHeight="12em"
+                        maxHeight="24em"
                         menuAppendTo="parent"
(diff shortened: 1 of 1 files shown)


Randy424:
@KevinFCormier 
There's a pattern of using '6em' for select maxHeights in our console repo, at first I was tempted to use this, but it felt a bit small. What do you think?


https://user-images.githubusercontent.com/21374229/129972643-d6c49ab7-aa10-4149-8bb0-b179479957b6.mov

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -543,6 +543,8 @@ function EditAnsibleJobModal(props: {
             <AcmForm>
                 <AcmForm>
                     <AcmSelect
+                        maxHeight="12em"
```
KevinFCormier, line 546: I'm not sure that we need the maximum height here... if the list is long, with the fix it will extend below the modal, making better use of screen space for scrolling through a long list.

Let's check with @imjoyjean though - do you think a dropdown on a modal should try to fit within the modal, or can it extend below as long as it doesn't get hidden inside the modal? ie. compare this img to the video Randy posted:
image

KevinFCormier: Hmm, t seems like if your window is not tall, the dropdown without a maximum height will just get clipped, so we may in fact need to set one.
image

imjoyjean: I think the ideal would be that the max height of the dropdown should be within the borders of the modal, and you can scroll up and down in that fixed height. Let me check with patternfly to see if they have any guidance for it.

KevinFCormier: @Randy424 Joy did some research and is suggesting setting a max height that fits about 10 items. This may overflow the modal, but hopefully not the user's window.

Randy424:
I've extended it to fit ~10 items
Screen Shot 2021-08-19 at 1 01 26 PM

KevinFCormier (approved):

chenz4027:
@Randy424 Hold label is still on

PR #828, opened 2021-09-08
Azure gov UI

Randy424:
Minimal requirements to support Azure Gov in UI

- https://github.com/open-cluster-management/backlog/issues/15759
- https://github.com/open-cluster-management/backlog/issues/15715


jamestalton:
My only concern here is that the region is not filtered for gov/public

KevinFCormier (approved):
This looks great. I agree with James - it would be ideal if the list of regions is dynamic based on the cloud name, but perhaps we can handle that in a subsequent PR.

Randy424:
@KevinFCormier Thanks! My opinion after trying to implement a filter solely on the console side yesterday evening is that the filter requires some dynamic action from both console and temptifly and the current onChange method doesn't meet our needs for the feature. So yeah, it's going to end up in a following PR. 
I am tracking work for that here: https://github.com/open-cluster-management/backlog/issues/15932

PR #832, opened 2021-09-09
managed-clusters-filter

jamestalton:


KevinFCormier (changes requested):
I think the filter options need to be dynamic in this case based on what is present... it's too much in a single dropdown:
image

PR #836, opened 2021-09-09
Set border to 0 instead of hiding with -1 margin

KevinFCormier:


KevinFCormier:
@jamestalton This didn't work. I think it has something to do with the style changes introduced here:
https://github.com/open-cluster-management/console/pull/626/files#diff-0ca255a9a471ade7928c16c1d58ba47b249e1bd5fb69fa284ad57afcddad6c30

Page still failed to display until I disabled the `margin: unset` style in the inspector. I will try reverting these changes.

PR #843, opened 2021-09-15
Region filter feature for Azure

Randy424:

regarding: https://github.com/open-cluster-management/backlog/issues/15932

To preview feature, simply create an Azure credential with a 'AzureUSGovernmentCloud' as the _cloudName_*
Then go to create cluster and and use that Azure gov credential. 
When you arrive at the Master Node step, the regions select will offer Azure-government specific regions. 
When you de-select the Azure government credential or replace it with a public credential, the public regions are swapped back in.

Screen Shot 2021-09-15 at 11 02 33 AM


Randy424:
/cc @jeswanke @KevinFCormier @chenz4027 
Thought I should add @chenz4027 to continue cross pollenating our PR reviews.

chenz4027 (approved):
lgtm! Thanks Randy

PR #846, opened 2021-09-16
Move "On Premises" into Providers section

mareklibra:
Depends on:
- [ ] https://github.com/open-cluster-management/temptifly/pull/136
- [x] https://github.com/openshift-assisted/assisted-ui-lib/pull/805

### DIFF
--- a/frontend/public/locales/en/common.json
+++ b/frontend/public/locales/en/common.json
@@ -42,6 +42,7 @@
     "outOf": "{{firstNumber}} out of {{secondNumber}}",
     "wizard.review.title": "Review your selections",
     "wizard.review.description": "Return to a step to make changes.",
+    "deprecated": "Deprecated",
 
     "yes": "Yes",
     "no": "No",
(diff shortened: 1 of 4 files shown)


mareklibra:
---

matthewcarleton:
awesome, thank you

I can’t tell b/c of the pop-up but is there an (?) on the centrally managed heading?

with this content
Simplify cluster self-service with infrastructure environments, which enable you to manage hosts, create ready-to-use host groups, and quickly create clusters based on provided infrastructure. To use infrastructure environments, select On Premise.

can we also add some padding on the bottom of the card so there’s space below the TP label?

mareklibra:
@matthewcarleton I had to add support for the tooltips but this is the result:

rawagner:
@mareklibra since you're at it, can you please rename the card to `On-premise` ? It is the correct term

mareklibra:
Closing, merged into #851

PR #855, opened 2021-09-18
No longer creating ssh-privatekey secret for cluster-pools

Randy424:
regarding: https://github.com/open-cluster-management/backlog/issues/16134

Removes ssh-privatekey yaml from create resource array in cluster pools create code.


Randy424:
/cc @KevinFCormier
FYI @jnpacker

PR #856, opened 2021-09-20
ACM-910 Add managedClusterSet and additionalLabels to the AI ClusterDetail create wizard

mareklibra:
Depends on:
- [x] https://github.com/openshift-assisted/assisted-ui-lib/pull/816


mareklibra:
Will be included in https://github.com/open-cluster-management/console/pull/865 , so closing here

PR #858, opened 2021-09-20
Safari wizard styling

KevinFCormier:
https://github.com/open-cluster-management/backlog/issues/15925


KevinFCormier:
@jamestalton Safari has issues with `calc` and `100vh`. Surprisingly, it seems like this calculation is not really necessary, and a height of 100% seems to work. If this is good with you, I will backport to 2.3 as well.

jamestalton (commented):
lgtm

jamestalton (approved):

PR #884, opened 2021-09-28
frontend performance queue updates 500ms interval

jamestalton:


frontend/src/atoms.tsx
```diff
@@ -192,63 +192,46 @@ export function LoadData(props: { children?: ReactNode }) {
...
-        async function processEvents() {
-            if (!eventDataQueue) return
-            const dataToProcess = eventDataQueue
-            for (const kind in setters) {
+        function processEventQueue() {
+            if (eventQueue.length === 0) return
+
+            const kindsMap = eventQueue?.reduce((kindsMap, eventData) => {
+                const kind = eventData.object.kind
+                if (kind) {
```
KevinFCormier, line 202: Is kind ever falsy?

KevinFCormier: Also, I guess we don't currently support duplicate kinds in different apiGroups?

frontend/src/atoms.tsx
```diff
@@ -293,7 +275,10 @@ export function LoadData(props: { children?: ReactNode }) {
             }
         }
         startWatch()
+
+        const timeout = setInterval(processEventQueue, 1000)
```
KevinFCormier, line 279: Have you played with different values? Would half a second work? And maybe this should be stored in configuration?

KevinFCormier (commented):
A couple comments.

frontend/src/atoms.tsx
jamestalton: A half sec would work.

frontend/src/atoms.tsx
jamestalton: kind should never be falsy, but sometimes I code defensively.
don't currently support duplicate kinds

PR #896, opened 2021-10-01
Now using patternfly toggleID prop for wizard toggles

Randy424:
regarding: https://github.com/open-cluster-management/backlog/issues/15292

Now using patternfly Toggle ID for wizard toggles (to support more robust testing).

Screen Shot 2021-10-01 at 6 44 36 PM


Randy424:
FYI @dhrpatel4

PR #902, opened 2021-10-05
ACM-16651: BM hosts on BM platform only

mareklibra:


mareklibra:
To fix the CI tests, we need to release new ai-ui-lib first

jtomasek:
Changes in this PR are included in https://github.com/open-cluster-management/console/pull/914 which also updates assisted-ui-lib dependency

PR #929, opened 2021-10-13
Update app navigation

jamestalton:
Signed-off-by: James Talton <jtalton@redhat.com
Screen Shot 2021-10-14 at 9 07 05 AM
>

A big part of this is consolidating route navigation under the different sections.
Example: When we provide "clusters" as a dynamic component to OCP, it needs to be able to handle all of its route navigation.

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/Clusters.test.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/Clusters.test.tsx
@@ -3,11 +3,13 @@
 import { render } from '@testing-library/react'
 import { MemoryRouter } from 'react-router-dom'
 import { RecoilRoot } from 'recoil'
+import { nockIgnoreRBAC } from '../../../lib/nock-util'
 import { waitForText } from '../../../lib/test-util'
 import ClusterManagementPage from './Clusters'
 
 describe('Cluster Management', () => {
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Home/Welcome/Welcome.tsx
```diff
@@ -0,0 +1,158 @@
...
+                title: 'Multicluster networking',
+                description:
+                    'Enable direct networking connection between different on-premises or cloud-hosted Kubernetes clusters by grouping them in cluster sets and enabling the Submariner add-on.',
+                link: NavigationPath.overview,
+            },
+        ],
+        []
+    )
+
+    const details = useMemo(
```
mareklibra, line 45: nit: There is no need for the `useMemo`, the object can be simply one-time created, i.e. in the top-level scope of this file.

mareklibra: that's true if `const details = [...]` is part of the `WelcomePage` function. If created outside of the WelcomePage (on the same), it will be created just once. There is no need to generate the array - it's all static if I am not mistaken.

jamestalton: In general I try to avoid moving variables up to the top-level scope.

PR #936, opened 2021-10-15
Only use CIM workaround if ClusterDeployment exists

KevinFCormier:
https://github.com/open-cluster-management/backlog/issues/17301


jamestalton (approved):
lgtm

PR #960, opened 2021-10-29
Upgrade tar to 6.1.11

KevinFCormier:
https://github.com/open-cluster-management/backlog/issues/16617

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -5,17 +5,17 @@
     "requires": true,
     "dependencies": {
         "@babel/code-frame": {
-            "version": "7.14.5",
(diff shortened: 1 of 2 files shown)


Randy424:
There are a few upgrade happening along side tar, I think an unintended upgrade is causing the test failure. Yaml object is referencing deprecated functions for example. 🤔
``` Function yaml.safeLoad is removed in js-yaml 4. Use yaml.load instead, which is now safe by default.```

KevinFCormier:
Thanks, @Randy424. We had an undeclared direct dependency on js-yaml, so we were just using whatever was provided by our sub-dependencies. Fixed now.

PR #963, opened 2021-11-01
Remove console-api CRUD queries from search pages

zlayne:

Related Issue: https://github.com/open-cluster-management/backlog/issues/17364


frontend/src/resources/managedclusteraction.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        result: actionResponse.status?.result,
+                    })
+                } else if (isComplete === 'Completed' && isActionDone !== 'ActionDone') {
+                    reject({ message: actionMessage })
+                }
+                deleteManagedClusterAction({ name: actionName, namespace: clusterName })
+            })
+            .catch((err) => {
+                if (retries-- > 0) {
+                    setTimeout(poll, 500, resolve, reject)
```
KevinFCormier, line 162: Should we make the poll interval configurable? In console-api, looks like the default was 100, not 500.

zlayne: I can update it to 100. In looking at the console-api deployment we are never overriding the default value.

PR #968, opened 2021-11-08
add policy actions, row and bulk

RobDolares:
Allows selection of a policy's related resources (placementBinding/Rule) for bulk and single policy deletion.


jamestalton (approved):
lgtm

PR #969, opened 2021-11-10
Add create-application route and GitOpsCluster

chenz4027:

- Added Advanced Configuration tab
- Added terminology section for the Advanced Configuration tab
- Added the route for create-application
- Added the method of getting gitopscluster resource and argo namespace from it


frontend/src/atoms.tsx
```diff
@@ -243,7 +243,7 @@ export function LoadData(props: { children?: ReactNode }) {
         addSetter(ChannelApiVersion, ChannelKind, setChannelsState)
         addSetter(PlacementRuleApiVersion, PlacementRuleKind, setPlacementRulesState)
         addSetter(SubscriptionApiVersion, SubscriptionKind, setSubscriptionsState)
-        addSetter('apps.open-cluster-management.io/v1alpha1', 'GitOpsCluster', setGitOpsClustersState)
+        addSetter('apps.open-cluster-management.io/v1beta1', 'GitOpsCluster', setGitOpsClustersState)
```
fxiang1, line 246: @chenz4027 Can you import the apiVersion and kind from gitops-cluster.ts just like the other resources? This way we only need to update the version in one place when it changes

chenz4027: Sure I'll make the change

fxiang1 (approved):
LGTM!

PR #983, opened 2021-11-16
lbFloatingIP addition, Kuryr addition

Randy424:
Regarding: https://github.com/open-cluster-management/backlog/issues/17837, https://github.com/open-cluster-management/backlog/issues/17838


frontend/public/locales/en/create.json
```diff
@@ -255,6 +255,8 @@
     "creation.ocp.otp.instance.flavor": "Instance flavor",
     "creation.ocp.cluster.ost.external.network.name": "External network name",
     "creation.ocp.cluster.ost.must.enter.external.network.name": "External network name is required.",
+    "creation.ocp.cluster.ost.lb.floating.ip": "LB floating IP",
+    "placeholder.creation.ocp.cluster.ost.lb.floating.ip": "Enter LB floating IP",
```
KevinFCormier, line 259: On main branch, `create.json` file is obsolete. Please add these in `plugin__acm-plugin.json`

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/templates/install-config.hbs
```diff
@@ -306,7 +306,10 @@ publish: Internal
   openstack:
     cloud: {{{../../cloud}}}
     externalNetwork: {{{../../externalNetworkName}}}
-    lbFloatingIP: {{{../../apiFloatingIP}}}
+    {{#if ../../lbFloatingIP}}
+    lbFloatingIP: {{{../../lbFloatingIP}}}
+    {{/if}}
+    apiFloatingIP: {{{../../apiFloatingIP}}}
```
KevinFCormier, line 312: I think `apiFloatingIP` should be conditional as well.

KevinFCormier (changes requested):
See inline comments. Also a couple other concerns:

1. I am not sure of the meaning of 'lt' and 'gt' in functions `isHidden_lt_OCP48` and `isHidden_gt_OCP46`. I had guessed that it meant "less than" or "greater than"... but `isHidden_gt_OCP46` only checks for 4.6 (which is fine if we are not considering older versions), and `isHidden_lt_OCP48` checks for 4.8, 4.9, and 4.10. So that makes sense as well, but actually, shouldn't apiFloatingIP check for `!ishidden_gt_OCP46`. Currently if I select OCP 4.7, it does not show.
2. The handlebars templating is based on whether the value is present, so if I enter an lbFloatingIP, and then change my OCP version, it will still be present in the YAML when I enter an apiFloatingIP. I wonder if these could really be the same form element, with just the labels and YAML key changing when the version changes? Might be a question for @jeswanke.

Randy424:
@KevinFCormier 
Demo of recent change:

https://user-images.githubusercontent.com/21374229/145893091-ea672c25-4882-4be5-ac52-a97ddec6a9e9.mov

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
```diff
@@ -500,6 +507,22 @@ export const isHidden_lt_OCP48 = (control, controlData) => {
     return true
 }
 
+export const isHidden_gt_OCP46 = (control, controlData) => {
+    const singleNodeFeatureFlag = controlData.find(({ id }) => id === 'singleNodeFeatureFlag')
+    const imageSet = controlData.find(({ id }) => id === 'imageSet')
+    //NOTE: We will need to adjust this in the future for new OCP versions!
```
KevinFCormier, line 513: This comment is not really accurate... the function checks for anything 4.6 (or older) but 4.6 is the oldest we've ever supported deploying, so we will not need to add new versions here.

KevinFCormier (approved):

PR #1005, opened 2021-11-23
csrf header

jamestalton:


KevinFCormier:
See my version here: https://github.com/open-cluster-management/console/pull/1004

PR #1006, opened 2021-11-23
Fix page crash when no policy report found

KevinFCormier:
https://github.com/open-cluster-management/backlog/issues/18112


jamestalton (approved):
lgtm

PR #1007, opened 2021-11-23
Add toggle group for different types of resources

chenz4027:


frontend/src/routes/Applications/AdvancedConfiguration.tsx
```diff
@@ -69,69 +153,160 @@ export default function AdvancedConfiguration() {
...
+                    plural={t('Applications')}
+                    columns={selectedResources.columns}
+                    keyFn={keyFn}
+                    items={selectedResources.items}
+                    extraToolbarControls={
+                        <QuerySwitcher
+                            key="switcher"
+                            options={options.map(({ id }) => ({
+                                id,
+                                contents: t(`application:resource.${id}`),
```
chenz4027, line 276: We would need the id to determine which toogle is selected and render the corresponding title. Thus I would need useTranslation('application'). @RobDolares would this affect your translation work that you commit in https://github.com/open-cluster-management/console/pull/986 ?

frontend/src/routes/Applications/AdvancedConfiguration.tsx
```diff
@@ -12,41 +17,120 @@ import {
...
 import { useTranslation } from 'react-i18next'
 import { useRecoilState } from 'recoil'
-import { applicationsState } from '../../atoms'
+import { channelsState, placementRulesState, placementsState, subscriptionsState } from '../../atoms'
 import { IResource } from '../../resources'
+import _ from 'lodash'
 
 export default function AdvancedConfiguration() {
-    const { t } = useTranslation()
+    const { t } = useTranslation('application')
```
chenz4027, line 35: Not sure if this application would affect any translation work.

frontend/src/routes/Applications/AdvancedConfiguration.tsx
chenz4027: If so, would you recommend any alternative ways of finding the title?

KevinFCormier: Can you add the title to your metadata above and use `selectedResources.title`?

frontend/src/routes/Applications/AdvancedConfiguration.tsx
KevinFCormier: I think the idea is to have a single file with translations (ie. not specify 'applications' here). The keys are just the english strings themselves. So then later, there is a tool to run to collect these into a single file. @RobDolares @jamestalton please correct me if I'm wrong.

frontend/src/routes/Applications/AdvancedConfiguration.tsx
chenz4027: Sure that's def doable. Thanks Kevin 👍

frontend/src/routes/Applications/AdvancedConfiguration.tsx
RobDolares: @KevinFCormier @chenz4027 - That is correct. we won't reference a specific translation file.  Instead the full text needs to be in the file where it will be translated.  We are still not sure how dynamic string injection is going to be handled, but in cases where enums are used for a dynamic string, i have been creating `switch` statements to return all possible values until we can determine best way to handle string interpolation. @jamestalton

chenz4027: Thanks Robert, I'll make the change accordingly.

frontend/public/locales/en/application.json
```diff
@@ -57,6 +57,10 @@
     "repo.type.label.chart": "Chart name",
     "repo.type.label.packageFilterVersion": "Package version",
     "repo.type.label.bucketPath": "Subfolder",
+    "resource.channels": "Channels",
+    "resource.placementrules": "Placement rules",
+    "resource.placements": "Placements",
+    "resource.subscriptions": "Subscriptions",
```
KevinFCormier, line 63: We can revert this change, right? And if you can confirm if this file is no longer used, let's remove it.

frontend/src/resources/placement.ts
```diff
@@ -0,0 +1,55 @@
...
+        apiVersion: PlacementApiVersion,
+        kind: PlacementKind,
+    })
+    return {
+        promise: placements.promise.then((placements) => {
+            return placements.map((placement) => placement.metadata.name).filter(isType)
+        }),
+        abort: placements.abort,
+    }
+}
```
KevinFCormier, line 55: Do we need this function? Could we just use the atom and useRecoilState hook?

frontend/src/routes/Applications/AdvancedConfiguration.tsx
```diff
@@ -69,69 +234,173 @@ export default function AdvancedConfiguration() {
...
+            { id: 'placementrules', title: 'Placement rules' },
+        ]
+
+        const selectedId = getSelectedId({ location, options, defaultOption, queryParam: 'resources' })
+        const selectedResources = _.get(table, `${selectedId}`)
+
+        return (
+            <AcmTablePaginationContextProvider localStorageKey="advanced-tables-pagination">
+                <AcmTable<IResource>
+                    plural={t('Applications')}
```
KevinFCormier, line 361: Use the correct title here

frontend/src/routes/Applications/CreateApplication/CreateApplication.tsx
```diff
@@ -64,19 +64,39 @@ export default function CreateApplicationPage() {
 export function CreateApplication() {
     // will need to pass argoNs to AppForm to get argo namespaces
     const [argoNs, setArgoNs] = useState<string[]>([])
+    const [availablePlacements, setAvailablePlacements] = useState<string[]>([])
+
     useEffect(() => {
         const fetchNs = async () => {
             try {
                 let newNs = await listAvailableArgoServerNS().promise
+                let newPlacements = await listPlacements().promise
```
KevinFCormier, line 73: Can we get both of these from Recoil state instead?

KevinFCormier (changes requested):

frontend/public/locales/en/application.json
chenz4027: Seems like overview is still using it.

frontend/src/resources/resource.ts
```diff
@@ -10,6 +10,7 @@ export interface IResourceDefinition {
 }
 
 export interface IResource extends IResourceDefinition {
+    spec?: any
```
KevinFCormier, line 13: Please remove this and instead set the type parameter correctly on line 150 of console/frontend/src/atoms.tsx

frontend/src/resources/gitops-cluster.ts
```diff
@@ -33,13 +32,3 @@ export function listGitOpsClusters() {
         kind: GitOpsClusterKind,
     })
 }
```
KevinFCormier, line 34: Looks like listGitOpsClusters is also unused now?

KevinFCormier (changes requested):

chenz4027:
@KevinFCormier Removed the unused function and refactored the atom instead of the IResource :)

PR #1015, opened 2021-11-29
Add application search-api queries

zlayne:

Add infrastructure to auto-generate search-api Apollo query hooks.
Add the following application queries:
- applications
- subscriptions
- placementRules
- channels

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -580,6 +580,15 @@
                 "@babel/helper-plugin-utils": "^7.8.3"
             }
         },
+        "@babel/plugin-syntax-flow": {
+            "version": "7.16.0",
+            "resolved": "https://registry.npmjs.org/@babel/plugin-syntax-flow/-/plugin-syntax-flow-7.16.0.tgz",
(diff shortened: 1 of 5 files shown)


KevinFCormier:
My hope was that we would not continue to use search-api queries for the app lifecycle UI... but if we need to for the time being in order to get the app lifecycle UI, I would be OK with that.

@fxiang1 @chenz4027 what are your thoughts on this?

chenz4027:
@KevinFCormier Hey Kevin, the reason I pinged Zack about the searchapi queries is because that advanced configuration table would need some fields that is provided by search, but not kube resource like app count and cluster count. If we could figure out a way to get those resources from kube instead of search, I'm good with that too

fxiang1:
I think for getting the Argo applications on the managed clusters we would still need to use search for now. I know there's a proxy api in the works that would allow us to make direct kube call to managed clusters but not sure when that work will complete.

fxiang1:
@chenz4027 I can help you with the app count and cluster count so you don't need to use search for the Advanced configuration table

chenz4027:
@fxiang1 That would be great!

zlayne:
Closing. We will use existing resource data instead of introducing new search queries.

PR #1019, opened 2021-11-30
Cluster set scroll fix

Randy424:
regarding: https://github.com/open-cluster-management/backlog/issues/18141
Screen Shot 2021-11-30 at 12 43 38 PM

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetAccessManagement/ClusterSetAccessManagement.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetAccessManagement/ClusterSetAccessManagement.tsx
@@ -301,7 +301,7 @@ function AddUsersModal(props: {
     }
 
     return (
-        <AcmModal variant={ModalVariant.medium} title={t('access.add.title')} isOpen={props.isOpen} onClose={reset}>
+        <AcmModal variant={ModalVariant.medium} style={{overflow:'visible'}} title={t('access.add.title')} isOpen={props.isOpen} onClose={reset}>
             <AcmForm style={{ gap: 0 }}>
                 <AcmAlertContext.Consumer>
                     {(alertContext) => (
@@ -354,7 +354,7 @@ function AddUsersModal(props: {
                             &nbsp;
                             <AcmSelect
                                 id="role"
-                                maxHeight="6em"
+                                maxHeight="25em"
                                 isRequired
                                 label={t('access.add.role')}
                                 placeholder={t('access.select.role')}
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetAccessManagement/ClusterSetAccessManagement.tsx
```diff
@@ -301,7 +301,13 @@ function AddUsersModal(props: {
     }
 
     return (
-        <AcmModal variant={ModalVariant.medium} title={t('access.add.title')} isOpen={props.isOpen} onClose={reset}>
+        <AcmModal
+            variant={ModalVariant.medium}
+            style={{ overflow: 'visible' }}
```
Randy424, line 306: @KevinFCormier In your view, should this change be applied to the ui-component instead?

frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetAccessManagement/ClusterSetAccessManagement.tsx
```diff
@@ -354,7 +360,7 @@ function AddUsersModal(props: {
                             &nbsp;
                             <AcmSelect
                                 id="role"
-                                maxHeight="6em"
+                                maxHeight="10em"
```
Randy424, line 363: @KevinFCormier 6em gets us to about 1.5 visible select elements, no reason to skimp here as Patter Fly guidelines do suggest 7 or less items be visible when select dropdown is active, so I extended the length to cover both elements in the dropdown, I might add more units but this list ("User role") does not look like it will change in size often...unless cluster set roles are something users can add to. In which case I would probably want to increase this to **25-30em**. I do wonder why we use **em** here too...

frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetAccessManagement/ClusterSetAccessManagement.tsx
KevinFCormier: Last time we had an issue like this, I think we used `menuAppendTo="parent"` on the `AcmSelect`... would that work here?

Randy424: Thanks, appendTo() is the more elegant solution for sure.

frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetAccessManagement/ClusterSetAccessManagement.tsx
KevinFCormier: I don't anticipate adding more options, and to be honest, I'm not sure what the history behind using **em** is, but the last case of this we used `maxHeight="18em"`

https://github.com/open-cluster-management/console/blob/main/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx#L545

PR #1026, opened 2021-12-01
Add delete function to app overview table

fxiang1:
Issue: https://github.com/open-cluster-management/backlog/issues/17682

- Implemented the delete modal for both appsub and appset
- The modal data loads much faster now that we're not using search

image

image

image


frontend/public/locales/en/application.json
```diff
@@ -10,6 +10,19 @@
     "channel.type.label.chart": "Chart name",
     "channel.type.label.packageFilterVersion": "Package version",
     "channel.type.label.bucketPath": "Subfolder",
+    "delete.resource.modal.application.heading": "Permanently delete {0} application?",
```
chenz4027, line 13: Given the translation effort that Robert is working on, I think we stops using application.json. Could you move the tests to where they are referred instead?

chenz4027: Ah I see. Yep I think you could remove the application.json as no one is using it

chenz4027 (approved):
lgtm

PR #1035, opened 2021-12-07
Failed provisions via UnknownErrors display red

Randy424:

### DIFF
--- a/frontend/src/resources/utils/get-cluster.ts
+++ b/frontend/src/resources/utils/get-cluster.ts
@@ -803,7 +803,8 @@ export function getClusterStatus(
                 const currentProvisionRef = clusterDeployment.status?.provisionRef?.name ?? ''
                 if (
                     provisionFailedCondition?.message?.includes(currentProvisionRef) ||
-                    provisionFailedCondition?.reason === 'InvalidInstallConfig'
+                    provisionFailedCondition?.reason === 'InvalidInstallConfig' ||
+                    provisionFailedCondition?.reason === 'UnknownError'
                 ) {
                     cdStatus = ClusterStatus.provisionfailed
                 } else {


frontend/src/resources/utils/get-cluster.ts
```diff
@@ -803,7 +803,8 @@ export function getClusterStatus(
                 const currentProvisionRef = clusterDeployment.status?.provisionRef?.name ?? ''
                 if (
                     provisionFailedCondition?.message?.includes(currentProvisionRef) ||
-                    provisionFailedCondition?.reason === 'InvalidInstallConfig'
+                    provisionFailedCondition?.reason === 'InvalidInstallConfig' ||
+                    provisionFailedCondition?.reason === 'UnknownError'
```
Randy424, line 807: For these situations, the deployment status will still report a failed provision, but the reason it provides is "UnknownError", however, in the message field we get enough detail to understand the real problem, in this case it would be  "No public route53 zone found matching name 'invalid.red-chesterfield.com' "

At first I thought, that the "reason" field should be used here because it strikes me as more likely to remain fixed, and it is more succinct...but it also, possibly, covers a larger range of issues. I wondered if it made sense to throw red when we encounter a "ProvisionFailed" + "UnkownError". 

@KevinFCormier, In this case, do you think it makes sense to use "ProvisionFailed" + "UnknownError" to designate a failed status, or should we include a message check.?

Randy424:
W/r to the build: I'm getting 500 error from prow. Not sure why. 

`ERRO[2021-12-07T19:15:54Z] 
  could not run steps: step src failed: the build src failed after 1m5s with reason DockerBuildFailed: Dockerfile build strategy has failed.
Writing manifest to image destination
Storing signatures
Adding transient rw bind mount for /run/secrets/rhsm
STEP 1: FROM image-registry.openshift-image-registry.svc:5...7410adc20f2eb1778786c33a29b03a1ff3c029cb4eca7337efd17286687
error: build error: error creating build container: Error ...valid status code from registry 500 (Internal Server Error)`

frontend/src/resources/utils/get-cluster.ts
KevinFCormier: I think we need more information on these conditions to understand when the failure is permanent, and when there might be a retries. We may need to combine with other conditions to know this. Do you have the full ClusterDeployment conditions for this case?

Randy424:
```status:
    cliImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:14aa7a76f70357de44414af03adc98d41ab7f269f2308ca524660e16626a5652
    conditions:
    - lastProbeTime: "2021-12-10T23:22:10Z"
      lastTransitionTime: "2021-12-10T23:22:10Z"
      message: |
        level=info msg=Credentials loaded from default AWS environment variables
        level=info msg=Consuming Install Config from target directory
        level=fatal msg=failed to fetch Common Manifests: failed to fetch dependency of "Common Manifests": failed to generate asset "DNS Config": getting public zone for "invalid.red-chesterfield.com": No public route53 zone found matching name "invalid.red-chesterfield.com"
      reason: UnknownError
      status: "True"
      type: ProvisionFailed
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Platform credentials passed authentication check
      reason: PlatformAuthSuccess
      status: "False"
      type: AuthenticationFailure
    - lastProbeTime: "2021-12-10T23:21:57Z"
      lastTransitionTime: "2021-12-10T23:21:57Z"
      message: Images required for cluster deployment installations are resolved
      reason: ImagesResolved
      status: "False"
      type: InstallImagesNotResolved
    - lastProbeTime: "2021-12-10T23:21:57Z"
      lastTransitionTime: "2021-12-10T23:21:57Z"
      message: InstallerImage is resolved.
      reason: InstallerImageResolved
      status: "False"
      type: InstallerImageResolutionFailed
    - lastProbeTime: "2021-12-10T23:21:57Z"
      lastTransitionTime: "2021-12-10T23:21:57Z"
      message: Provision is not stopped
      reason: ProvisionNotStopped
      status: "False"
      type: ProvisionStopped
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: no ClusterRelocates match
      reason: NoMatchingRelocates
      status: "False"
      type: RelocationFailed
    - lastProbeTime: "2021-12-10T23:21:57Z"
      lastTransitionTime: "2021-12-10T23:21:57Z"
      message: All pre-provision requirements met
      reason: AllRequirementsMet
      status: "True"
      type: RequirementsMet
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: AWSPrivateLinkFailed
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: AWSPrivateLinkReady
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: ActiveAPIURLOverride
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: ClusterInstallCompleted
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: ClusterInstallFailed
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: ClusterInstallRequirementsMet
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: ClusterInstallStopped
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: ControlPlaneCertificateNotFound
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: DNSNotReady
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: DeprovisionLaunchError
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: Hibernating
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: IngressCertificateNotFound
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: InstallLaunchError
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: SyncSetFailed
    - lastProbeTime: "2021-12-10T23:21:47Z"
      lastTransitionTime: "2021-12-10T23:21:47Z"
      message: Condition Initialized
      reason: Initialized
      status: Unknown
      type: Unreachable
    installRestarts: 11
    installStartedTimestamp: "2021-12-10T23:21:57Z"
    installVersion: 4.8.22
    installerImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3095f7e19c274f0ed9db42482b8e121b17d1e55bade7977e6dcd68e2cac00eb6
    provisionRef:
      name: rbrunopi-aws-cluster-pool-8xf4z-11-xjzdl
```

Randy424: @KevinFCormier Also, it looks like cluster table is able to pick up on this failure, while the cluster pool table does not...

Randy424: @KevinFCormier 
https://github.com/open-cluster-management/console/blob/main/frontend/src/resources/utils/get-cluster.ts#L803-L807

This is the problem.

My best guess as to why this like this block of code exists (specifically these two lines):
```
const currentProvisionRef = clusterDeployment.status?.provisionRef?.name ?? ''
                if ( provisionFailedCondition?.message?.includes(currentProvisionRef) || ...
``` 

...I think this was just a way to isolate one failure case: The one in which the provision fails and in the failure message we receive the deployment provisionRef. 
There is at least one scenario where provisionRef value will not appear in the clusterDeployment (I stumbled upon it when I tried a vanilla cluster provision with an invalid base domain). In that situation, the provisionRef is removed from the deployment immediately after failure. The code above will then use an empty string for the if statement which will result in a true outcome each time, ergo, will throw a failed status in the managed cluster table. So we are already using the case "Unknown Error" as a failure scenario in our status logic...but it looks like because the clusterpool deployment behaves differently, our status logic is fooled. Clusterpool deployments preserve the provisionRef, and the because the invalid base domain error does not contain that provisionRef, this above block will in this case be skipped, and the UI will continue to report "Creating"

KevinFCormier: I notice that `installRestarts` is 11 in your ClusterDeployment... so I suspect in this particular case it keeps trying.

Randy424: Great point--it looks like the cluster yaml has installAttemptlimit defined by default. The cluster pool yaml omits that field and the install attempts continue indefinitely...so maybe the creating status is warranted haha. I think I'll test they result of adding a installAttemptlimit to the yaml...it should throw failure.

PR #1036, opened 2021-12-07
Wrap useTranslation and Trans to set namespace

KevinFCormier:
- Set namespace to `plugin__acm-plugin` so that translations work both standalone and as dynamic plugin
- Add linting rules to prevent developers from using `react-i18next` directly


jamestalton (approved):
lgtm

PR #1044, opened 2021-12-08
Translation clean-up

KevinFCormier:
Merging the translation namespaces did not take duplicate keys into account. Thanks @showeimer for noticing "Managed clusters" had become just "Clusters".

I checked the original translation files and found the following duplicate keys with different messages:

`bulk.message.delete`
**bma**: You are about to delete bare metal assets. The bare metal assets will no longer be available.
**credentials**: You cannot create new clusters from deleted credentials. Clusters that you previously created will not be affected.
**infraenv**: You are about to delete infrastructure environments. The infrastructure environments will no longer be available.

`bulk.title.delete`
**bma**: Permanently delete bare metal assets?
**credentials**: Permanently delete credentials?
**infraenv**: Delete infrastructure environments?

`bulk.title.import`
**bma**: Import bare metal assets
**cluster**: Import clusters

`clusters`
**cluster**: Managed clusters
**create**: Managed clusters
**discovery**: Clusters


KevinFCormier:
Cases for **credentials** were already handled.

jamestalton (approved):
lgtm

PR #1060, opened 2021-12-15
Update backend watch to run a list, reconcile and then watch

jamestalton:


backend/config/LOG_LEVEL
```diff
@@ -1 +1 @@
-debug
\ No newline at end of file
+trace
```
KevinFCormier, line 1: Was it intentional to commit this change?

jamestalton: yes

PR #1066, opened 2021-12-16
Add timer to improve render performance

fxiang1:

https://github.com/open-cluster-management/backlog/issues/17682

- Added a timer to slow down rendering the page so it doesn't consume too much CPU

Before the fix:
image

After the fix:
image


chenz4027 (approved):
lgtm, thanks Feng!

PR #1080, opened 2022-01-06
Aws gov handlebar edit

Randy424:
https://github.com/stolostron/backlog/issues/18664


frontend/src/routes/Infrastructure/Clusters/ClusterPools/CreateClusterPool/CreateClusterPool.tsx
```diff
@@ -45,6 +45,10 @@ const Portals = Object.freeze({
     cancelBtn: 'cancel-button-portal-id',
 })
 
+Handlebars.registerHelper('serviceEndpoint', (options) => {
+    if (options && options[0]?.endpointName) return '    serviceEndpoints:'
```
KevinFCormier, line 49: Is it possible to have the indentation be part of the template instead?

KevinFCormier: Hmm, I see why it can't now... 🤔

KevinFCormier (changes requested):
Hey Randy,

Your handlebars helper only checks if there is a name for the first endpoint, but it needs to check if there is a name for any endpoint to avoid this problem:
image

Could you try setting up your helper like this: https://handlebarsjs.com/guide/builtin-helpers.html#sub-expressions

The helper would just check if any serviceEndpoint has an endpointName, then you can keep the indentation in the template only.

Also, I think all instances of `Url` and `url` should be `URL` in the strings.

KevinFCormier (approved):

PR #1115, opened 2022-01-21
Remove application table timer

KevinFCormier:
I think there may have been a previous environmental issue that was triggering constant rerendering of the table. Removing this code to reduce rendering, I only see renders as expected from the polling search query. Verified using this React dev tools option:

image

I think there may be more room to improve by updating the useQuery hook to only update state when the result of the query actually changes. Will investigate further and open a separate PR for that if it seems feasible.

### DIFF
--- a/frontend/src/routes/Applications/Overview.tsx
+++ b/frontend/src/routes/Applications/Overview.tsx
@@ -157,7 +157,7 @@ export default function ApplicationsOverview() {
         open: false,
     })
     const tableItems: IResource[] = []
-    const { data, loading, startPolling } = useQuery(useCallback(() => queryRemoteArgoApps(), []))
+    const { data, loading, startPolling } = useQuery(queryRemoteArgoApps)
     useEffect(startPolling, [startPolling])
 
     const calculateClusterCount = (resource: ArgoApplication, clusterCount: any, clusterList: string[]) => {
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Applications/Overview.tsx
```diff
@@ -871,46 +893,6 @@ export default function ApplicationsOverview() {
...
-            .catch((err) => console.error(err))
-        return () => canDeleteApplicationSetPromise.abort()
-    }, [])
-
-    useEffect(() => {
-        setTableContent(generateTable())
-
-        const interval = setInterval(() => {
-            setTableContent(generateTable())
-        }, 5000)
```
KevinFCormier, line 904: Because the closure of the generateTable() function captures the state at the time useEffect fires, this interval does not achieve anything because it will render the same state every 5 seconds.

frontend/src/routes/Applications/Overview.tsx
```diff
@@ -871,46 +893,6 @@ export default function ApplicationsOverview() {
...
-        return () => {
-            clearInterval(interval)
-        }
-    }, [canCreateApplication, canDeleteApplication, canDeleteApplicationSet, data])
```
KevinFCormier, line 908: Because this call to `useEffect` has `data` as a dependency, each time the search query updates the data (every 5 seconds) the effect will fire again, so we get an updated rendering every 5 seconds. But if a change to a local kube resources comes through via the watch and recoil state, this has the effect of delaying that update until one of the periodic search query refreshes. Removing this effect, I see a rendering every 5 seconds, plus immediate rendering when recoil state is updated. I did not observe excessive rerendering.

chenz4027 (approved):
lgtm, thanks for the quick fix!

PR #1121, opened 2022-01-24
Add provider connection changed to add credential

ellajuengst:
https://github.com/stolostron/backlog/issues/19371

Signed-off-by: Ella Juengst <ellajuengst@gmail.com>

### DIFF
--- a/frontend/src/routes/Home/Overview/OverviewPage.tsx
+++ b/frontend/src/routes/Home/Overview/OverviewPage.tsx
@@ -218,11 +218,11 @@ const PageActions = (props: { timestamp: string; reloading: boolean; refetch: ()
                     component={Link}
                     variant={ButtonVariant.link}
                     to={NavigationPath.addCredentials}
-                    id="add-provider-connection"
+                    id="add-credential"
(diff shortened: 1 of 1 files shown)


KevinFCormier (changes requested):
Looks good!

We do not support translation yet, but we are working to get there. You may have noticed we use a mix of symbolic keys for the strings along with the actual strings.

In the interest of improving things, can you please:
1. Remove the old string from `frontend/public/locales/en/plugin__acm-plugin.json`
2. From the `frontend` directory, run `npm run i18n`. This will add your new string to the translation file. It will also sort the strings, so you will see a bunch of changes - please verify nothing is removed or the content altered. You might also notice some errors. I see the string `common:learn.more` just repeats the same value for its string and this is a visible issue in the UI:

image

Would you mind fixing that one up as well? It should just say "Learn more" and there is already an entry in the translation file for it. Thanks.

ellajuengst:
#1126

ellajuengst:
#1127

KevinFCormier (approved):
Changes addressed in other PRs.

PR #1128, opened 2022-01-25
Non existing secret catch

ellajuengst:
https://github.com/stolostron/backlog/issues/17841

1. Added catch to alert user on cluster creation when secret is undefined (non-existing) and provide direction on what steps to take when this error occurs.
2. Filtered list of automation templates to only show those with an existing Ansible secret
3. Added tooltip to indicate to the user that the dropdown only shows templates with existing Ansible secret


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/CreateCluster.tsx
```diff
@@ -295,8 +295,14 @@ export default function CreateClusterPage() {
                 })
                 break
             case 'templateName':
-                control.available = curatorTemplates.map((template) => template.metadata.name)
-                setAvailableTemplates(control, curatorTemplates)
```
ellajuengst, line 290: Randy and I determined that this function call was redundant and repeating the same task as in the line above it (line 298). I removed this function call and filtered out templates where no secret exists.

RobDolares (approved):
/lgtm

PR #1131, opened 2022-01-26
Cluster set fix for cluster claims and cluster pools

Randy424:
regarding: https://github.com/stolostron/backlog/issues/19173

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetManageResources/ClusterSetManageResources.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetManageResources/ClusterSetManageResources.tsx
@@ -1,6 +1,7 @@
 /* Copyright Contributors to the Open Cluster Management project */
 
 import {
+    ClusterDeployment,
     IResource,
     ManagedClusterKind,
     managedClusterSetLabel,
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetManageResources/ClusterSetManageResources.tsx
```diff
@@ -62,31 +63,37 @@ export function ClusterSetManageResourcesPage() {
...
+    clusterDeployments?.forEach((deployment) => deploymentDictionary.set(deployment.metadata.name, deployment))
+
+    const [managedClusters, managedClusterSets] = useRecoilValue(
+        waitForAll([managedClustersState, managedClusterSetsState])
     )
     const { canJoinClusterSets, isLoading } = useCanJoinClusterSets()
     const canJoinClusterSetList = canJoinClusterSets?.map((clusterSet) => clusterSet.metadata.name)
     const [selectedResources, setSelectedResources] = useState<IResource[]>(
-        [...managedClusters, ...clusterPools].filter(
+        [...managedClusters].filter(
```
KevinFCormier, line 76: Since this array now only contains ManagedClusters and not ClusterPools, should we use a more specific type?

Randy424: I've changed the type, but let me know if that is enough specificity...The page itself and some of the variables are using vague language ('resource(s)' to refer to the managedClusters), because the page was dealing with more than one resource. I was resisting the urge to specify the entire file for 'clusters', but wasn't sure if I should do so.

KevinFCormier: Yeah, that's fine. I had the same thought, but now am wondering if we might want to revisit this at some point... so that the members that cannot have their membership changed are still shown but not editable. But let's just go with this for now.

PR #1133, opened 2022-01-27
Fix cluster status popover 2.4

KevinFCormier:
https://github.com/stolostron/backlog/issues/19380


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/StatusField.tsx
```diff
@@ -157,7 +157,7 @@ export function StatusField(props: { cluster: Cluster }) {
                 maxWidth: '448px',
                 bodyContent: (
                     <>
-                        <Trans i18nKey={`cluster:status.${props.cluster?.status}`} components={{ bold: <strong /> }} />
+                        <Trans i18nKey={`cluster:status.${props.cluster?.status}.message`} components={{ bold: <strong /> }} />
```
KevinFCormier, line 160: @Randy424 The `.message` was removed by https://github.com/stolostron/console/pull/999

frontend/public/locales/en/cluster.json
```diff
@@ -352,6 +352,7 @@
     "status.detached.message": "The cluster creation was successful, but the cluster is not managed by the hub cluster. Import the cluster to enable management capabilities, such as search, policy, applications, and observability.",
     "status.detaching.message": "The cluster is being removed from management. Access to management features, such as search, policy, applications, and observability are being removed.",
     "status.notaccepted.message": "The cluster was not approved by the cluster administrator to attach to the hub cluster.",
+    "status.notstarted.message": "The cluster provision could not be started.",
```
KevinFCormier, line 355: @Randy424 The `.message` strings end with a period and seem to be longer than the `.alert.title` strings. Is there any more information we could add about this new status?

Randy424: https://github.com/Randy424/console/blob/6531fb52c3b27c92723a1f0b26f1858ff0d025cb/frontend/src/resources/utils/get-cluster.ts#L795-L800

So within the cluster status logic, there is currently only one scenario that will trigger the "notstarted" status, and that is when the install config is misconfigured (and there are [a few things Hive checks ](https://github.com/openshift/hive/blob/b212e8a45e956e8e59a5f8fcb108f298b3708442/pkg/controller/clusterdeployment/clusterdeployment_controller.go#L741-L789)to determine this). So we could include the detail that the install config is malformed.

KevinFCormier: Thanks Randy. "malformed" to me sounds like a syntax error, which it could be (one case is unmarshal fails), but it can also be a mismatch between certain items in the install-config vs. the ClusterDeployment:
https://github.com/openshift/hive/blob/b212e8a45e956e8e59a5f8fcb108f298b3708442/pkg/controller/clusterdeployment/installconfigvalidation.go#L22

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/StatusField.tsx
cadawson: Looks good. We refer to "cluster provision" as "cluster creation" most of the time. If that is the same in the UI, then I might look at:

The cluster creation could not start because the validation of the install-config.yaml file failed.

cadawson (changes requested):
Quick note, but looks good!

Randy424 (approved):
/lgtm

PR #1137, opened 2022-01-28
Add cluster proxy configuration

rawagner:


rawagner:
included in https://github.com/stolostron/console/pull/1247

PR #1140, opened 2022-01-31
Enable BM hosts on non-bm platforms too

rawagner:
Integration of https://github.com/openshift-assisted/assisted-ui-lib/pull/1103


rawagner:
included in https://github.com/stolostron/console/pull/1176

PR #1141, opened 2022-01-31
Updates in assisted-installer utils

jtomasek:
- Reset agent role on unbind
- Don't unhold installation on updating networking
   Cluster deployment networking configuration is now automatically saved
   on user input and only once backend validations pass the installation can
   be started by the user. Unholding the installation is therefore only
   done after user clicks 'Install' rather than on updating the networking
   configuration.


rawagner:
included in https://github.com/stolostron/console/pull/1247

PR #1149, opened 2022-02-02
governance overview and wizards

jamestalton:


frontend/src/components/StatusIcons.tsx
```diff
@@ -1,76 +1,41 @@
...
+            tooltip: props.compliantTooltip,
         })
     }
 
     if (props.unknown !== undefined && props.unknown) {
         statuses.push({
             key: 'ready',
             count: props.unknown,
-            icon: <QuestionCircleIcon color="var(--pf-global--icon--Color--light)" />,
+            icon: <ExclamationTriangleIcon color="var(--pf-global--icon--Color--light)" />,
```
zlayne, line 38: I think the design has the yellow/orange color instead of the grey. 
`<ExclamationTriangleIcon color="orange"`

frontend/src/resources/utils/resource-request.ts
```diff
@@ -61,6 +62,37 @@ export function createResource<Resource extends IResource, ResultType = Resource
...
 
+export async function createResources(resources: IResource[]): Promise<string | undefined> {
+    if (!Array.isArray(resources)) return 'Error - resources are not an array'
+    for (const resource of resources) {
+        try {
+            const existingResource = await getResource(resource).promise
+            if (existingResource)
+                return 'Resource of kind {0} in namespace {1} already exists'
+                    .replace('{0}', resource.metadata?.name ?? '')
+                    .replace('{1}', resource.metadata?.namespace ?? '')
```
zlayne, line 73: need to filter the existing resource from `resources` so we don't try to create it below.

frontend/src/components/StatusLabeledIcons.tsx
```diff
@@ -1,85 +1,47 @@
...
+            subtitle: props.compliantSubtitle,
         })
     }
 
     if (props.unknown !== undefined && props.unknown) {
         statuses.push({
             key: 'ready',
             count: props.unknown,
-            icon: <QuestionCircleIcon color="var(--pf-global--icon--Color--light)" />,
+            icon: <ExclamationTriangleIcon color="var(--pf-global--icon--Color--light)" />,
```
zlayne, line 44: same color change:
`<ExclamationTriangleIcon color="orange">`

frontend/src/resources/utils/resource-request.ts
jamestalton: This is for the create case and we return an error string if any already exist.  Will need to adjust for the case of update existing.

zlayne (approved):

PR #1158, opened 2022-02-03
cleanup console backend logging

jamestalton:

### DIFF
--- a/backend/src/lib/server.ts
+++ b/backend/src/lib/server.ts
@@ -96,23 +96,20 @@ export function startServer(options: ServerOptions): Promise<Http2Server | undef
                         if (options.logRequest) {
                             options.logRequest(req, res)
                         } else {
+                            if (req.url === '/authenticated') return
                             if (req.url === '/readinessProbe') return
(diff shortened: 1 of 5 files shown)


backend/src/lib/server.ts
```diff
@@ -96,23 +96,20 @@ export function startServer(options: ServerOptions): Promise<Http2Server | undef
                         if (options.logRequest) {
                             options.logRequest(req, res)
                         } else {
+                            if (req.url === '/authenticated') return
                             if (req.url === '/readinessProbe') return
                             if (req.url === '/livenessProbe') return
+                            if (req.url === '/apis/authorization.k8s.io/v1/selfsubjectaccessreviews') return
```
KevinFCormier, line 102: Is this related to logging?

jamestalton: Yes, those lines cause a lot of logging and it makes it harder to tell what real actions happened on the backend.
It seems like those could be debugged easier from the frontend dev tools networking console tab easier.

PR #1179, opened 2022-02-10
19223standalone header

mark-nc:
Changes for https://github.com/stolostron/backlog/issues/19223


frontend/src/App.tsx
```diff
@@ -58,6 +84,280 @@ interface IRouteGroup {
...
+    useEffect(() => {
+        // Get the username from the console backend
+        const resp = getUsername()
+        resp.promise
+            .then((payload) => {
+                payload && payload.body && payload.body.username ? setName(payload.body.username) : setName('undefined')
+            })
+            .catch((error) => {
+                // eslint-disable-next-line no-console
+                console.error(error)
```
KevinFCormier, line 147: Debugging code?

frontend/src/App.tsx
```diff
@@ -58,6 +84,280 @@ interface IRouteGroup {
...
+            .then((payload) => {
+                payload && payload.token_endpoint ? (oauthTokenEndpoint = payload.token_endpoint) : ''
+            })
+            .catch((error) => {
+                // eslint-disable-next-line no-console
+                console.error(error)
+            })
+        const logoutUrl = getBackendUrl() + '/logout'
+        console.log(location.href)
+        console.log(location.origin)
```
KevinFCormier, line 243: Some more console.logs to clean up. There are a bunch of others as well.

mark-nc: Yes, was waiting to see if logout for `kube:admin` worked using those for debugging.  I will pull those out.

mark-nc: console.log statements have been removed

frontend/src/App.tsx
mark-nc: No, in case the getUsername response had an error rather than an actual response (should be a pretty rare case if it can even occur) this catch is just to be sure the error ends up in the console, but making sure the name value is set to some string.

PR #1180, opened 2022-02-10
Updating RbacDropdown to only show tooltip if also disabled

ellajuengst:
RbacDropdown passes its tooltip prop to AcmDropdown only if its isDisabled prop is set to true. Otherwise it passes undefined as the tooltip. 
Signed-off-by: Ella Juengst <ellajuengst@gmail.com>


ellajuengst:
@KevinFCormier

PR #1184, opened 2022-02-11
Integrate new status logic

rawagner:
AI PR https://github.com/openshift-assisted/assisted-ui-lib/pull/1145


rawagner:
included in https://github.com/stolostron/console/pull/1287

PR #1226, opened 2022-02-17
Use SPA navigation to search

KevinFCormier:
Now that search is in the same app, we should use `<Link>` instead of `<a>`.


chenz4027 (dismissed):
lgtm, thanks Kevin!

PR #1228, opened 2022-02-21
add sync editor

jeswanke:


KevinFCormier:
@jeswanke I found a few issues.

1. The **Editor changes** section uses a 14px heading, which is the same as the label on fields. I think this might need to be 20px to match the section headings, and some feedback from design might be useful.
1. Secrets are not hidden in the **Editor changes** section if I reveal them, make changes, then hide them again.
1. Edit credential fails on submit.

image

```
TypeError
Cannot read properties of null (reading 'getAllDecorations')
Call Stack
 getResourceEditorDecorations
  src/components/SyncEditor/decorate.ts:111:10
 decorate
  src/components/SyncEditor/decorate.ts:24:60
 eval
  src/components/SyncEditor/SyncEditor.tsx:246:64
```

image

KevinFCormier:
@jeswanke Thanks for the updates. It's much closer now, except that the secret is not actually updated when I edit. The page does not crash and I get the toast saying it was successful, but it is not actually changed.

frontend/src/components/AcmDataForm.tsx
```diff
@@ -107,11 +113,13 @@ function requiredValidationMessage() {
 
 const minWizardSize = 1000
 const defaultPanelSize = 600
+const EDITOR_CHANGES = 'Editor changes'
```
KevinFCormier, line 116: Can we externalize this string, please?

KevinFCormier (changes requested):
@jeswanke I requested you externalize your strings. Also, when YAML editing is off and you go to create/edit/view a Credential, the drawer starts to pop out for a moment before it hides itself. That seems like a bug.

frontend/src/components/AcmDataForm.tsx
jeswanke: ok, now t( )

PR #1232, opened 2022-02-22
Make setup script work for any installation namespace

KevinFCormier:
QE often installs in into `ocm` namespace rather than `open-cluster-management`. Update setup script to work with any namespace.


fxiang1 (approved):
Looks good! Thanks Kevin!

PR #1237, opened 2022-02-23
Enable BMH only for selected platforms

rawagner:
requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1163


rawagner:
merged as part of https://github.com/stolostron/console/pull/1287

PR #1245, opened 2022-02-24
sonar smells

jeswanke:

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -4,7 +4,7 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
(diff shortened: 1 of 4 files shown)


frontend/public/index.html
```diff
@@ -4,7 +4,7 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
-        <link rel="icon" type="image/svg+xml" href="favicon.svg" />
+        <!-- <link rel="icon" type="image/svg+xml" href="favicon.svg" /> -->
```
KevinFCormier, line 7: Why this change, @jeswanke?

jeswanke: kept coming up in console that it didn't exist but looks like it exists in release so put it back

PR #1303, opened 2022-03-09
Fix empty state text for Application's Advanced configuration table #20262

veniceofcode:
https://github.com/stolostron/backlog/issues/20262

Signed-off-by: Raghuram Banda <rbanda@redhat.com>

### DIFF
--- a/frontend/src/routes/Applications/components/ToggleSelector.tsx
+++ b/frontend/src/routes/Applications/components/ToggleSelector.tsx
@@ -1,7 +1,8 @@
+/* eslint-disable no-debugger */
 /* Copyright Contributors to the Open Cluster Management project */
 
 import _ from 'lodash'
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Applications/components/ToggleSelector.tsx
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable no-debugger */
```
chenz4027, line 1: We would need to remove this line

chenz4027 (changes requested):

frontend/src/routes/Applications/components/ToggleSelector.tsx
veniceofcode: updated and pushed the change

PR #1319, opened 2022-03-11
Cluster pool updates

ellajuengst:
https://github.com/stolostron/backlog/issues/19598
1. Minor updates to clusterpool creation form
2. Added warning to clusterclaim modal when no running clusters
3. Added table for new claims when no associated deployment yet.


frontend/public/locales/en/translation.json
```diff
@@ -352,9 +352,11 @@
     "clusterClaim.create.timeOut": "Timed out waiting for a cluster to be assigned to the claim. Try checking the cluster claim later.",
     "clusterClaim.create.title": "Claim cluster",
     "clusterClaim.create.title.success": "Cluster successfully claimed",
+    "clusterClaim.delete": "Delete Claim",
```
KevinFCormier, line 355: Lowercase 'c' for claim.

frontend/public/locales/en/translation.json
```diff
@@ -2016,6 +2018,7 @@
     "table.submariner.connection": "Connection status",
     "table.submariner.nodes": "Gateway nodes labeled",
     "table.type": "Type",
+    "table.userIdentityStamp": "User Identity Stamp",
```
KevinFCormier, line 2021: Can we call this **Created by** instead?

KevinFCormier (commented):
A couple comments added for some of the strings. I will review more thoroughly next week. 

We should also:

1. Remove the timeout and error that appears when the claim is not immediately fulfilled. Just let the modal close (warning is meant to take the place of this).
2. Check with design on styling for the warning. Not sure if we are using the right kind of alert here.
3. We probably need a confirmation dialog for deleting the claim. Usually it's harmless, but if the claim were to get fulfilled then deleting the claim does delete that cluster, so we probably want to make sure the user really wants to delete.
4. I think we should also update the **Available to claim** column to only say Yes for Running clusters, update the **Available clusters** column to reflect this count, and make the **Claim cluster** button always available. 

I am working on my PR to fix the statuses for the clusters. Will try to have ready for review soon. I just need to add some test coverage.

image

frontend/src/routes/Infrastructure/Clusters/ClusterPools/ClusterPools.tsx
```diff
@@ -250,6 +256,26 @@ export function ClusterPoolsTable(props: {
...
+                            {
+                                cells: [
+                                    {
+                                        title: (
+                                            <>
+                                                {clusterPoolClaims.length > 0 && (
+                                                    <div style={{ marginTop: '16px', marginBottom: '16px' }}>
+                                                        <TextContent>
+                                                            <Text component={TextVariants.h3}>
+                                                                {t('clusterClaims')}
```
KevinFCormier, line 268: This heading needs to be **Pending claims** so it is clear it is not a list of all claims

PR #1337, opened 2022-03-15
Template changes, architecture field in creation wizard

Randy424:

regarding: https://github.com/stolostron/backlog/issues/20455 - https://github.com/stolostron/backlog/issues/20680

Screen Shot 2022-03-15 at 3 05 33 PM
Screen Shot 2022-03-15 at 3 05 42 PM
Screen Shot 2022-03-15 at 3 05 59 PM


Randy424:
FYI Architecture field should be position in the same way across providers...
https://docs.openshift.com/container-platform/4.6/installing/installing_aws/installing-aws-customizations.html#installation-configuration-parameters-optional_installing-aws-customizations

KevinFCormier:
@Randy424 Shouldn't we be able to determine the architecture from the instance type and then always set it? Currently we do not list any of the AWS ARM architecture instances (a1), nor do we for GCP, but in 4.10 `arm64` is another architecture option for both of these providers.

Randy424:
@KevinFCormier Ah, Gotcha. I think the doc page I was looking at while working on this was outdated. At this time I was running under the assumption that all the providers only supported amd64.

Picking supported values depending on provider would be simple enough, but looking at the 4.10 installer doc now, it looks like (at least where specified) that openshift-install is "supporting" amd64 and arn64 architecture in every provider doc (AWS, GCP, Alibaba, Azure, etc) with the disclaimer: _"[See Supported installation methods for different platforms in Installing](https://docs.openshift.com/container-platform/4.10/installing/installing_alibaba/installing-alibaba-customizations.html#installation-configuration-parameters-optional_installing-alibaba-customizations)"_

Seems to be a lot of potential overlap...We could hunt down exactly which provider supports which architecture, but I thought it might be simpler to include what we explicitly supported in the drop down and allow users to enter custom values when necessary via the combobox.

KevinFCormier:
@Randy424 My point was more that each instance type has a certain architecture, and we should set it based on that, rather than making the user select it. I don't think that we currently offer any arm64 instance types, but perhaps we should.

Randy424:
@KevinFCormier Ah, I see now. Machine instance types, right? That somehow registered as provider when I first read it haha. We might have to search for break downs of the available machine types for arm64 on other vendors. In our documentation AWS seems to be the only one with a [table](https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-aws-network-customizations.html#installation-supported-aws-machine-types_installing-aws-network-customizations).

...I wonder if it actually the case that AWS is the only one that has arm64 support at the moment.

PR #1346, opened 2022-03-16
Fix populating form fields when going back from Automation to Details…

rawagner:
… screen

requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1218


rawagner:
included in https://github.com/stolostron/console/pull/1361

PR #1348, opened 2022-03-16
RHV provider label detection logic added

Randy424:

Regarding: https://github.com/stolostron/backlog/issues/19273

ManagedClusterInfos supports RHV with this PR: https://github.com/stolostron/multicloud-operators-foundation/pull/457/files#diff-1464a4dd52c947d03556e70191424b21fd8f8a4fe2ea7abe2ce0c145aeaf4ce3R274-R279

The CloudVendor field is copied to the the 'cloud' resource label here: https://github.com/stolostron/multicloud-operators-foundation/blob/b852c6001862038284f080113fb0c7479726f0c1/pkg/controllers/autodetect/autodetect_controller.go#L95-L99

### DIFF
--- a/frontend/src/routes/Home/Overview/OverviewPage.tsx
+++ b/frontend/src/routes/Home/Overview/OverviewPage.tsx
@@ -38,41 +38,14 @@ import { ClusterManagementAddOn } from '../../../resources/cluster-management-ad
 import { fireManagedClusterView } from '../../../resources/managedclusterview'
 import { searchClient } from '../Search/search-sdk/search-client'
 import { useSearchResultCountLazyQuery, useSearchResultItemsLazyQuery } from '../Search/search-sdk/search-sdk'
-
-export function mapProviderFromLabel(provider: string): Provider {
-    switch (provider.toLowerCase()) {
-        case 'amazon':
-            return Provider.aws
-        case 'azure':
-            return Provider.azure
-        case 'baremetal':
-            return Provider.baremetal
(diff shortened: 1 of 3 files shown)


Randy424:
/cc @KevinFCormier 
FYI @dtthuynh

KevinFCormier (changes requested):
There's at least one other place where we map the providers that this PR doesn't cover:
https://github.com/stolostron/console/blob/main/frontend/src/routes/Home/Overview/OverviewPage.tsx#L43

Ideally these should share the same code.

Once that is done, on the Overview page, you'll need to hook up the display name for RHV as well:
image

Randy424:
- PR for the display name change is up.
- Provider detection function used in Overview has now been replaced with shared util function

frontend/src/routes/Home/Overview/OverviewPage.tsx
```diff
@@ -38,41 +38,14 @@ import { ClusterManagementAddOn } from '../../../resources/cluster-management-ad
...
-            return Provider.gcp
-        case 'ibm':
-            return Provider.ibm
-        case 'ibmpowerplatform':
-            return Provider.ibmpower
-        case 'ibmzplatform':
-            return Provider.ibmz
-        case 'redhat':
-            return Provider.redhatcloud
-        case 'vmware':
```
KevinFCormier, line 60: Can we move the VMWARE case to the new code?

PR #1359, opened 2022-03-18
BZ2065276: Enable adding hosts in more cluster states than just pendingimport

jtomasek:
Scale up the cluster is possible in pendingimport, ready and unknown
cluster states.


jtomasek:
Included in https://github.com/stolostron/console/pull/1361. Closing

PR #1424, opened 2022-03-28
BZ 2065071: Fix missing clusterDeployment for Scale Up

mareklibra:


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/assisted-installer/utils.ts
```diff
@@ -322,7 +322,10 @@ export const useClusterDeployment = ({
 }): CIM.ClusterDeploymentK8sResource | undefined => {
     const [clusterDeployments] = useRecoilValue(waitForAll([clusterDeploymentsState]))
     return useMemo(
-        () => clusterDeployments.find((cd) => cd.metadata.name === name && cd.metadata.namespace === namespace),
+        () =>
```
jtomasek, line 325: Is this needed? ``Array.find()` already returns `undefined` if no match is found.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/ClusterActionDropdown.tsx
```diff
@@ -412,7 +412,7 @@ export function ClusterActionDropdown(props: { cluster: Cluster; isKebab: boolea
             {
                 id: 'ai-scale-up',
                 text: t('managed.ai.scaleUp'),
-                click: () => setScaleUpModalOpen(true),
+                click: (cluster: Cluster) => setScaleUpModalOpen(cluster),
```
jtomasek, line 415: IIUC just a cluster name is needed here. Would be nice to limit the state to just that instead of whole cluster.

jtomasek (changes requested):

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/assisted-installer/utils.ts
mareklibra: Right. That was meant as an optimization, since the `name === undefined` is a frequent case, the array does not need to be searched.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/ClusterActionDropdown.tsx
mareklibra: done

jtomasek (approved):
/lgtm

PR #1425, opened 2022-03-28
Fix table actions

rawagner:

requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1261


jtomasek:
Included in https://github.com/stolostron/console/pull/1436

PR #1442, opened 2022-03-29
Fix logs tab for resources with pods

fxiang1:
Issue: https://github.com/stolostron/backlog/issues/21106

- Removed the Logs tab for resources that deploy pods
- Made sure all resources (ReplicationController, StatefulSet, Daemon, Pod) that deploy pods show a child node for Pod
- Display the Logs tab in the Pod node instead


chenz4027 (approved):
lgtm, thanks for the fix 👍

PR #1454, opened 2022-03-30
validate infra env form before submitting

rawagner:
requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1276 and https://github.com/stolostron/temptifly/pull/156


rawagner:
included in https://github.com/stolostron/console/pull/1468

PR #1461, opened 2022-03-30
Product version any user

KevinFCormier:
https://github.com/stolostron/backlog/issues/20234

### DIFF
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -185,7 +185,7 @@ async function listAndWatch(options: IWatchOptions) {
 }
 
 async function listKubernetesObjects(options: IWatchOptions) {
-    const serviceAcccountToken = getServiceAccountToken()
+    const serviceAccountToken = getServiceAccountToken()
     let resourceVersion = ''
     let _continue: string | undefined
     let items: IResource[] = []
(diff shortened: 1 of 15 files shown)


backend/src/routes/events.ts
```diff
@@ -185,7 +185,7 @@ async function listAndWatch(options: IWatchOptions) {
 }
 
 async function listKubernetesObjects(options: IWatchOptions) {
-    const serviceAcccountToken = getServiceAcccountToken()
+    const serviceAcccountToken = getServiceAccountToken()
```
mark-nc, line 188: The const var name still has the extra "c" `const serviceAcccountToken`

backend/src/routes/events.ts
```diff
@@ -253,7 +253,7 @@ async function listKubernetesObjects(options: IWatchOptions) {
 }
 
 async function watchKubernetesObjects(options: IWatchOptions, resourceVersion: string) {
-    const serviceAcccountToken = getServiceAcccountToken()
+    const serviceAcccountToken = getServiceAccountToken()
```
mark-nc, line 256: The const var name still has the extra "c" `const serviceAcccountToken`

mark-nc (changes requested):
Just two comments in backend/src/routes/events.ts but the rest looks good.

backend/src/routes/events.ts
KevinFCormier: Fixed

backend/src/routes/events.ts
KevinFCormier: Fixed

mark-nc (approved):
/lgtm

PR #1463, opened 2022-03-31
Bind all hosts from discovery step

rawagner:
AI PR https://github.com/openshift-assisted/assisted-ui-lib/pull/1279


jtomasek:
Included in https://github.com/stolostron/console/pull/1468

rawagner:
included in https://github.com/stolostron/console/pull/1468

PR #1464, opened 2022-03-31
Fix subscription blank problem

chenz4027:
Fix the Create application (subscription) wizard being blank


fxiang1 (approved):
LGTM!

PR #1469, opened 2022-04-01
MGMT-9313: Sorting by nodes and adding a sortable creation date in cluster list

ammont82:
Related to https://issues.redhat.com/browse/MGMT-9313

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
@@ -50,6 +50,7 @@ import { ClusterActionDropdown } from './components/ClusterActionDropdown'
 import { DistributionField } from './components/DistributionField'
 import { StatusField } from './components/StatusField'
 import { useAllClusters } from './components/useAllClusters'
+import { getDateTimeCell } from '../../helpers/table-row-helpers'
 
 function InfraEnvLinkButton() {
     const { t } = useTranslation()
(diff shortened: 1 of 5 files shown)


ammont82:
/hold
this is a 2.6 feature, please dont merge yet

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -548,3 +554,23 @@ export function useClusterNodesColumn(): IAcmTableColumn<Cluster> {
         },
     }
 }
+
+export function useClusterCreatedDateColumn(clusterDeployments: ClusterDeployment[]): IAcmTableColumn<Cluster> {
```
jtomasek, line 558: Would it be simpler to add the desired `creationTimeStamp` property to the `Cluster` type which is generated from various resources in `get-cluster.ts` module? We could avoid having to reach for `ClusterDeployments` in this component and most likely the sorting would be as simple as `sort: 'creationTimeStamp'`.

What do you think?

jtomasek (changes requested):

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
ammont82: Yes, it works better in this way. I've changed it

jtomasek (approved):
/lgtm

PR #1487, opened 2022-04-04
Fix topology search data not updating

fxiang1:
Issue: https://github.com/stolostron/backlog/issues/21257

- Change the way we query for search data
- Switched fetchPolicy from `cache-first` to `network-only` as the former was returning stale data


chenz4027 (approved):
lgtm

PR #1503, opened 2022-04-05
major edits

jeswanke:
stolostron/backlog#21350

any errors cause synceditor to return last valid resources plus the error messages

1. before if user deleted whole yaml, there was no error---now there's now an error that it requires a policy

image

2. if user just adds the policy, the errors will now validate what's required by a policy:

image

and synceditor won't send a bad resources to wizard until all errors are resolved


jamestalton (approved):
lgtm

PR #1509, opened 2022-04-06
Fix CRD topology status

KevinFCormier:
https://github.com/stolostron/backlog/issues/21470


fxiang1 (approved):
LGTM! and Thanks Kevin for fixing all the search URLs!

PR #1541, opened 2022-04-12
Let ManagedCluster cloud label take precedence over cluster claim

showeimer:
Signed-off-by: Sho Weimer <sweimer@redhat.com>


chenz4027 (approved):
lgtm

PR #1570, opened 2022-04-21
Need to enforce https for Ansible credentials

Randy424:
@jamestalton 
I was able to successfully setup the timeout but it didn't solve the problem. I noticed that none of the event emitters, even the one for timeouts set by my request, were being triggered when using an _http_ protocol for the host url (and the emitters worked for invalid https urls). I've done some more digging and haven't nailed the exact reason yet, but from what I've already found I suspect it has to do with the CORS configuration; that it could be stopping the request before it takes flight. 

At any rate, I don't think their api supports http anyways. One thing we might do is enforce https for Ansible credentials, as that will prevent users from running into this issue. That is what I have done in this PR. 
I wanted to ask if I need to take further steps to investigate this or if there is a more optimal solution? Maybe requiring https for this credential is enough?

regarding: https://github.com/stolostron/backlog/issues/21632


Randy424:
/hold 

Looks like Ansible tower does support http insecure requests. Tested on Postman.
Screen Shot 2022-04-21 at 11 32 36 AM

Randy424:
FYI @jamestalton @KevinFCormier 
This is ready for a review!

PR #1571, opened 2022-04-21
When loading state is false, stop re-attempting fetch

Randy424:
regarding: https://github.com/stolostron/backlog/issues/20510

@jeswanke Would you mind reviewing this? I tested to see that we still fetch the corresponding secret when choosing a previously used repo, and checked that the fetch is called once, but since this keeps coming back we should probably be thorough. Also Thanks for the break down on this issue back in March. Referring to those notes helped me with the fix this time.

### DIFF
--- a/frontend/src/routes/Applications/CreateApplication/Subscription/controlData/utils.js
+++ b/frontend/src/routes/Applications/CreateApplication/Subscription/controlData/utils.js
@@ -609,6 +609,7 @@ export const setAvailableSecrets = (control, result) => {
         }
     } else {
         control.isLoading = loading
+        if (!loading) control.isLoaded = true
     }
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Applications/CreateApplication/Subscription/controlData/utils.js
```diff
@@ -609,6 +609,7 @@ export const setAvailableSecrets = (control, result) => {
         }
     } else {
         control.isLoading = loading
+        if (!loading) control.isLoaded = true
```
Randy424, line 612: Control.isLoading is being updated before and after asynchronous loading is being performed. When `setAvailableSecrets` is first called, loading is in progress. Additional calls are made, but when there are no git secrets in the cluster, control.isLoaded is never updated.

 The idea here is that when control.isLoading is set to false (indicating that a fetch has finished), control.isLoaded needs to be updated to prevent subsequent calls, regardless of the contents of the results.

PR #1575, opened 2022-04-21
performance fix batch event handling

jamestalton:

processEventQueue() was called on every Loaded - which is sent after every event
this was causing every event to refresh the UI vs batching the updates

https://github.com/stolostron/backlog/issues/21852

### DIFF
--- a/frontend/src/atoms.tsx
+++ b/frontend/src/atoms.tsx
@@ -446,7 +446,6 @@ export function LoadData(props: { children?: ReactNode }) {
                             eventQueue.length = 0
                             break
                         case 'LOADED':
-                            processEventQueue()
                             setLoading(false)
                             break
                         case 'SETTINGS':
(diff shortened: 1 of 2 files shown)


KevinFCormier:
@jamestalton Was wondering if we want to wait until we see at least one LOADED event so the user doesn't see intermediate states... but if it does take longer than 500 ms to see  the first LOADED event, maybe we do want to start showing the UI even if states may keep updating after that?

jamestalton:
This should wait until one loaded event to setLoading(false) which will display a spinner until the load is ready.
I'm thinking that since we are bulk updating the UI should update really quickly now.

We could add a
        setTimeout(()=>setLoading(false), 2000)
if we do not want to wait for the LOADED event, but I am not sure it is needed.

KevinFCormier:
Ah yes, that's right. We start processing the data, but we don't display it until we've seen the LOADED event. Very good. Ready for review?

PR #1578, opened 2022-04-21
Allow cluster creation with no ClusterImageSets

KevinFCormier:
https://github.com/stolostron/backlog/issues/21673
https://github.com/stolostron/backlog/issues/21831

This reverts https://github.com/stolostron/console/pull/1543 because it is not applicable for non-CIM cluster creation. In the Hive-based cluster and cluster pool creation, we allow the user to enter a release image URL and automatically create a ClusterImageSet.

This PR also labels that ClusterImageSet so that it is visible in the UI for subsequent uses of the wizards.

We may want to adopt the same approach for the CIM wizard.


KevinFCormier:
FYI @rawagner

jeswanke (approved):

PR #1581, opened 2022-04-22
Add the window.getEditorValue method

mprahl:
Without having access to the Monaco Editor's API, it's very difficult to
get the entire text of the editor in Cypress based tests. Exposing the
Monaco Editor's getValue method makes this much easier without
exposing any additional APIs that may be of concern.


jamestalton:
We should not be adding values to the window global.
This would add that to the global for OCP/MCE/ACM.

mprahl:
@jamestalton could you please review this again after the latest change we discussed in Slack?

jamestalton (approved):
lgtm

PR #1583, opened 2022-04-22
Fix node status and Argo editor link

fxiang1:
Changes:

- Fix node status showing pending when there's only one pending resource from a cluster, should show a warning in this case
- Fix issue when RBAC user clicks Argo editor link they get no permission to list Routes, changed to list Routes from namespace instead


chenz4027 (approved):
lgtm!

PR #1587, opened 2022-04-25
Provide node link only if search available

KevinFCormier:
https://github.com/stolostron/backlog/issues/21922


RobDolares (approved):
lgtm

PR #1616, opened 2022-04-28
Fix double counting of clusters

KevinFCormier:
-Refactor code to avoid side-effects and duplication
-Change a tags to SPA links
-Externalize strings

https://github.com/stolostron/backlog/issues/22027


chenz4027 (approved):
lgtm, thanks Kevin

PR #1619, opened 2022-04-29
Bug 2080161 [UI] OpenShift version is not populated for draft clusters in cluster list

ammont82:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=2080161

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx
@@ -14,25 +14,46 @@ import { ButtonVariant } from '@patternfly/react-core'
 import { ArrowCircleUpIcon, ExternalLinkAltIcon } from '@patternfly/react-icons'
 import { Fragment, ReactNode, useState } from 'react'
 import { useTranslation } from '../../../../../lib/acm-i18next'
-import { useRecoilState } from 'recoil'
-import { ansibleJobState } from '../../../../../atoms'
+import { useRecoilState, useRecoilValue, waitForAll } from 'recoil'
+import { ansibleJobState, clusterImageSetsState } from '../../../../../atoms'
 import { RbacButton } from '../../../../../components/Rbac'
 import { rbacCreate, rbacPatch } from '../../../../../lib/rbac-util'
 import { BatchUpgradeModal } from './BatchUpgradeModal'
+import { useAgentClusterInstall } from '../CreateCluster/components/assisted-installer/utils'
+import { CIM } from 'openshift-assisted-ui-lib'
+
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx
```diff
@@ -14,23 +14,45 @@ import { ButtonVariant } from '@patternfly/react-core'
...
-    if (!props.cluster?.distribution) return <>-</>
+    if (!props.cluster?.distribution) {
+        if (props.cluster?.status === ClusterStatus.draft) {//cluster is in draft we try to get version from clusterimage
+            const clusterImage = clusterImageSets.find((clusterImageSet) => clusterImageSet.metadata?.name == agentClusterInstall.spec?.imageSetRef?.name)   
+            const version =  getVersionFromReleaseImage(clusterImage?.spec?.releaseImage)                     
+            return <>{version ? `OpenShift ${version}` : '-' }</>
+        }
+        else {
+            return <>-</>
+        }
```
jtomasek, line 53: How about changing this to not to be limited only to the draft status? Simply if `props.cluster.distribution` does not exist, we look at `agentClusterInstall`, try to find the `clusterImage` and get version from the release image. Then:
```
if (version) {
  return <>{version ? `OpenShift ${version}` : '-' }</>
} else {
  return <>-</>
}

**Review by jtomasek — CHANGES_REQUESTED** [review]

File: frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx (hunk 1 above)
Comment (ammont82) [reply frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx]: ok, is a good idea. I'll change it

Comment (ammont82) [reply frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx]: changed!

File: frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx [hunk 2]
```diff
@@ -14,25 +14,49 @@ import { ButtonVariant } from '@patternfly/react-core'
...
+            const version = getVersionFromReleaseImage(clusterImage?.spec?.releaseImage)
 
+            if (version) {
+                return <>{version ? `OpenShift ${version}` : '-'}</>
+            } else {
+                return <>-</>
+            }
+        } else {
+            return <>-</>
+        }
```
jtomasek, line 57:
```suggestion
        }
        return <>-</>
```

jtomasek (changes requested):

jtomasek (approved):

PR #1628, opened 2022-05-02
Fix deleting app with no subscription

chenz4027:
Fix issue where only happens when the user
- create an app with subscription
- delete the subscription
https://github.com/stolostron/backlog/issues/22110#issuecomment-1114894639


KevinFCormier (changes requested):
The problem is that `getSubscriptionsFromAnnotation` is returning `[""]` when the annotation value is `""`. I think we should return `[]` in this case. This function is used in a few other places, so this will fix any problems in those other spots as well.

PR #1630, opened 2022-05-02
Fix argo editor link and app table crash

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/22145

- Fix issue when user creates an Argo app in a namespace that they forgot to deploy a Gitops instance the app table would crash
- Fix an issue with the Argo editor link not working with non-default Gitops installs


chenz4027 (approved):
lgtm

PR #1635, opened 2022-05-03
fix app not found

jeswanke:

### DIFF
--- a/frontend/src/routes/Applications/ApplicationDetails/ApplicationDetails.tsx
+++ b/frontend/src/routes/Applications/ApplicationDetails/ApplicationDetails.tsx
@@ -5,6 +5,7 @@ import { Alert } from '@patternfly/react-core'
 import {
     AcmActionGroup,
     AcmAlert,
+    AcmLoadingPage,
     AcmPage,
     AcmPageHeader,
     AcmRoute,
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Applications/ApplicationDetails/ApplicationDetails.tsx
```diff
@@ -74,6 +75,12 @@ export const ApplicationContext = createContext<{
 const namespaceString = ':namespace'
 const nameString = ':name'
 
+// in case where application is just created
+// because recoil state is throttled
+// the new application might yet be in recoil state
+// so if application not found
+const WAIT_FOR_APPLICATION_DELAY = 1000
```
KevinFCormier, line 82: I would suggest defining and exporting the constant value here:
https://github.com/stolostron/console/blob/main/frontend/src/atoms.tsx#L489

KevinFCormier: That's not what I meant. I was suggesting to make a constant for the 500 ms value, and to use the same one in your code. Does the value need to be higher than 500? If so, it could be that you always wait for twice the polling time in your code.

jeswanke: done

PR #1639, opened 2022-05-04
Fallback to local-cluster if there are no searchClusters

fxiang1:

Issues:
- https://github.com/stolostron/backlog/issues/22063
- https://github.com/stolostron/backlog/issues/22061

Looks like there might not be searchClusters found sometimes so we need to fallback to `local-cluster`.


chenz4027 (approved):
lgtm, but @fxiang1 do we consider opening an issue to them given "might not be searchClusters found sometimes"
@RobDolares and I found that there's a bug in search, not sure if this is related
image

fxiang1:
> lgtm, but @fxiang1 do we consider opening an issue to them given "might not be searchClusters found sometimes" @RobDolares and I found that there's a bug in search, not sure if this is related image

@chenz4027 Thanks! I think if the subscription app didn't deploy to the managed clusters at all then search probably won't find any remote clusters. So in this case the behaviour is expected and there's not much search can do.

I saw that issue too that you and Robert found but I think it's not related to this issue.

chenz4027:
kk @fxiang1  thanks for the clarification.

PR #1679, opened 2022-05-11
Fix sync

chenz4027:
https://github.com/stolostron/backlog/issues/22475

The issue is 
- We used updateResource before
- As we changed that to patch, the before and after should be different, hence the issue

image


fxiang1 (approved):
LGTM!

PR #1688, opened 2022-05-13
Default resources with no cluster to local-cluster for YAML tab

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/22528

- Set resources with no cluster to local-cluster


chenz4027 (approved):
lgtm!

PR #1692, opened 2022-05-16
Read only lines for automation (installAttemptsLimit)

Randy424:

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -110,7 +110,7 @@
         "stream-browserify": "^3.0.0",
         "style-loader": "^3.2.1",
         "swr": "^1.0.1",
-        "temptifly": "^0.4.28",
+        "temptifly": "^0.5.13",
         "ts-import-plugin": "^1.6.7",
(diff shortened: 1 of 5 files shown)


frontend/package.json
```diff
@@ -110,7 +110,7 @@
         "stream-browserify": "^3.0.0",
         "style-loader": "^3.2.1",
         "swr": "^1.0.1",
-        "temptifly": "^0.4.28",
+        "temptifly": "^0.5.13",
```
KevinFCormier, line 113: Not sure we should be making this big of a jump in the 2.4.z stream. @jeswanke has that fix been backported to temptifly 0.4.z?

Randy424: I'll give that a try if @jeswanke hasn't already? Will add you both on the PR.

jeswanke: @Randy424 the 0.5.* is the temptifly for 2.5--you need to wait till I add the immutable stuff to 0.4.* before you can use in 2.4---there are differences between the two

PR #1700, opened 2022-05-18
Fix pulse status not resetting to green when set to red or yellow

fxiang1:

Issue: 
- https://github.com/stolostron/backlog/issues/22639

Change:
- Fix the node details table pulse status not resetting to green when a red resource was processed


chenz4027 (approved):
lgtm

PR #1711, opened 2022-05-24
Fix bare metal install on OCP 4.10

KevinFCormier:
https://github.com/stolostron/backlog/issues/22690


KevinFCormier:
Closing. Will only deliver if requested by customer.

PR #1715, opened 2022-05-26
Cluster status with unavailable add-ons

KevinFCormier:
https://github.com/stolostron/backlog/issues/22656


KevinFCormier:
Not pursuing this. There is an RFE to add a new column for add-on status separate from overall cluster status.

PR #1717, opened 2022-05-31
[WIP] Coverage merge test

KevinFCormier:
Testing if Sonar can merge coverage for same file from 2 different reports


KevinFCormier:
Closing. Simpler approach is to provide both files to Sonar directly.

PR #1719, opened 2022-06-01
Bug 2091625: [UI] BMA deprecation followups

ammont82:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=2091625
Add deprecation label and alert to BMA page:
image

### DIFF
--- a/frontend/src/routes/Infrastructure/BareMetalAssets/BareMetalAssetsPage.tsx
+++ b/frontend/src/routes/Infrastructure/BareMetalAssets/BareMetalAssetsPage.tsx
@@ -1,5 +1,4 @@
 /* Copyright Contributors to the Open Cluster Management project */
-
 import {
     BareMetalAsset,
     BareMetalAssetConditionReasons,
@@ -20,7 +19,15 @@ import {
     AcmRoute,
     AcmTable,
 } from '@stolostron/ui-components'
-import { ActionList, ActionListItem, Bullseye, ButtonVariant, PageSection, TextContent } from '@patternfly/react-core'
+import {
+    ActionList,
+    ActionListItem,
+    Alert,
+    Bullseye,
+    ButtonVariant,
+    PageSection,
+    TextContent,
+} from '@patternfly/react-core'
 import { fitContent } from '@patternfly/react-table'
 import { Fragment, useEffect, useState } from 'react'
 import { Trans, useTranslation } from '../../../lib/acm-i18next'
@@ -48,6 +55,8 @@ export default function BareMetalAssetsPage() {
             header={
                 <AcmPageHeader
                     title={t('bmas')}
+                    label={t('deprecated')}
+                    labelColor="orange"
                     titleTooltip={
                         <>
                             {t('bmas.tooltip')}
@@ -66,6 +75,23 @@ export default function BareMetalAssetsPage() {
         >
             <AcmPageContent id="bare-metal-assets">
                 <PageSection>
+                    <Alert
+                        title={t('bareMetalAsset.warning.title')}
+                        variant="warning"
+                        actionLinks={
+                            <a
+                                href={DOC_LINKS.CREATE_CLUSTER_ON_PREMISE}
+                                target="_blank"
+                                rel="noopener noreferrer"
+                                style={{ paddingLeft: 0 }}
+                            >
+                                {t('Learn more')}
+                            </a>
+                        }
+                        isInline
+                    >
+                        {t('bareMetalAsset.warning.text')}
+                    </Alert>
                     <BareMetalAssetsTable bareMetalAssets={bareMetalAssets} deleteBareMetalAsset={deleteResource} />
                 </PageSection>
             </AcmPageContent>
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -169,7 +169,7 @@
     "bareMetalAsset.tableHeader.namespace": "Namespace",
     "bareMetalAsset.tableHeader.role": "Role",
     "bareMetalAsset.tableHeader.status": "Status",
-    "bareMetalAsset.warning.text": "The process for creating bare metal clusters using the infrastructure provider is deprecated and will be removed in a future update. Using the assisted installer is best practice for installing bare metal clusters.",
+    "bareMetalAsset.warning.text": "The capability to create clusters with bare metal assets is deprecated, and will be removed in a future release. Use the assisted installation methods in the Red Hat Advanced Cluster Management console to create a new cluster.",
     "bareMetalAsset.warning.title": "Feature deprecation",
     "Base DNS domain": "Base DNS domain",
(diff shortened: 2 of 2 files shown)


matthewcarleton:
> @ammont82 If this is for 2.5.1, the z-stream is not open yet, but the PR will need to be against the `release-2.5` branch. I would also consider making the wording more similar to https://github.com/stolostron/console/blob/main/frontend/public/locales/en/translation.json#L172 which was reviewed by @cadawson from the doc squad.

It feels strange to say "assisted installer" where we don't use that in the UI. We only use assisted installation, that's why the text reads "assisted installation flows" in the suggested text here.

matthewcarleton:
> > > @ammont82 If this is for 2.5.1, the z-stream is not open yet, but the PR will need to be against the `release-2.5` branch. I would also consider making the wording more similar to https://github.com/stolostron/console/blob/main/frontend/public/locales/en/translation.json#L172 which was reviewed by @cadawson from the doc squad.
> > 
> > 
> > It feels strange to say "assisted installer" where we don't use that in the UI. We only use assisted installation, that's why the text reads "assisted installation flows" in the suggested text here.
> 
> Ah, this PR is actually changing the same text. So here is the evolution of this message:
> 
> 1. The Bare Metal Assets flow is deprecated and will be removed in the next feature. For Bare Metal clusters consider the assisted installation flows
> 2. The process for creating bare metal clusters using the infrastructure provider is deprecated and will be removed in a future update. Using the assisted installer is best practice for installing bare metal clusters.
> 3. (Proposed): The Bare Metal assets feature will be removed in the next release. For Bare Metal management consider infrastructure environments and the assisted installation creation flows.
> 
> * I agree if we don't use the term "assisted installer" in the UI we should use "assisted installation" instead.
> * I don't like the term "flow"... I don't think we ever use it in the UI or documentation. Similarly "feature" as in "Bare Metal assets feature" doesn't seem quite right.
I agree - I don't feel strongly about either.
> * Is it clear what "in the next release" means? I think we are referring to 2.6 but a customer might think 2.5.1 or 3.0.
I'd be happy to say 2.6 to be really specific.
> * Any reason to not explicitly say it is deprecated? Just because the badge in the UI already says that?
 Good idea - replace "removed" with "deprecated"
> * Looks like we need agreement on what should be capitalized and what shouldn't. For "Bare metal" and "Assisted installation," only the first letter appears capitalized in menus and headings, so I would guess that in this context, nothing should be capitalized? @cadawson can you confirm?

So could it be:
The Bare Metal assets capability will be deprecated in the 2.6 release. For bare metal management consider infrastructure environments and the assisted installation existing discovered hosts option.

KevinFCormier:
> So could it be:
The Bare Metal assets capability will be deprecated in the 2.6 release. For bare metal management consider infrastructure environments and the assisted installation existing discovered hosts option.

- It is already deprecated in 2.5, to be removed in 2.6. So I would just keep the Deprecated badge but say "removed" in the message.
- "Bare Metal" should be lower case?

matthewcarleton:
> > So could it be:
> > The Bare Metal assets capability will be deprecated in the 2.6 release. For bare metal management consider infrastructure environments and the assisted installation existing discovered hosts option.
> 
> * It is already deprecated in 2.5, to be removed in 2.6. So I would just keep the Deprecated badge but say "removed" in the message.
Ya that's right. 

> * "Bare Metal" should be lower case?
I'm not sure about that one. Let me run this past our content person and see what she thinks.

matthewcarleton:
> Hi folks. Sorry for the late response.
> 
> When something is deprecated, we don't normally mention when it will be removed. We don't always know, or the plans might change. That is why we we say "In a future release..."
> 
> "Bare metal" should follow normal capitalization. Lower case, unless it starts a sentence.
> 
> It is deprecated in 2.5, meaning it is not going to have updates anymore.
> 
> The "assisted installer" piece is a confusing one. For ACM, we refer to it as Central Infrastructure Management in the UI, I believe. The actual "assisted installer" function is in the OCP docs and area. I don't know if we want to just refer to the "assisted installer" even though it is OCP, or do we want to stick with CIM to be consistent. I am not sure which "assisted installation methods" are available in the cluster wizard, but that might be a good thing to mention, as you propose. Update: I see now where "assisted installation" comes from.
> 
> Your suggestions are good. Here is one more to consider:
> 
> The capability to create clusters with bare metal assets is deprecated, and will be removed in a future release. Use the assisted installation methods in the Red Hat Advanced Cluster Management console to create a new cluster.
> 
> Or similar? @KevinFCormier @matthewcarleton

@cadawson this sounds good to me!

jtomasek (approved):
/lgtm

frontend/src/routes/Infrastructure/BareMetalAssets/BareMetalAssetsPage.css
```diff
@@ -0,0 +1,11 @@
+.titleBMA_deprecated {
```
rawagner, line 1: this whole file is not needed if you use `label` instead of `description` for ACMPageHeader

frontend/src/routes/Infrastructure/BareMetalAssets/BareMetalAssetsPage.tsx
```diff
@@ -34,6 +42,7 @@ import { deleteResources } from '../../../lib/delete-resources'
 import { DOC_LINKS, viewDocumentation } from '../../../lib/doc-util'
 import { canUser, rbacDelete, rbacPatch } from '../../../lib/rbac-util'
 import { NavigationPath } from '../../../NavigationPath'
+import './BareMetalAssetsPage.css'
```
rawagner, line 45:
```suggestion

```

frontend/src/routes/Infrastructure/BareMetalAssets/BareMetalAssetsPage.tsx
```diff
@@ -48,6 +57,7 @@ export default function BareMetalAssetsPage() {
             header={
                 <AcmPageHeader
                     title={t('bmas')}
+                    description={<DeprecatedLabelWithText />}
```
rawagner, line 60:
```suggestion
                    label={t('deprecated')}
                    labelColor="orange"
```

frontend/src/routes/Infrastructure/BareMetalAssets/BareMetalAssetsPage.tsx
```diff
@@ -442,3 +469,14 @@ export function BareMetalAssetsTable(props: {
...
+export function DeprecatedLabelWithText() {
+    const { t } = useTranslation()
+    return (
+        <div className="titleBMA_deprecated">
+            <Label color="orange" style={{ margin: '0 0 0 5px' }}>
+                {t('deprecated')}
+            </Label>
+        </div>
+    )
+}
```
rawagner, line 482:
```suggestion

```

rawagner (changes requested):

jtomasek (approved):
/lgtm

PR #1732, opened 2022-06-07
App table to support OCP apps and Flux apps

chenz4027:


frontend/src/routes/Applications/Overview.tsx
```diff
@@ -184,6 +191,45 @@ export default function ApplicationsOverview() {
     const [placementRules] = useRecoilState(placementRulesState)
     const [namespaces] = useRecoilState(namespacesState)
 
+    // Openshift Application resources
+    const [cronJobs] = useRecoilState(cronJobsState)
+    const [daemonSets] = useRecoilState(daemonSetsState)
+    const [deployments] = useRecoilState(deploymentsState)
+    const [jobs] = useRecoilState(jobsState)
+    const [pods] = useRecoilState(podsState)
```
chenz4027, line 199: I still fetched pods because many pods have label of 'app' annotations so I think we should still display them in our table. We could discuss that FYI @fxiang1

fxiang1: The pods are created by resources like deployment, since you're already listing deployment resources it should already contain the same label as the pods.

The other problem is that for many openshift apps like `openshift-apiserver`, they use a different label for the pods(I'm not sure why):

The label for the `deployment` resource is `app=openshift-apiserver`:

image

but for the individual pods it's `app=openshift-apiserver-a`:

image

In this case it will list both `openshift-apiserver` and `openshift-apiserver-a` on the table but they are actually the same app.

chenz4027: Thanks for the explanation @fxiang1 , I'll remove the display of the pod instances given this behavior to avoid duplication

fxiang1 (approved):
LGTM! Thanks Maggie!

fxiang1 (approved):
LGTM! Thanks Maggie!

PR #1734, opened 2022-06-08
DO NOT MERGE - Cypress websockets

jamestalton:
Switching to [socket.io](http://socket.io/) instead of server side events for event streaming from backend to frontend.
[socket.io](http://socket.io/) uses websockets underneath (OCP is already using websockets, websockets are a little more mainstream than server side events.)
[socket.io](http://socket.io/) has auto reconnect logic and ping logic to test connections
[socket.io](http://socket.io/) has a fallback to long polling using REST api calls
Goal is to have Cypress tests that can be run against a live cluster and also a mock
Since [socket.io](http://socket.io/) allows a polling mode that uses REST and cypress allows mocking REST, we were able to get mocks of the [socket.io](http://socket.io/) events working


Randy424:
Linking the story here: https://github.com/stolostron/backlog/issues/23459

Randy424:
I patched the latest image from the PR to my cluster. The polling seems to trigger a loading state. And we see the loading state return every 45 seconds or so. It's currently disrupting the test work.

https://user-images.githubusercontent.com/21374229/176547293-a272aeac-fb0f-4958-bdfe-8a3b6b67d2f0.mov

PR #1739, opened 2022-06-09
No proxy note

jeswanke:
stolostron/backlog#17722

image


zhiweiyin318:
@KevinFCormier  looks good to me.  thanks.

cadawson (changes requested):
It looks good, John. I added an updated version that might be a little more clear. Thanks!

cadawson (changes requested):
It looks good, John. I added an updated version that might be a little more clear. Thanks!

PR #1749, opened 2022-06-13
Ansible operator install alert

Randy424:
https://github.com/stolostron/backlog/issues/13951

- Ansible operator installation warning now appears in the creation wizard (styled to resemble the same alert in the policy automation wizard).
- The hint rendered on the Automations page now only renders when the operator is not installed. 
- Unit testing has also been added for each case.

Screen Shot 2022-06-13 at 11 41 15 AM

Screen Shot 2022-06-13 at 11 45 03 AM

### DIFF
--- a/frontend/src/components/AcmDataForm.tsx
+++ b/frontend/src/components/AcmDataForm.tsx
@@ -102,6 +102,7 @@ export interface AcmDataFormProps {
     mode?: 'form' | 'wizard' | 'details'
     isHorizontal?: boolean
     edit?: () => void
+    operatorError?: ReactNode
 }
 
 function generalValidationMessage() {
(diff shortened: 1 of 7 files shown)


frontend/src/components/AcmDataForm.tsx
```diff
@@ -469,6 +474,11 @@ export function AcmDataFormWizard(props: {
             ),
             component: section.type === 'Section' && (
                 <Form isHorizontal={isHorizontal}>
+                    {operatorError && (
```
Randy424, line 477: One question I have about this alert, it currently does not prevent the the wizard from creating an automation template. GRC's alert also does not prevent creation. Is this the behavior we want?

PR #1760, opened 2022-06-16
Find token for ServiceAccount by type/annotation

KevinFCormier:
Starting in Kubernetes 1.24 (which OCP 4.11 uses), the ServiceAccount will no longer list the secret containing the service token, so we have to look through all the secrets to find it.


Randy424 (approved):
/lgtm

PR #1768, opened 2022-06-17
Remove recoilState to use search query

chenz4027:
https://github.com/stolostron/backlog/issues/22969

image


frontend/src/lib/search.ts
```diff
@@ -85,7 +85,25 @@ export function queryRemoteOCPAppResources(): IRequestResult<ISearchResult> {
...
+    return postRequest<SearchQuery, ISearchResult>(getBackendUrl() + apiSearchUrl, {
+        operationName: 'searchResult',
+        variables: {
+            input: [
+                {
+                    filters: [
+                        {
+                            property: 'kind',
+                            values: 'kustomization',
+                        },
```
chenz4027, line 106: I wonder if it's possible to filter label by its key, tried with regex /xxxx=*/ but failed... not sure if that's supported. For now I used filter in later code to filter resources with specific labels

chenz4027: Ah I see, thanks Feng! Guess we will have to it manually later then

fxiang1 (approved):
LGTM! Thanks Maggie!

PR #1782, opened 2022-06-23
Update jest to v28

jtomasek:
This change updates jest and related packages to version 28 which supports
multiple package exports
(https://nodejs.org/docs/latest-v12.x/api/packages.html#packages_exports)
which is used by openshift-assisted-ui-lib.
Related jest issue: https://github.com/facebook/jest/issues/9771

Example of importing from 'openshift-assisted-ui-lib/cim' with correct types reference:
Screenshot 2022-06-23 at 8 46 22


rawagner:
great update @jtomasek !

/lgtm

PR #1784, opened 2022-06-23
Migrate temptifly into console#23472

jeswanke:


frontend/src/routes/Infrastructure/InfraEnvironments/CreateInfraEnv.tsx
```diff
@@ -6,17 +6,15 @@ import { useState } from 'react'
...
 import { NavigationPath } from '../../../NavigationPath'
 import infraEnvTemplate from './infraenv-template.hbs'
 import InfraEnvForm, { Portals } from './InfraEnvForm'
 
 // include monaco editor
 import 'monaco-editor/esm/vs/editor/editor.all.js'
 import 'monaco-editor/esm/vs/basic-languages/yaml/yaml.contribution.js'
-import '../Clusters/ManagedClusters/CreateCluster/style.css'
-import 'temptifly/dist/styles.css'
+import '../../Applications/CreateApplication/Subscription/style.css'
```
KevinFCormier, line 17: What is the reason for this change? The previously reference stylesheet is still used elsewhere.

jeswanke: that style.css is used to position the template editor --- but the one in createcluster is good for when it's a wizard while the one in app subscription is good for when it's a form--and infra is using it as a form

PR #1795, opened 2022-06-28
Filter non-insight sourced violations from PolicyReports

zlayne:

### DIFF
--- a/frontend/src/routes/Home/Overview/OverviewPage.tsx
+++ b/frontend/src/routes/Home/Overview/OverviewPage.tsx
@@ -294,7 +294,7 @@ export default function OverviewPage() {
             let policyReportModerateCount = 0
             let policyReportLowCount = 0
             policyReportsForSelectedClusters.forEach((policyReport: PolicyReport) => {
-                policyReport.results.filter((result: PolicyReportResults) => {
+                policyReport.results.forEach((result: PolicyReportResults) => {
                     if (result.source === 'insights') {
                         switch (result.properties.total_risk) {
                             case '4':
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Home/Overview/OverviewPage.tsx
```diff
@@ -292,7 +292,8 @@ export default function OverviewPage() {
             const policyReportCriticalCount = policyReportsForSelectedClusters.reduce(
                 (total: number, currentValue: PolicyReport) => {
                     const criticalResults = currentValue.results.filter(
-                        (result: PolicyReportResults) => result.properties.total_risk === '4'
+                        (result: PolicyReportResults) =>
+                            result.source === 'insights' && result.properties.total_risk === '4'
                     )
                     return total + criticalResults.length
```
KevinFCormier, line 298: I would suggest adding a function to create this reducer function, with an argument for the total_risk value, rather than repeating it 4 times.

zlayne: Good idea - I changed it a little bit to reduce the number of loops we were previously running.

frontend/src/routes/Home/Overview/OverviewPage.tsx
```diff
@@ -289,42 +289,30 @@ export default function OverviewPage() {
...
-                    return total + lowResults.length
-                },
-                0
-            )
+            let policyReportCriticalCount = 0
+            let policyReportImportantCount = 0
+            let policyReportModerateCount = 0
+            let policyReportLowCount = 0
+            policyReportsForSelectedClusters.forEach((policyReport: PolicyReport) => {
+                policyReport.results.filter((result: PolicyReportResults) => {
```
KevinFCormier, line 297: This should be a forEach now rather than a filter, I think?

PR #1802, opened 2022-07-01
Fix duplicate Argo apps displayed as OCP apps

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/22974

- Fix duplicate Argo apps which is also being displayed as OCP apps


chenz4027 (approved):
lgtm

PR #1803, opened 2022-07-02
Added fields for vSphere Folder and ResourcePool

briantward:
Adds the vSphere folder and vSphere resource pool variables that are found in the `install-config.yaml` to the ACM UI.

https://docs.openshift.com/container-platform/4.10/installing/installing_vsphere/installing-vsphere-installer-provisioned-customizations.html#installation-installer-provisioned-vsphere-config-yaml_installing-vsphere-installer-provisioned-customizations

https://github.com/stolostron/backlog/issues/26159


briantward:
I've tested this with expectations that it populates over to the install-config.yaml correctly; however, I've not personally tested it end to end with a custom folder and resource pool.  Also, I'd like to confirm that it works correctly with spaces in object names.

I'm interested in feedback from the development team.  We have a customer looking for this feature, and we'll open a BZ for it shortly.

Oh, and I did notice after the fact that install-config.yaml also offers another field we are missing: diskType.  That might be a quick addition.

frontend/src/routes/Credentials/CredentialsForm.test.tsx
```diff
@@ -207,6 +207,8 @@ describe('add credentials page', () => {
                 cluster: 'cluster',
                 datacenter: 'datacenter',
                 defaultDatastore: 'defaultDatastore',
+                folder: 'folder',
+                resourcePool: 'resourcePool'
```
KevinFCormier, line 211: Thanks for the contribution! Please let me know when a Bugzilla is opened and if you'd like to continue working on this PR or if you'd like us to put the finishing touches on it ourselves. Tests failed with:
```
[frontend] [prettier] [error] src/routes/Credentials/CredentialsForm.test.tsx: SyntaxError: ',' expected. (212:17)
[frontend] [prettier] [error]   210 |                 folder: 'folder',
[frontend] [prettier] [error]   211 |                 resourcePool: 'resourcePool'
[frontend] [prettier] [error] > 212 |                 imageContentSources: '',
[frontend] [prettier] [error]       |                 ^
[frontend] [prettier] [error]   213 |             },
[frontend] [prettier] [error]   214 |             true
[frontend] [prettier] [error]   215 |         )
```

KevinFCormier:
@cadawson Could you please review the changes to `frontend/public/locales/en/translation.json` when you have a moment - or delegate to appropriate person on docs team. Thanks!

frontend/public/locales/en/translation.json
```diff
@@ -1227,6 +1227,8 @@
     "Enter your vSphere cluster name": "Enter your vSphere cluster name",
     "Enter your vSphere datacenter": "Enter your vSphere datacenter",
     "Enter your vSphere default defaultDatastore": "Enter your vSphere default defaultDatastore",
+    "Enter your vSphere folder absolute path": "Enter your vSphere folder absolute path",
```
cadawson, line 1230: Could also use: "Enter the absolute path to your vSphere folder": "Enter the absolute path to your vSphere folder"

frontend/public/locales/en/translation.json
```diff
@@ -1227,6 +1227,8 @@
...
+    "Enter your vSphere resource pool absolute path": "Enter your vSphere resource pool absolute path",
```
cadawson, line 1231: Similarly: "Enter the absolute path to your vSphere resource pool": "Enter the absolute path to your vSphere resource pool"

cadawson (changes requested):
A couple of suggestions, but looks good. Thanks!

KevinFCormier:
> @KevinFCormier thanks for the progress on this rfe/pr. I see you caught that hive ClusterDeployment doesn't have the resourcepool field. What implication would that have downstream, since (it appears to me) the console is also generating its own install-config.yaml ?

@briantward If we add `resourcePool` to the `ClusterDeployment`, it is not saved in the resource. But we are still putting it in the `install-config.yaml` which Hive will use during provisioning, so I think it should work. 

Does that answer your question? I don't really understand your reference to "downstream" here.

briantward:
> > @KevinFCormier thanks for the progress on this rfe/pr. I see you caught that hive ClusterDeployment doesn't have the resourcepool field. What implication would that have downstream, since (it appears to me) the console is also generating its own install-config.yaml ?
> 
> @briantward If we add `resourcePool` to the `ClusterDeployment`, it is not saved in the resource. But we are still putting it in the `install-config.yaml` which Hive will use during provisioning, so I think it should work.
> 
> Does that answer your question? I don't really understand your reference to "downstream" here.

@KevinFCormier that answers my question. I only meant downstream execution logic. I could also open an RFE to update the ClusterDeployment API to be in sync with install-config, even if it isn't strictly needed here. I'm not sure how much it matters to that team or other integrations.

jeswanke (approved):

PR #1805, opened 2022-07-04
Rename infra environments to host inventory

rawagner:
as described in the doc https://docs.google.com/document/d/1WGGjqC_fzrWf1cUigu8pa6hBwcf__3tW5R-UHqyhrms , we want to update the naming as the old one is too confusing


frontend/src/routes/Infrastructure/InfraEnvironments/HostInventoryBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { AcmButton } from '@stolostron/ui-components'
+import { ActionGroup, ButtonVariant, Card, CardBody, CardFooter, CardTitle, StackItem } from '@patternfly/react-core'
+import { useState } from 'react'
+import { ExternalLinkAltIcon } from '@patternfly/react-icons'
+import { useTranslation } from '../../../lib/acm-i18next'
+import { DOC_LINKS } from '../../../lib/doc-util'
+
+const localStorageKey = `HostInventoryBannerDismissed`
+// t('Hosts are discovered and stored in infrastructure environments. You can manage your host inventory with infrastructure environments which allow to set the same configurations for all hosts that it contains.')
```
mareklibra, line 11: nit: leftover

frontend/public/locales/en/translation.json
```diff
@@ -1517,6 +1520,7 @@
     "managedClusterSet.form.removed": "Removed",
     "managedClusterSet.form.transferred": "Transferred",
     "managedClusterSet.form.unchanged": "No change",
+    "Managing hosts with infrastructure environments makes it easy to set settings across multiple hosts.": "Managing hosts with infrastructure environments makes it easy to set settings across multiple hosts.",
```
mareklibra, line 1523: opinion: I understand the benefits of having "real texts" as keys for translations but I would shorten them for similarly long cases

rawagner: I disagree.
How do we say which key is too long ?
Also it does not really matter how this file looks like, its not something we would be looking at. The important part, which needs to be readable, is in the actual code in which case a using a real text is a clear winner over some key

mareklibra (dismissed):

PR #1807, opened 2022-07-04
Fix route for local-cluster for 2.5.2

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/23531

- Fix route for local-cluster in 2.5.2

Found that using ManageClusterView still works for local-cluster but it's not as reliable as get the resource straight from kube.


fxiang1:
Missed 2.5.1 will hold and wait for 2.5.2.

chenz4027 (approved):
lgtm

PR #1818, opened 2022-07-08
Use console backend to display user saved searches

zlayne:

Related issue: https://github.com/stolostron/backlog/issues/23869


frontend/src/resources/userpreference.tsx
```diff
@@ -0,0 +1,90 @@
...
+                    name: username,
+                },
+                spec: {
+                    savedSearches,
+                },
+            })
+        })
+        .catch((error) => {
+            // eslint-disable-next-line no-console
+            console.error(error)
```
KevinFCormier, line 54: Did you intend to keep these?

KevinFCormier (commented):
Left one question inline about the console logs. Also, please review the code smells. I am not concerned if they are in test cases, but there might be some there we should fix.

frontend/src/resources/userpreference.tsx
zlayne: I was planning to include them yes. And I will comb through the code smells - thanks.

PR #1836, opened 2022-07-13
Sonar coverage

dhaiducek:
Resolves PR coverage reporting since `lcov` uses a relative path inside of `frontend/` where the `jest` commands are run but SonarCloud is scanning from the project root so it can't find the files that `lcov` is referencing:
```
15:06:52.989 DEBUG: Using 'frontend/coverage/lcov.info' to resolve LCOV files
15:06:52.989 INFO: Analysing [/go/src/github.com/stolostron/console/frontend/coverage/lcov.info]
15:06:53.377 WARN: Could not resolve 9 file paths in [/go/src/github.com/stolostron/console/frontend/coverage/lcov.info]
15:06:53.378 DEBUG: Unresolved paths:
src/atoms.tsx
```

Specifically, this adds a `projectRoot` argument to specify the root directory. In order to implement that, I stripped the CLI arguments from the `package.json` file so that settings in the `jest.config.*` file wouldn't be overridden.

I also swapped out the `jest.config.js` for a `jest.config.ts` since I was playing around trying to get it to work, but I left this change since TypeScript seems to be preferred in this repo and it's a nicer experience to update.


dhaiducek:
@KevinFCormier So. The messages in the SonarCloud logs that I thought were causing a problem are actually ignored files, so the "unresolved files" message is expected. Oops. 😬 

Still, it seems like this PR could still be merged since unnecessarily overriding the `jest.config.*` file with CLI arguments in `package.json` could cause confusion in the future.

Feel free to unhold it or let me know if any of the configurations I migrated to `jest.config.ts` should be reverted or moved back to CLI arguments (see the [Jest docs](https://jestjs.io/docs/configuration) for details).

KevinFCormier:
@dhaiducek It looks neater the way you have it.

PR #1846, opened 2022-07-15
Logout instead of reloading when authentication expires

KevinFCormier:
https://github.com/stolostron/backlog/issues/23543


KevinFCormier:
Closing - will address in 2.5.2+

PR #1851, opened 2022-07-18
Fix log viewer crash for pod deploy failures 2.5

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/24370

- Add check for podModel with length of 0

The UI doesn't crash anymore:
image


jamestalton (approved):
lgtm

PR #1859, opened 2022-07-20
Change path of locales for standalone app

rawagner:
The old 'multicloud' path clashes with backend logic which is removing 'multicloud' path from every request


jtomasek (approved):

rawagner:
closing in favor of https://github.com/stolostron/console/pull/1863

PR #1864, opened 2022-07-20
Add warning on policies that will auto-prune

JustinKuli:
This feature is new, and it seems helpful to warn users when deleting
a policy may cause (potentially) unexpected behavior on the managed
clusters.

Refs:
 - https://github.com/stolostron/backlog/issues/24029

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>


heyethankim:
> Here's what the draft wording looks like:
> 
> Screen Shot 2022-07-20 at 3 03 41 PM
> 
> (By the way, I feel like my buttons and stuff look a little different than what I see in the "real" cluster - I might have some weird CSS issue in my environment?)

The same comment applies here regarding "Remove" vs. Delete." "Remove" is more like "Set it aside" rather than "Erase it." Using both words could be confusing. My understanding is that we would like to "Erase it," not "Put it aside." 

The button style should be as follows: 
image

frontend/public/locales/en/translation.json
```diff
@@ -1062,6 +1062,7 @@
     "deleting": "Deleting",
     "Deleting": "Deleting",
     "Deleting <italic>{{policyAutomation}}</italic> is irreversible. Any associated Ansible job will be deleted.": "Deleting <italic>{{policyAutomation}}</italic> is irreversible. Any associated Ansible job will be deleted.",
+    "Deleting this policy may remove some related objects on the managed cluster(s).": "Deleting this policy may remove some related objects on the managed cluster(s).",
```
dockerymick, line 1065: Deleting this policy might delete some related objects on the managed cluster(s).

frontend/public/locales/en/translation.json
```diff
@@ -1948,6 +1949,7 @@
     "Show only:": "Show only:",
     "show.less": "Show less",
     "show.more": "{{number}} more",
+    "Some policies have a Prune Object Behavior set": "Some policies have a Prune Object Behavior set",
```
dockerymick, line 1952: Some policies have the `prune` parameter set.

Not sure what we might agree for the parameter name

frontend/public/locales/en/translation.json
```diff
@@ -1948,6 +1949,7 @@
...
     "Some resources failed to deploy. Use View resource YAML link below to view the details.": "Some resources failed to deploy. Use View resource YAML link below to view the details.",
```
dockerymick, line 1953: Some resources failed to deploy. Use the _View resource YAML_ link to view details. 

Remove the words like below, above, left, right, etc.

dockerymick (changes requested):

frontend/public/locales/en/translation.json
JustinKuli: I agree, @heyethankim, do you have any new thoughts on the name?

dockerymick (dismissed):
See the changes, thanks for accepting my suggestions! Still curious about the potential name for the feature, otherwise this looks good to me

JustinKuli:
Updated screenshot:

Screen Shot 2022-07-22 at 10 51 54 AM

frontend/public/locales/en/translation.json
heyethankim: Well, I like your name suggestion. 😀 Would you like to bring it up during the GRC security meeting on Monday?

RobDolares (approved):
/lgtm

PR #1871, opened 2022-07-21
Do not cache product version

KevinFCormier:
During update, version changes after pods are running

https://github.com/stolostron/backlog/issues/24247


Randy424 (approved):
/lgtm

PR #1878, opened 2022-07-21
Use regex to check version

KevinFCormier:
https://github.com/stolostron/backlog/issues/24442

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
@@ -580,27 +580,32 @@ export const architectureData = [
     },
 ]
 
+const versionRegex = /release:([\d]+)\.([\d]+)\.([\d]+)/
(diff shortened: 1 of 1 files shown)


Randy424 (dismissed):
/lgtm

Cool function and definitely preferable to what was going on before. I'm curious though, why include an explicit reference to base 10 radix `parseInt(matches[1], 10)`, for match[1] and not for match[2]? I was messing around with the function in a JS playground, and they seem to behave the same...leads me to think that the default value for that parameter is 10?

So we might not need that first 10. Just my '10' cents, though. Still looks great!

KevinFCormier:
@Randy424 I meant to supply the `10` radix parameter on both calls. That's what the flagged "bug" is... basically, in some older browsers, it might try interpret the number as octal if it starts with 0 or hex if it starts with 0x.
https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=1878&id=open-cluster-management_console&open=AYIiaqe1LF7m9kplp61o

There's also a security bug reported because of the regex... trying a version that should be less susceptible to DOS attack.

Randy424 (approved):
/lgtm

PR #1890, opened 2022-07-22
Fix issue unable to delete placementrule with same name in different namespaces 2.6

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/24495

This issue is caused by using the same placementrule but in difference namespaces. The code is not checking the namepace so placementrules with the same name but different namespace will be associated with the wrong subscription

- Add a namespace check when search for placementrules

With the fix, the warning is now gone and users can delete the placementrule:
image


chenz4027 (approved):
lgtm! Thanks Feng

PR #1893, opened 2022-07-25
Add explanatory text for pending import

KevinFCormier:
Delay before redirect and add alert

https://github.com/stolostron/backlog/issues/23457


frontend/public/locales/en/translation.json
```diff
@@ -1370,8 +1370,9 @@
     "import.command.generated": "1. Copy this command",
     "import.command.launchconsole": "Launch console",
     "import.command.pendingimport": "Cluster is pending import",
+    "import.command.pendingimport.message": "If you imported with server URL and API token or kubeconfig, please wait for the page to refresh with an update. If it fails or you suspect a failure, you can run a command to attempt manual import instead.",
```
KevinFCormier, line 1373: @cadawson How does this string look?

I incorporated your other suggested change in this PR.

KevinFCormier: Good point about suspecting a failure. I don't want to go into detail about how they know it has failed... so perhaps it would be better to suggest they can run the command if the import does not complete. Also, I think "update and refresh" is redundant. How about:

"If you imported by using the server URL and API token, or by using the kubeconfig file, wait for the import to complete. If the import does not complete, run the import command to try a manual import."

cadawson: I like it. 

One thing that I just noticed that we could also use for the last sentence, if you like this better: "If the import does not complete, try importing manually by running the import command."

Either one, I think will work.

Randy424 (approved):
/lgtm

PR #1934, opened 2022-08-01
wipes out any YAML changes install-config in Create Cluster flow#18661

jeswanke:
does https://github.com/stolostron/backlog/issues/18661

this fixes a few issues related to syncing form changes with yaml editing

========================================================
**issue 1-- user edits install-config tab, then changes anything in form (clicking secrets checkbox is a form change)**
**TemplateEditor**

  **handleParse**-- in case there are form changes after the yaml tab has been edited, two versions of the yaml are saved:
     1. the templateYAML is saved as base--this is the last form generated version of the yaml before editing started
     2. what the user is currently typing is saved as typingYAML
     
  **handleControl**-- if the user had been typing in the tab, typingYAML is saved as typedYAML

**refresh-source-from-stack**
  **mergeOtherYAMLTabs** -- 
     1. if the user is still typing on the tab, the typingYAML is encoded and stuffed into the install-config.yaml on the main tab
     2. if the user has stopped typing and a form change has occurred, the base/typed/current yamls are merged and that's 
         encoded into the main tab
             (merging works by taking the changes between the currently generated yaml and the last base template and applying
               those changes to the typedYAML, which ten becomes the mergedTemplate)

**TemplateEditor**
   handleParse-- if the user then starts typing again:
      1. the base template becomes the last merged template
      2. and the current typing is saved in typingYAML

===========================================================

**issue 2 -- SyncEditor utility (reconcile.ts/mapResources) wasn't accurately mapping between resources to merge**
merging works by getting the diff between two form generated yamls and applying those changes to the user edited yaml
with the install-config tab there was just one resource to merge
but most editors have multiple resources (ClusterDeployment/etc), so base/typed/current yamls will have multiple resources
so in order to do the merge, the editor has to find the same resource in the base/types/current
  then merge that resource 
  then repeat for each remaining resource
ideally each resource has the same name and kind to make the match--however if the resource has no name yet or the user has changed the name, the SyncEditor falls back on this strategy:
   1. all resources are grouped by their kind (ex: all Secrets)
   2. each resource is stringified and the npm package stringSimilarity is used to find the closest matching similar resource
two things were fixed for this issue:
  1.  in reconcile.ts, when stringifying the resources, encoded yaml was NOT removed from the resource which caused the
        matching to fail
  4. temptifly did this matching with a different algorithm -- it now uses the same mapResources algorithm as SyncEditor

=======================================================

**issue 3 -- SyncEditor--editing policies, adding an item to an array**
two things fixed:
1. when you started to add an item to an array (ex: within array, you typed a dash:
         array:
            - item1
            -
   a null would appear ( - null )
    this is fixed in process.ts
2. when you finally entered the new item, the whole array would be decorated as if it were new and not just the new item
    this is fixed in changes.ts

====================================================

**issue 4-- SubscriptionApplication was sluggish**
in temptifly, when user selects Git or Helm or Object Storage, it does a cloneDeep of the controlData for that selection and
     inserted it into current controlData
if there's a custom controlData type, it was also cloning the 'component' data member (the React component that's rendered in the wizard/form)-- which is slow and unnecessary
a new cloneControlData method was created to omit any 'component' member


Randy424 (approved):
/lgtm

PR #1943, opened 2022-08-02
Bump `react-core` version to include OUIA fix

dhaiducek:
Bump to include:
- https://github.com/patternfly/patternfly-react/commit/d8f546344a8a361a73796096a1de7d4cd44fbcff

Addresses:
- https://github.com/stolostron/backlog/issues/24658

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -5972,28 +5972,33 @@
             }
         },
         "@patternfly/react-core": {
-            "version": "4.221.11",
-            "resolved": "https://registry.npmjs.org/@patternfly/react-core/-/react-core-4.221.11.tgz",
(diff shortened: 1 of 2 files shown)


dhaiducek:
This fixes a case where the GRC page was crashing when anything was entered into the table search box.

jamestalton:
PatternFly broke the Select component in that version.
I have submitted an issue to PatternFly. https://github.com/patternfly/patternfly-react/issues/7757
If we bump the core, it will break the argo cd application subscription.

In addition... remember that ACM stand-alone uses one version of PatternFly, but when in the context of OCP as a dynamic plug, we use the version of PatternFly that OCP gives ACM. Meaning that 4.10, 4.11, and 4.12 will get different versions of PatternFly. :disappointed:

Really need to have some Cypress tests that run weekly against all those versions.

PR #1951, opened 2022-08-03
Add infrastructure and control plane data to review step

fxiang1:

Issue: https://github.com/stolostron/backlog/issues/24789

- Add `reviewinfo` type to temptifly for data that we want to show in the review step but not in the generated YAML
- Add infrastructure and control plane type(if applicable) to all the different cluster creation flows


chenz4027 (approved):
lgtm

PR #1962, opened 2022-08-05
BZ 2113886: Handle OCP console link when running as dynamic plugin

rawagner:


jtomasek (approved):
/lgtm

PR #1965, opened 2022-08-05
change helper text

chenz4027:


fxiang1 (changes requested):
@chenz4027 Wondering if we can keep the word `Choose` so it's consistent with the other pages? ie: `Choose a control plane type for you cluster.`
image

fxiang1 (approved):
LGTM! Thanks Maggie!

PR #1978, opened 2022-08-09
Set subrow key to avoid crash

dhaiducek:
Code in `react-core` using `UseMemo` to fetch an OUIA ID was causing the UI to crash. This uses the parent row to set the `key` (since the subrow wouldn't necessarily have the same data as its parent for the `keyfn` to return a value).

This PR is an alternative to the package update proposed here:
- https://github.com/stolostron/console/pull/1943

Addresses:
- https://github.com/stolostron/backlog/issues/24658

Closes #1943


RobDolares (approved):
/lgtm

jamestalton (approved):
lgtm

PR #1980, opened 2022-08-09
Fix catalog crash when opening tooltip

fxiang1:

Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2116971

- Updated control plane to use new tooltip format in react-data-view
cpt


chenz4027 (approved):
lgtm!

PR #1990, opened 2022-08-10
change violation filters to cluster level

willkutler:
Signed-off-by: Will Kutler <wkutler@redhat.com>

for https://github.com/stolostron/backlog/issues/24946

Previously, policies that were applied on multiple managed clusters could only be filtered by the overall compliance state, which seemed to conflict with the compliance status shown on the policies page. This PR changes the filters so that if a policy has different statuses on different managed clusters, it will show up in the filter for each cluster's compliance state.


dhaiducek (approved):
LGTM! Thanks for the update, @willkutler!

PR #2041, opened 2022-08-24
2.7 add credential wizard

chenz4027:
https://issues.redhat.com/browse/ACM-1640

What's implemented:
- create credential modal in subscription wizard, automation wizard, create cluster wizard and create clusterpool wizard
- Auto select the newly created credential in everywhere except subscription wizard due to an issue where onControlInitialize only works for parent controldata. Will create another PR to add the equivalent behavior once that's added in Temptifly FYI @jeswanke 

image


fxiang1 (approved):
LGTM! Thanks Maggie!

PR #2048, opened 2022-08-25
ignore install-yaml secret

jeswanke:

https://github.com/stolostron/backlog/issues/25307

### DIFF
--- a/frontend/src/components/TemplateEditor/TemplateEditor.js
+++ b/frontend/src/components/TemplateEditor/TemplateEditor.js
@@ -1202,7 +1202,7 @@ export default class TemplateEditor extends React.Component {
                         metadata: { name, namespace },
                         data,
                     } = resource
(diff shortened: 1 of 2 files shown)


frontend/src/components/TemplateEditor/TemplateEditor.js
```diff
@@ -1200,8 +1200,9 @@ export default class TemplateEditor extends React.Component {
                     const {
                         kind,
                         metadata: { name, namespace },
+                        data,
                     } = resource
-                    if (kind === 'Secret') {
+                    if (kind === 'Secret' && (!data || !data['install-config.yaml'])) {
```
KevinFCormier, line 1205: I think the logic would be easier to read this way:
```suggestion
                    if (kind === 'Secret' && !(data && data['install-config.yaml'])) {
```
Or if something like this works in this context:
```suggestion
                    if (kind === 'Secret' && !data?.['install-config.yaml']) {
```

PR #2058, opened 2022-08-29
Fix credentials copy during cluster create

KevinFCormier:
https://github.com/stolostron/backlog/issues/25428


fxiang1 (approved):
Looks great Kevin! Like how you changed it so we don't need to return a value for setAvailableConnections

PR #2066, opened 2022-08-30
More readable search errors

zlayne:

Related issues: 
- https://github.com/stolostron/backlog/issues/24788
- https://github.com/stolostron/backlog/issues/22391


frontend/public/locales/en/translation.json
```diff
@@ -89,6 +89,7 @@
     "Already enforcing": "Already enforcing",
     "Already informing": "Already informing",
     "Amazon Web Services": "Amazon Web Services",
+    "An error occurred while contacting the search service.": "An error occurred while contacting the search service.",
```
jlpadilla, line 92: For consistency, I see that most other errors don't start with "An ..."

frontend/src/routes/Home/Search/SearchPage.tsx
```diff
@@ -48,29 +49,39 @@ const useStyles = makeStyles({
...
-                    }
-                    subtitle={schemaError?.message || completeError?.message}
-                />
-            </div>
+            <EmptyState>
+                <EmptyStateIcon icon={InfoCircleIcon} color={'var(--pf-global--info-color--100)'} />
+                <Title size="lg" headingLevel="h4">
+                    {t('search.filter.info.title')}
+                </Title>
+                <EmptyStateBody>{schemaError?.message || completeError?.message}</EmptyStateBody>
```
jlpadilla, line 62: Do we always have one of these 2 error messages? I wonder if we should add `... || 'Default empty state message'`

zlayne: We will always have one or the other here - the conditional above checks that at least one is defined.

PR #2073, opened 2022-09-06
testing debug flag --foreground-scripts

Randy424:

Testing install flag on npm ci to reveal error causing failure from cb() never called 
`npm ci --no-optional --foreground-scripts`

source - https://github.com/npm/statusboard/issues/395#issuecomment-1003256492


KevinFCormier:
@Randy424 Good thought, but it looks like `--foreground-scripts` is new in npm 7.x, and we are on 6.x.
https://docs.npmjs.com/cli/v7/using-npm/config#foreground-scripts

PR #2089, opened 2022-09-08
Migrate wizards

RobDolares:
related task: https://github.com/stolostron/backlog/issues/23471


frontend/src/lib/validation.ts
```diff
@@ -287,3 +290,77 @@ export function validateNoProxy(value: string, t: TFunction) {
...
+//         for (const char of value) {
+//             if (!lowercaseAlphaNumericCharacters.includes(char) && char !== '-')
+//                 return `${t('validate.kubernetesDnsName.char')}`
+//         }
+//         if (!lowercaseAlphaNumericCharacters.includes(value[0])) return `${t('validate.kubernetesDnsName.startchar')}`
+//         if (!lowercaseAlphaNumericCharacters.includes(value[value.length - 1]))
+//             return `${t('validate.kubernetesDnsName.endchar')}`
+//     }
+//     return undefined
+// }
```
KevinFCormier, line 306: Looks like these were already ported over? Can you confirm the logic is the same and remove commented portions from this file if they are not needed?

frontend/src/lib/validation.ts
```diff
@@ -200,13 +203,13 @@ export function validateCloudsYaml(yamlValue: string, cloudValue: string, t: TFu
...
-export function validateWebURL(url: string, t: TFunction, supportedProtocols?: string[]) {
-    const protocols = supportedProtocols ? supportedProtocols : ['http', 'https']
+export function validateWebURL(url: string, _item: unknown, t?: TFunction) {
+    t = t ? t : (value) => value
     if (
         validator.isURL(url, {
             require_protocol: true,
             require_valid_protocol: true,
-            protocols: protocols,
+            protocols: ['http', 'https'],
```
KevinFCormier, line 212: Why did you change this? The credentials form specifically wants to allow only `https` but this hard-codes it to include `http`.

RobDolares: i recall a merge conflict a while back I think i meant to revert then.  Reverted now.

frontend/src/lib/validation.ts
RobDolares: confirmed and removing this commented out validation.  Leaving the additional validations(RFC1123 and RFC 1135) commented in case they are used in the future.

PR #2097, opened 2022-09-13
Update Openstack OS Image validation, remove unused code

Randy424:
regarding: https://github.com/stolostron/backlog/issues/25376

- Baremetal os image validation has been removed.
- Unused transaltion strings and testing has been removed.
- Given that the possible input for the clusterOSImage is rather nebulous (i.e. it can be a URL with or without a SHA checksum or a "glance image name", which I haven't actually been able to find constraints for), I have not replaced the validation function.


KevinFCormier (approved):
I also could not find any information about valid formats for Glance image names, so I agree with removing the validation. Looks good. Thanks, Randy.

PR #2099, opened 2022-09-14
Bug 2124581: Add alert to Host Inventory card when no OpenShift images are available

jtomasek:
When installing a cluster using host inventory, if there are no clusterImageSets available, user is not able to select and OpenShift version and gets blocked in the cluster creation wizard. This change disables the Host Inventory card and adds hint on how to add release images.
Screenshot 2022-09-14 at 12 00 59

Fixes BZ2124581


frontend/public/locales/en/translation.json
```diff
@@ -1584,6 +1584,8 @@
     "No PolicySets found": "No PolicySets found",
     "No PolicySets match the current search filter": "No PolicySets match the current search filter",
     "No proxy": "No proxy",
+    "No release image is available. Follow cluster creation prerequisite documentation to learn how to add release images.": "No release image is available. Follow cluster creation prerequisite documentation to learn how to add release images.",
```
rawagner, line 1587: this does not seem to be used anywhere

rawagner (changes requested):

frontend/public/locales/en/translation.json
jtomasek: Here https://github.com/stolostron/console/pull/2099/files#diff-77f63e21935d0f15393b4c046cd3ba5ec42d3941b0aaa039ee7e2b45f85c9d9cR108

PR #2102, opened 2022-09-14
Display deploy errors in topology

fxiang1:

Issues:
- https://github.com/stolostron/backlog/issues/23083
- https://github.com/stolostron/backlog/issues/25559

We can't fully support displaying status from SubscriptionStatus without implementing this story(https://github.com/stolostron/backlog/issues/22348) because getting all SubscriptionStatus from all clusters using ManagedClusterView is not scalable and would cause performance issues. Instead I'm just looking at the SubscriptionReport to see if there are failures and directing the user to the SubscriptionStatus YAML via search.

image


chenz4027 (approved):
lgtm

PR #2103, opened 2022-09-15
change error button to link with relative path

Randy424:
regarding: https://github.com/stolostron/backlog/issues/25732

### DIFF
--- a/frontend/src/lib/error-output.tsx
+++ b/frontend/src/lib/error-output.tsx
@@ -1,7 +1,7 @@
 /* Copyright Contributors to the Open Cluster Management project */
-import { Button, ButtonVariant } from '@patternfly/react-core'
 import { ExternalLinkAltIcon } from '@patternfly/react-icons'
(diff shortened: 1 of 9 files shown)


frontend/src/lib/error-output.tsx
```diff
@@ -13,19 +13,10 @@ export function getOperatorError(configMaps: ConfigMap[], isOperatorInstalled: b
                 {openShiftConsoleUrl && openShiftConsoleUrl !== '' ? (
                     <div>
                         {t('Install the Operator through the following link: ')}
```
KevinFCormier, line 15: Since we don't use `openShiftConsoleUrl` and in 2.7 we know we are running inside the OCP console, we should be able to remove this. Do we use the `console-public` ConfigMap anywhere else? If not, I think there might be a watch for it we can remove.

RobDolares (approved):
/lgtm

PR #2112, opened 2022-09-20
Fix searchbar dividers

zlayne:

Need to change the divider orientation definition due to PatternFly version differences in MCE and standalone.


RobDolares (approved):
/lgtm

PR #2117, opened 2022-09-22
Add support for project resources in topology

fxiang1:

Issue:
https://github.com/stolostron/backlog/issues/25662

Currently search doesn't support finding project resources because the UID conflicts with namespace resources(in OCP project is essentially a copy of namespace). This fix is more like a workaround due to the search limitation.

- Added support for project resources by looking for the corresponding namespace resource and attaching the namespace results to the project node

image


chenz4027 (approved):
lgtm!

PR #2134, opened 2022-09-28
Remove colon as delimiter in searchbar

zlayne:

Exclude colon as searchbar delimiter char
Fix string for saved searches


RobDolares (approved):
/lgtm

PR #2154, opened 2022-10-03
small changes to cluster templates and helm repositories list views

batzionb:
1. added gray background
2. removed icon from resource links
3. added link for pipeline field

### DIFF
--- a/frontend/src/routes/Infrastructure/ClusterTemplates/ClusterTemplatesPage/ClusterTemplatesTab.tsx
+++ b/frontend/src/routes/Infrastructure/ClusterTemplates/ClusterTemplatesPage/ClusterTemplatesTab.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { k8sDelete, ResourceLink, useK8sModel } from '@openshift-console/dynamic-plugin-sdk';
 import {
   Button,
+  Card,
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Infrastructure/ClusterTemplates/ClusterTemplatesPage/ClusterTemplatesTab.tsx
```diff
@@ -137,33 +158,35 @@ const ClusterTemplatesTab = () => {
...
-        errorId="templates-load-error"
-        errorMessage={t('The cluster templates could not be loaded.')}
-      >
-        <TableComposable
-          aria-label="Cluster templates table"
-          id="cluster-templates-table"
-          variant="compact"
+    <Page>
+      <PageSection>
+        <TableLoader
```
jtomasek, line 163: The content of the page section needs to be wrapped in a `<Card/>` so the table, loader and error state get correct shadow as shown here: https://www.patternfly.org/v4/components/table/react-demos/compact/ 
The TableLoader will need some spacing inside the Card, so use also <CardBody> as an optional wrapper prop in TableLoader to wrap error and skeletons.

jtomasek (changes requested):

PR #2165, opened 2022-10-07
ACM/ODF extensions document

GowthamShanmugam:
Signed-off-by: Gowtham Shanmugasundaram <gshanmug@redhat.com>

### DIFF
--- a/frontend/src/plugin-extensions/doc/acm-extensions.md
+++ b/frontend/src/plugin-extensions/doc/acm-extensions.md
@@ -1,0 +1,44 @@
+# ACM Plugin Extension Types
+
+1.  [`acm.application/action`](#acmapplicationaction)
+2.  [`acm.application/list/column`](#acmapplicationlistcolumn)
+
+---
+
+## `acm.application/action`
+
+### Summary
+
+Application action is used to add extra actions for the applications.
+
+### Properties
+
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
This PR is not valid anymore, once the extensions migration PR got merged on OCP dynamic-plugin-sdk repo (https://github.com/openshift/dynamic-plugin-sdk/pull/166), we will move the ACM custom extensions to dynamic-plugin-sdk repo. Only the ACM customs extensions usage code will remains the same.

vojtechszocs:
@GowthamShanmugam @KevinFCormier

Hi, once we merge openshift/dynamic-plugin-sdk#166 you can add ACM specific extensions under `packages/lib-extensions/src/extensions/acm`, similar to `packages/lib-extensions/src/extensions/core` structure.

Since `packages/lib-extensions/src/extensions/acm` would be an independent unit of code (which may potentially reuse other types in this package), we recommend adding an ACM specific `OWNERS` file there as well, which makes PR reviews easier.

https://github.com/kubernetes/test-infra/blob/master/prow/plugins/approve/approvers/README.md

All extension types in this package would then be published to npmjs as `@openshift/dynamic-plugin-sdk-extensions`.

Note: we plan to add scripts to generate static docs for all extension types, ideally via https://api-extractor.com/

PR #2171, opened 2022-10-13
Start ocp console when running npm run plugins

rawagner:


frontend/PLUGIN.md
```diff
@@ -35,47 +35,7 @@ npm run setup
...
-
-### [OCP Console Git Repo](https://github.com/openshift/console)
-
-Ensure that [OCP Console's dependencies](https://github.com/openshift/console#dependencies) are met in your local environment.
-
-```
-git clone git@github.com:openshift/console.git
-```
-
-### Running OCP Console in development mode
```
jtomasek, line 48: I think it is worth keeping this section to cover the case when it is necessary to run the plugins against development mode console e.g. when a change in console codebase is needed.

package.json
```diff
@@ -10,8 +10,9 @@
         "frontend:reinstall": "cd frontend && npm reinstall",
         "backend:reinstall": "cd backend && npm reinstall",
         "start": "concurrently npm:frontend npm:backend -n frontend,backend -c green,blue",
-        "plugins": "concurrently npm:backend npm:serve:plugins -c green,blue,magenta",
+        "plugins": "concurrently npm:backend npm:serve:plugins npm:ocp-console -n frontend,backend,console -c green,blue,magenta",
```
KevinFCormier, line 13: frontend and backend are in the wrong order here

start-ocp-console.sh
```diff
@@ -0,0 +1,85 @@
...
+mkdir -p ocp-console
+oc process -f ocp-console-oauth-client.yaml | oc apply -f -
+
+oc get oauthclient console-oauth-client -o jsonpath='{.secret}' > ocp-console/console-client-secret
+
+oc get secrets -n default --field-selector type=kubernetes.io/service-account-token -o json | \
+    jq '.items[0].data."ca.crt"' -r | python -m base64 -d > ocp-console/ca.crt
+
+
+CONSOLE_IMAGE=${CONSOLE_IMAGE:="quay.io/openshift/origin-console:latest"}
```
KevinFCormier, line 14: I wonder if we should pin this at the 4.10 level by default, or maybe parameterize it?

KevinFCormier (changes requested):
A few little changes requested in the comments, but this is AWESOME! We had been meaning to look into this and hadn't gotten to it yet.

@jeswanke @Randy424 @RobDolares please give this a try!

start-ocp-console.sh
```diff
@@ -0,0 +1,85 @@
...
+
+set -euo pipefail
+
+mkdir -p ocp-console
+oc process -f ocp-console-oauth-client.yaml | oc apply -f -
+
+oc get oauthclient console-oauth-client -o jsonpath='{.secret}' > ocp-console/console-client-secret
+
+oc get secrets -n default --field-selector type=kubernetes.io/service-account-token -o json | \
+    jq '.items[0].data."ca.crt"' -r | python -m base64 -d > ocp-console/ca.crt
```
KevinFCormier, line 11: Might be good to document the `python` dependency.

frontend/PLUGIN.md
rawagner: Ok, I've updated the README to contain sections for both - running the plugins against OCP downloaded from quay and running against local development build

start-ocp-console.sh
rawagner: Sounds good, by default it will download 4.10, but we can still use custom version by defining `CONSOLE_VERSION`

start-ocp-console.sh
rawagner: added `Dependencies` section to readme

rawagner:
@KevinFCormier glad you like it. The PR is now updated, please take a look again

package.json
rawagner: fixed

KevinFCormier (approved):

PR #2182, opened 2022-10-17
final

jeswanke:


frontend/src/lib/test-shots.ts
```diff
@@ -0,0 +1,432 @@
+/* Copyright Contributors to the Open Cluster Management project */
+import React from 'react'
+import debounce from 'lodash/debounce'
+import capitalize from 'lodash/capitalize'
+import camelCase from 'lodash/camelCase'
+import StackTrace from 'stacktrace-js'
+//import { inspect } from 'util'
+import path from 'path'
+
+if (process.env.NODE_ENV !== 'production') {
```
KevinFCormier, line 10: Do we need this in development? What about checking for `test` instead?

KevinFCormier (commented):
@jeswanke Could you please provide some description in the PR of what problem this solves?

frontend/src/lib/test-util.ts
```diff
@@ -321,13 +321,7 @@ export function nocksAreDone(nocks: Scope[]) {
...
 
 export async function waitForNocks(nocks: Scope[]) {
-    const timeout = options.timeout * nocks.length * 3
-    const timeoutMsg = (error: Error) => {
-        error.message = `!!!!!!!!!!! Test timed out in waitForNocks()--waited ${timeout / 1000} seconds !!!!!!!!!!!!!`
-        error.stack = ''
-        return error
-    }
-    await waitFor(() => expect(nocksAreDone(nocks)).toBeTruthy(), { timeout, onTimeout: timeoutMsg })
+    await waitFor(() => expect(nocksAreDone(nocks)).toBeTruthy())
```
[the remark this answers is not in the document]
jeswanke: Actually I added that a while ago when I didn't understand nocks---there's a built-in timeout for every waitfor of 5 seconds. Since nocks aren't dependent on a lagging backend, there would be something wrong if even a bunch of nocks didn't finish  in 5 seconds

frontend/src/lib/test-shots.ts
jeswanke: these are functions meant to help writing tests--to be run in development or test modes but not production

PR #2191, opened 2022-10-18
Fix catalog routes

KevinFCormier:
https://github.com/stolostron/backlog/issues/26405

**Note:** This PR removes the `/multicloud/infrastructure/clusters/create/infrastructure` and `/multicloud/infrastructure/clusters/pools/create/infrastructure` routes.

Instead, if the `/multicloud/infrastructure/clusters/create` or `/multicloud/infrastructure/clusters/pools/create` routes are accessed without or with an invalid `infrastructureType` search param, instead of displaying an empty form or crashing, the catalog is displayed.

The `useBackCancelNavigation()` hook is used to remember in [History.state](https://developer.mozilla.org/en-US/docs/Web/API/History/state) whether back navigation can be used for a **Back** button (and how many steps maximum) and whether back navigation can be used for a **Cancel** button (and how many steps it takes). Using back keeps the browser history cleaner and supports different entry points to a flow.

First navigations into a catalog or wizard are done using the `createBackCancelLocation()` function. Then the hook provides callbacks `nextStep` (to add another catalog page or wizard in the flow), and `back` and `cancel` which take default locations for when a user navigates directly to these pages and using the back action is not available, so that a sensible location can be used.

Cluster pool creation has 2 entry points - from the main cluster pools page, and from the cluster pools tab of the cluster sets page. So using `useBackCancelNavigation()` allows the Back and Cancel buttons to return to where the user started these actions.

When coming from the cluster sets page, the cluster set is supposed to be pre-selected. This PR also restores that behaviour.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7171,14 +7171,14 @@
             }
         },
         "@stolostron/react-data-view": {
-            "version": "1.0.38",
-            "resolved": "https://registry.npmjs.org/@stolostron/react-data-view/-/react-data-view-1.0.38.tgz",
-            "integrity": "sha512-jCk+yEVNWyPVh1ZG5OKfmgZdWJa6biD8CRz2VrOlv8B4MX0W2WHOUSTBxZtgaDw3KgMqhIO5S178JXcDYed1pA==",
+            "version": "1.2.0",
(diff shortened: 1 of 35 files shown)


fxiang1 (approved):
LGTM! Thanks Kevin for fixing those issues :)

chenz4027:
Thanks Kevin for fixing this issue 👍

PR #2201, opened 2022-10-21
Update cluster templates page navigation + tweaks

jtomasek:
* Using redirect to `/~tabs` route to handle cluster templates page tabs collided with a resource detail route, resulting in first the detail page with error to get displayed and then getting replaced with actual cluster templates page. This change fixes it by using query params to manage active tab state.
* Use custom tabs implementation instead of HorizontalNav which does not work with query params
* Tweak TableLoader and ErrorState components
* Remove unused `/~tabs` plugin route

### DIFF
--- a/frontend/src/routes/Infrastructure/ClusterTemplates/ClusterTemplateDetails/ClusterTemplateDetailsPage.tsx
+++ b/frontend/src/routes/Infrastructure/ClusterTemplates/ClusterTemplateDetails/ClusterTemplateDetailsPage.tsx
@@ -21,7 +21,7 @@ import { useClusterTemplate } from '../hooks/useClusterTemplates';
 import ClusterTemplateDetailsSections from './ClusterTemplateDetailsSections';
 
 export const getResourceListPageUrl = (resourceGVK: K8sGroupVersionKind) =>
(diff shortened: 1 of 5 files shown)


frontend/src/routes/Infrastructure/ClusterTemplates/ClusterTemplateDetails/ClusterTemplateDetailsPage.tsx
```diff
@@ -21,7 +21,7 @@ import { useClusterTemplate } from '../hooks/useClusterTemplates';
 import ClusterTemplateDetailsSections from './ClusterTemplateDetailsSections';
 
 export const getResourceListPageUrl = (resourceGVK: K8sGroupVersionKind) =>
-  `/k8s/cluster/${resourceGVK.group}~${resourceGVK.version}~${resourceGVK.kind}/~tabs`;
+  `/k8s/cluster/${resourceGVK.group}~${resourceGVK.version}~${resourceGVK.kind}`;
```
batzionb, line 24: Maybe move this code to utils and reuse it in ClusterTemplatesPage?

PR #2203, opened 2022-10-21
Speed up tests a little

KevinFCormier:

### DIFF
--- a/Makefile.prow
+++ b/Makefile.prow
@@ -24,4 +24,4 @@ unit-tests:
 	if [ ! -d "test-output" ]; then \
 		mkdir test-output; \
 	fi
-	npm run test -- --runInBand
+	npm run test -- --maxWorkers=2


Makefile.prow
```diff
@@ -24,4 +24,4 @@ unit-tests:
 	if [ ! -d "test-output" ]; then \
 		mkdir test-output; \
 	fi
-	npm run test -- --runInBand
+	npm run test -- --maxWorkers=2
```
Randy424, line 27: @ Nice! Though, makes me wonder if there was a practical reason to run tests serially in the past.

Randy424 (approved):
/lgtm

Makefile.prow
KevinFCormier: I looked into the git history, and tests had been set up to run serially, both in the build and locally, for a very long time. We do get a lot of intermittent failures in the build system if we allow parallel tests, but so far, 2 workers seems to be OK.

PR #2209, opened 2022-10-26
Change default to 4.11

KevinFCormier:
For 4.10, you cannot specify multiple plugins via the `BRIDGE_PLUGINS` environment variable. 

Support for that was added in https://github.com/openshift/console/pull/11531

Changing default to 4.11 for now.


RobDolares (approved):
/lgtm

PR #2218, opened 2022-10-31
I18n console

RobDolares:
Related issue: https://github.com/stolostron/backlog/issues/23473


frontend/src/routes/Home/Search/components/SavedSearchQueries.tsx
```diff
@@ -164,8 +166,8 @@ export default function SavedSearchQueries(props: {
                                 key={index}
                                 cardHeader={{
                                     hasIcon: true,
-                                    title: query.name,
-                                    description: query.description,
+                                    title: t(query.name),
```
KevinFCormier, line 169: Are these predefined searches, or are they user supplied? Generally we should only provide literal strings to the `t` function; otherwise, all possible calls to the `t` function should be enumerated in a comment. Same comment for line below.

frontend/src/routes/Home/Search/searchDefinitions.test.ts
```diff
@@ -12,7 +12,7 @@ import searchDefinitions, {
     GetUrlSearchParam,
 } from './searchDefinitions'
 
-test('Correctly returns formatSearchbarSuggestions without T in timestamp', () => {
+test('Correctly returns useFormatSearchbarSuggestions without T in timestamp', () => {
```
KevinFCormier, line 15: These look like leftover changes from something you tried and abandoned.

frontend/src/routes/Home/Search/searchDefinitions.tsx
```diff
@@ -1367,6 +1360,22 @@ const searchDefinitions: any = {
...
+            </div>
+        )
+    }
+    return (
+        <div>
+            <ExclamationCircleIcon color="var(--pf-global--danger-color--100)" /> {t('With violations')}
+        </div>
+    )
+}
+
```
KevinFCormier, line 1378: I don't think there is any value in factoring this out. The uppercase function name is strange - looks like a component, but not really being used as one.

frontend/src/routes/Home/Search/searchDefinitions.tsx
```diff
@@ -1497,12 +1506,7 @@ export function CreateApplicationTopologyLink(item: any) {
         const link = `${NavigationPath.applicationTopology
             .replace(':namespace', item.namespace)
             .replace(':name', item.name)}?apiVersion=${apiversion}`
-        return (
-            <a href={link}>
-                {/* TODO Not translating - caused issue: https://github.com/open-cluster-management/backlog/issues/9184 */}
-                {'View topology'}
-            </a>
-        )
+        return <a href={link}>{'View topology'}</a>
```
KevinFCormier, line 1509: Have you investigated https://github.com/open-cluster-management/backlog/issues/9184 (mentioned in the comment you removed) and whether translating this string would cause an issue?

frontend/src/routes/Home/Search/searchDefinitions.tsx
```diff
@@ -1511,7 +1515,6 @@ export function CreateExternalLink(item: any) {
     if (item.consoleURL) {
         return (
             <a target="_blank" rel="noopener noreferrer" href={`${item.consoleURL}`}>
-                {/* TODO Not translating - caused issue: https://github.com/open-cluster-management/backlog/issues/9184 */}
```
KevinFCormier, line 1514: Have you investigated https://github.com/open-cluster-management/backlog/issues/9184 and whether translating this string would cause an issue?

KevinFCormier: Also, you only removed the comment but did not add translation...

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1073,8 +1075,8 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
                 >
                     {props.emptyState ?? (
                         <AcmEmptyState
-                            title={`No ${props.plural} found`}
-                            message={`You do not have any ${props.plural} yet.`}
+                            title={t('No {{items}} found', { items: props.plural })}
+                            message={t('You do not have any {{items}} yet', { items: props.plural })}
```
KevinFCormier, line 1079: I'm not sure if this will work - we should ask the translation team if a double translation like this is OK. Some languages might not use the plural form of the word when substituting in. We might have to require that tables either have an empty state or have 2 properties to specify these strings that are used for the default empty state.

frontend/src/wizards/Argo/ArgoWizard.tsx
```diff
@@ -538,67 +551,71 @@ export function ArgoWizard(props: ArgoWizardProps) {
...
+                            label={t('Prune propagation policy')}
                             path="spec.template.spec.syncPolicy.syncOptions"
                             inputValueToPathValue={checkboxPrunePropagationPolicyToSyncOptions}
                             pathValueToInputValue={checkboxSyncOptionsToPrunePropagationPolicy}
                         >
                             <Select
-                                label="Propogation policy"
-                                options={['foreground', 'background', 'orphan']}
+                                label={t('Propogation policy')}
+                                options={[t('foreground'), t('background'), t('orphan')]}
```
KevinFCormier, line 618: Not sure that these should be translated, since they appear to be the actual values for the select options. The actual values will be substituted into YAML, yes?

Note that the options can be an array of Option, rather than just a string, which will allow you to handle translation properly.

frontend/src/wizards/Argo/ArgoWizard.tsx
```diff
@@ -651,28 +671,37 @@ export function DeploymentWindow(props: { timeZone: string[] }) {
...
-                options={['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday']}
+                options={[
+                    t('Monday'),
+                    t('Tuesday'),
+                    t('Wednesday'),
+                    t('Thursday'),
+                    t('Friday'),
+                    t('Saturday'),
+                    t('Sunday'),
+                ]}
```
KevinFCormier, line 690: Careful here as well to distinguish values vs. displayed values

frontend/src/wizards/Placement/MatchExpression.tsx
```diff
@@ -30,20 +32,20 @@ export function MatchExpression(props: { labelValuesMap?: Record<string, string[
...
                 path="operator"
                 options={[
-                    { label: 'equals any of', value: 'In' },
-                    { label: 'does not equal any of', value: 'NotIn' },
-                    { label: 'exists', value: 'Exists' },
-                    { label: 'does not exist', value: 'DoesNotExist' },
+                    { label: t('equals any of'), value: 'In' },
+                    { label: t('does not equal any of'), value: 'NotIn' },
+                    { label: t('exists'), value: 'Exists' },
+                    { label: t('does not exist'), value: 'DoesNotExist' },
```
KevinFCormier, line 48: Here is an example of how to distinguish label and value.

frontend/src/wizards/Policy/PolicyWizard.tsx
```diff
@@ -442,23 +475,25 @@ export function PolicyWizardTemplates(props: { policies: IResource[] }) {
...
                 </WizRadioGroup>
 
                 <Select
                     path="objectDefinition.spec.severity"
-                    label="Severity"
-                    placeholder="Select severity"
-                    options={['low', 'medium', 'high']}
+                    label={t('Severity')}
+                    placeholder={t('Select severity')}
+                    options={[t('low'), t('medium'), t('high')]}
```
KevinFCormier, line 496: Careful!

frontend/src/wizards/Placement/PlacementBinding.tsx
```diff
@@ -102,29 +107,30 @@ function PlacementBinding(props: { bindingSubjectKind: string; bindingSubjectApi
...
+    const { t } = useTranslation()
     return (
         <Fragment>
             <Select
                 path="kind"
-                label="Subject kind"
+                label={t('Subject kind')}
                 required
-                options={['PolicySet', 'Policy']}
+                options={[t('PolicySet'), t('Policy')]}
```
KevinFCormier, line 117: Another case where we need to distinguish value/label. But in this case, we are referring to a "kind" which should probably not be translated anyway. Translation team should know not to translate something like "PolicySet"... hard to say how they would know not to translate "Policy". In any case, never rely on a translation for proper functioning of the code though.

KevinFCormier (changes requested):
Looks pretty good overall. For the moment, I have not reviewed for completeness. Let's get this PR wrapped up, and then we can search for any missed strings.

But there are a number of issues to take a look at. I didn't necessarily find all cases, but please have another review, especially the places where you translated select options.

frontend/src/routes/Home/Search/SearchResults/SearchResults.tsx
```diff
@@ -38,15 +38,96 @@ function SearchResultTables(props: {
...
+            // 'Reclaim policy',
+            // 'Remediation action',
+            // 'Replicas',
+            // 'Requests',
+            // 'Restarts',
+            // 'Role',
+            // 'Rules',
+            // 'schedule',
+            // 'Scope',
+            // 'Scope refers to the cluster associated to the PolicyReport.',
```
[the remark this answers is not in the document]
RobDolares: This is hint for tableheader 'scope' in the result table for kind:policyreport

frontend/src/routes/Home/Search/SearchResults/SearchResults.tsx
```diff
@@ -38,15 +38,96 @@ function SearchResultTables(props: {
 
     const renderContent = useCallback(
         (kind: string, items: ISearchResult[]) => {
+            const colDefs = _.get(
+                searchDefinitions,
+                `[${kind.toLowerCase()}].columns`,
+                searchDefinitions['genericresource'].columns
+            )
+
+            // The following SearchDefinition strings for the SearchResults table need to be added to translations file:
```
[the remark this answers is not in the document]
KevinFCormier: And also via ``t(`${item.tooltip}`)`` I guess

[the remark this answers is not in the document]
RobDolares: yes, will update to say SearchResults table headers and applicable tooltips.

frontend/src/routes/Home/Search/components/SavedSearchQueries.tsx
RobDolares: I handled this elsewhere for the suggestedQueryTemplates, so removing here.

frontend/src/routes/Home/Search/searchDefinitions.tsx
RobDolares: factored out because i wasn't able to get useTranslation hook to work within cell(), but also just refactored slightly so switch statement is also factored out into what im labeling as `GetViolations` now.

KevinFCormier: You could set up `searchDefinitions.tsx` to be a hook just like `useWizardStrings`... that way you also would not need to list all the terms in a big comment. This is much less error-prone, because developers will not know to update the big comment when they add search definitions.

frontend/src/wizards/Policy/PolicyWizard.tsx
RobDolares: updated with label/value

frontend/src/wizards/Placement/PlacementBinding.tsx
RobDolares: updated to [`{label:'t(...)', value: '...'}`]
Translation team said to wrap anything questionable wrt kube resources, and they will use their discretion.

frontend/src/wizards/Argo/ArgoWizard.tsx
RobDolares: updated to `[{label:'t(...)', value: '...'}]`

frontend/src/ui-components/AcmTable/AcmTable.tsx
RobDolares: Will first make initial inquiry about the double translation to translation team.

frontend/src/wizards/Argo/ArgoWizard.tsx
RobDolares: will revert to untranslated strings for now

RobDolares: reverting

frontend/src/ui-components/AcmTable/AcmTable.tsx
RobDolares: reached out to translation team 11/7/22.  Waiting for feedback

frontend/src/routes/Home/Search/searchDefinitions.tsx
RobDolares: reverted functions until revisit changing to hook

frontend/public/locales/en/translation.json
```diff
@@ -15,9 +15,14 @@
...
     "{{count}} infrastructure environment": "{{count}} infrastructure environment",
     "{{count}} infrastructure environment_plural": "{{count}} infrastructure environments",
+    "{{count}} more": "{{count}} more",
+    "{{count}} more_plural": "{{count}} more",
+    "{{count}} selected": "{{count}} selected",
+    "{{count}} selected_plural": "{{count}} selected",
     "{{count}} user": "{{count}} user",
     "{{count}} user_plural": "{{count}} users",
     "{{hostCount}} hosts in {{infraEnvCount}} infrastructure environment": "{{hostCount}} hosts in {{infraEnvCount}} infrastructure environment",
+    "{{loading}}...": "{{loading}}...",
```
KevinFCormier, line 25: Remove please!

frontend/public/locales/en/translation.json
```diff
@@ -1081,6 +1195,8 @@
     "details": "Details",
     "Details": "Details",
     "details.breadcrumb.search": "Search",
+    "Development Console": "Development Console",
```
KevinFCormier, line 1198: Please try to remove strings that are not actually used.

frontend/src/wizards/Argo/ArgoWizard.tsx
KevinFCormier: If you reverted, the change has not been pushed. Marking as unresolved.

frontend/src/wizards/Placement/PlacementBinding.tsx
```diff
@@ -52,46 +56,47 @@ export function PlacementBindings(props: {
...
             />
             <Select
                 path="placementRef.kind"
-                label="Placement kind"
-                helperText="The placement rule used to select clusters for placement."
+                label={t('Placement kind')}
+                helperText={t('The placement rule used to select clusters for placement.')}
                 required
-                options={['Placement', PlacementRuleKind]}
+                options={[t('Placement'), PlacementRuleKind]}
```
KevinFCormier, line 75: You know what to do.

frontend/src/wizards/Argo/ArgoWizard.tsx
RobDolares: up now

frontend/src/ui-components/AcmTable/AcmTable.tsx
rh-tchuang: @KevinFCormier understood.
I'd like to double confirm that when translating the strings, if there's a segment containing possible plural form, for example "selected %s node", will there be a "selected %s nodes" for the plural candidate?

KevinFCormier: Hey @rh-tchuang,

Yes, with i18next, when we have an interpolation in the message with the name `count`, we automatically get a 2nd key with the suffix `_plural`. See https://github.com/stolostron/console/blob/356ab3059f1b222fee26ab8f7ee83cceb2740f31/frontend/public/locales/en/translation.json#L19:L20 for example.

The translation library automatically selects the right one at runtime based on the value of `count`. Note that this format for plurals is outdated (format v3), but for compatibility reasons with OpenShift console, we will be maintaining it for awhile. Eventually we will need to migrate to the updated format, which uses different suffixes like `_one` or `_other`. https://www.i18next.com/translation-function/plurals 

But it should be clear from the English string whether it is a plural or singular case.

The example in this PR was not really about plurals. 
`message={t('You do not have any {{items}} yet', { items: props.plural })}`
Here, `props.plural` contains a translated term for the type of item we are displaying. For example, for a table of applications, `props.plural` will be "applications". If there are no applications, the message "You do not have any applications yet" will be displayed. This means we are inserting a translated word into another translated message, which I imagine is not ideal. There is no singular form for this case.

PR #2219, opened 2022-10-31
Fix nodepool status not updating

fxiang1:

Jira https://issues.redhat.com/browse/ACM-1945

- Fixed the status check to check for string as the value from Kube is not a boolean

image


chenz4027 (approved):
lgtm!

PR #2168, opened 2022-10-12
Enable Korean, Japanese, simplified Chinese

KevinFCormier:
Signed-off-by: Kevin Cormier <kcormier@redhat.com>

PR #2229, opened 2022-11-01
Hypershift upgrade modal

fxiang1:
https://issues.redhat.com/browse/ACM-1745

- Create a new modal to support upgrading hypershift clusters
- The control plane and nodepools can be upgraded separately so logic was implemented to support that

image

image


frontend/public/locales/en/translation.json
```diff
@@ -83,6 +83,7 @@
     "Add credentials and custom certificate authority (CA) certificates to connect to private helm chart repository.": "Add credentials and custom certificate authority (CA) certificates to connect to private helm chart repository.",
     "Add hosts": "Add hosts",
     "Add new resource": "Add new resource",
+    "Add nodepool": "Add nodepool",
     "Add Nodepool": "Add Nodepool",
```
chenz4027, line 87: If we no longer use the upper case, maybe we should remove it?

chenz4027: One thing that I'm concerned is that "Add Nodepool" has translations in Japanese, Chinese and Korean. By changing that we would miss it from existing translations - which is fine as long as we have new translations. I wonder for the texts that are introduced (not just add nodepools, but other texts), is that a way to generate them in those three languages for the PR? 

Language translation PR: https://github.com/stolostron/console/pull/2168

chenz4027 (commented):
Also the PR still has WIP in the title ;)

frontend/public/locales/en/translation.json
KevinFCormier: @chenz4027 You do not need to worry about the non-en languages. We will have additional rounds of translation before the end of the release. This was just an initial batch of translations to make sure the setup is working.

chenz4027: Oh perfect, thanks @KevinFCormier 👍 
/lgtm

fxiang1: That's a good catch @chenz4027. I removed the extra the `Add Nodepool` with the upper case.

chenz4027 (approved):
/lgtm

Thanks Feng

PR #2246, opened 2022-11-08
Fix navigation from search due to string capitilization changes

zlayne:

Related issue: https://issues.redhat.com/browse/ACM-2003

### DIFF
--- a/frontend/src/lib/search.ts
+++ b/frontend/src/lib/search.ts
@@ -46,10 +46,10 @@ export function queryStatusCount(cluster: string): IRequestResult<ISearchResult>
                 },
                 {
                     filters: [
-                        { property: 'compliant', values: ['!Compliant'] },
+                        { property: 'compliant', values: ['NonCompliant'] },
                         { property: 'kind', values: ['policy'] },
                         { property: 'namespace', values: [cluster] },
-                        { property: 'cluster', values: 'local-cluster' },
+                        { property: 'cluster', values: ['local-cluster'] },
                     ],
                 },
             ],
--- a/frontend/src/routes/Home/Search/__snapshots__/searchDefinitions.test.ts.snap
+++ b/frontend/src/routes/Home/Search/__snapshots__/searchDefinitions.test.ts.snap
@@ -1,6 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Correctly returns CreateApplicationTopologyLink 1`] = `"-"`;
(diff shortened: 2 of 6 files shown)


frontend/src/lib/search.ts
```diff
@@ -46,10 +46,10 @@ export function queryStatusCount(cluster: string): IRequestResult<ISearchResult>
                 },
                 {
                     filters: [
-                        { property: 'compliant', values: ['!Compliant'] },
+                        { property: 'compliant', values: ['NonCompliant'] },
                         { property: 'kind', values: ['policy'] },
                         { property: 'namespace', values: [cluster] },
-                        { property: 'cluster', values: 'local-cluster' },
+                        { property: 'cluster', values: ['local-cluster'] },
```
KevinFCormier, line 52: I think this might partially address https://issues.redhat.com/browse/ACM-2004
Still getting an error though:
{"errors":[{"message":"internal system error","path":["searchResult",1,"count"]},{"message":"internal system error","path":["searchResult",1,"related"]}],"data":{"searchResult":[{"count":5,"related":[{"kind":"Application","count":1,"__typename":"SearchRelatedResult"}],"__typename":"SearchResult"},{"count":null,"related":null,"__typename":"SearchResult"}]}}

frontend/src/lib/search.ts
```diff
@@ -46,10 +46,10 @@ export function queryStatusCount(cluster: string): IRequestResult<ISearchResult>
                 },
                 {
                     filters: [
-                        { property: 'compliant', values: ['!Compliant'] },
+                        { property: 'compliant', values: ['NonCompliant'] },
```
KevinFCormier, line 49: I am concerned about whether this is equivalent. I think the previous query may have included policies that are not yet reporting status?

Is the `!` operator no longer supported? There are a number of other places where it is used. For example:
https://github.com/stolostron/console/blob/main/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/StatusSummaryCount.tsx#L129

Search for `'!` to see others.

KevinFCormier (commented):
A few questions...

frontend/src/lib/search.ts
zlayne: My aim was to address bug ACM-2004 (I didn't know a bug had been created already). I wasn't seeing the `internal system error` on my end but will try to recreate this.

zlayne:
@KevinFCormier Tests are passing now. I would like to merge this PR for some specific bugs that have come up. I will handle the capitalization for other search queries via this JIRA bug: https://issues.redhat.com/browse/ACM-2010.

PR #2251, opened 2022-11-09
Support pending status

willkutler:
GRC added a new "pending" status that will be set when a policy is waiting on a dependency. This PR makes that new status show up properly in the UI. 

https://github.com/stolostron/backlog/issues/26714


JustinKuli (commented):
Overall looks good! I did notice that on the Details page of a policy, the Violations column still says "No status: local-cluster" instead of "Pending: local-cluster". Is it possible to sneak that into this PR? Or if it would take more work we could handle it separately.

willkutler:
@JustinKuli that should be doable, I think I just missed it. I'll fix that and rebase

JustinKuli (approved):
LGTM!

JustinKuli:
I don't think I have the required approval permission. CC'ing some folks who do

/cc @RobDolares 
/cc @KevinFCormier

PR #2256, opened 2022-11-10
Fix linting for missing component keys

zlayne:

Fixes page crash for: `TypeError: Cannot read property 'setExtraStackFrame' of undefined`

### DIFF
--- a/frontend/src/components/AcmDataForm.tsx
+++ b/frontend/src/components/AcmDataForm.tsx
@@ -370,7 +370,7 @@ export function AcmDataFormDefault(props: {
     return (
         <Form isHorizontal={isHorizontal}>
             {formData.sections.map((section) => {
-                if (sectionHidden(section)) return <Fragment />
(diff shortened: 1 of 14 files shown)


frontend/src/components/AcmDataForm.tsx
```diff
@@ -779,8 +784,13 @@ export function AcmDataFormDetails(props: { formData: FormData; wizardSummary?:
                                                       }}
                                                   >
                                                       {section.inputs &&
-                                                          section.inputs.map((input) => (
-                                                              <AcmInputDescription input={input} />
+                                                          section.inputs.map((input, index) => (
+                                                              <AcmInputDescription
+                                                                  input={input}
+                                                                  key={`section--${formSection.title}-${
+                                                                      input.type ?? index
```
KevinFCormier, line 791: `formSection.title` and `input.type` are not likely to make a unique combination here. You will often have multiple `Text` inputs in a section. Why not use `input.id`?

zlayne: changed to use id

PR #2257, opened 2022-11-10
Fix the credential form issues

chenz4027:
https://issues.redhat.com/browse/ACM-1914 

Fix both crash and auto select issues

image


KevinFCormier (commented):
For `<CredentialsForm>` you have a `control` property, which is set to a `setNewSecret` state updater. Would it be possible to simply call this function only after the credential has been successfully created? That way you avoid having to create a new state to track whether creation has failed.

Perhaps this call can be done later in the function, like just before we close the modal in the successful case:
https://github.com/stolostron/console/blob/main/frontend/src/routes/Credentials/CredentialsForm.tsx#L1311

Also, is there a better name we could use for the `control` property?

chenz4027:
Thanks @KevinFCormier , will look into it

chenz4027:
Thanks @KevinFCormier , will fix it

KevinFCormier (approved):
Thanks for the updates!

PR #2264, opened 2022-11-14
Add Multiarch support for CIM

jtomasek:
Depends on https://github.com/openshift-assisted/assisted-ui-lib/pull/1747
Depends on https://github.com/openshift-assisted/assisted-ui-lib/pull/1768


rawagner (approved):

KevinFCormier:
Approving without meeting code coverage requirements, because with corrected Typescript types, more optional chaining is required because of optional `metadata` on resources, but it is not reasonable to test all of these branches.

PR #2266, opened 2022-11-15
templateeditor translations

RobDolares:
Related issues: https://github.com/stolostron/backlog/issues/23473
Signed-off-by: rhowingt@redhat.com <rhowingt@redhat.com>


frontend/src/components/TemplateEditor/controls/ControlPanelComboBox.test.js
```diff
@@ -77,13 +80,15 @@ const propsMulti = {
...
-    i18n: jest.fn(),
+    i18n: t,
 }
 
 describe('ControlPanelComboBox component', () => {
     it('basic combo', async () => {
         render(<ControlPanelComboBox {...propsPlain} />)
 
+        screen.logTestingPlaygroundURL()
+
```
KevinFCormier, line 91: Remove, please.

frontend/src/components/TemplateEditor/controls/__snapshots__/ControlPanelCheckbox.test.js.snap
```diff
@@ -14,7 +14,6 @@ exports[`ControlPanelCheckbox component renders as expected 1`] = `
       >
         <input
           aria-invalid="false"
-          aria-label="creation.app.name"
```
KevinFCormier, line 17: Rather than allowing these aria-labels to be removed from the snapshot, please set up the tests so that we either get the real translations or we get the translation keys as the values.

frontend/src/components/TemplateEditor/controls/ControlPanelFormGroup.js
```diff
@@ -34,7 +35,7 @@ class ControlPanelFormGroup extends React.Component {
                                 <Popover id={`${controlId}-label-help-popover`} bodyContent={tooltip}>
                                     <button
                                         id={`${controlId}-label-help-button`}
-                                        aria-label="More info"
+                                        aria-label={i18n && i18n('More info')}
```
KevinFCormier, line 38: i18n should not be optional

KevinFCormier (changes requested):
Left a few comments.

frontend/src/components/TemplateEditor/controls/__snapshots__/ControlPanelTextArea.test.js.snap
```diff
@@ -51,6 +51,7 @@ exports[`ControlPanelTextArea component renders as expected 1`] = `
             class="pf-c-form-control"
             data-testid="area-controlId"
             id="controlId"
+            placeholder="Enter creation.app.name"
```
KevinFCormier, line 54: I am a bit concerned that `creation.app.name` is not translated here. 
- In general, I don't think we should be using the default placeholders that substitute translated strings into other translated strings. But this may be difficult to enforce since temptifly is not written in Typescript.
- Is the `name` property for controls of type `singleselect` translated?  It appears to be from here: https://github.com/stolostron/console/blob/main/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataAWS.js#L696 but I don't see where that happens. But if it is translated, I would expect it to be translated everywhere it is used.

frontend/src/components/TemplateEditor/controls/__snapshots__/ControlPanelTextInput.test.js.snap
```diff
@@ -54,6 +54,7 @@ exports[`ControlPanelTextInput component renders as expected 1`] = `
             data-ouia-safe="true"
             data-testid="text-controlId"
             id="controlId"
+            placeholder="Enter creation.app.name"
```
KevinFCormier, line 57: Same comment here as for ControlPanelSingleSelect.

frontend/src/components/TemplateEditor/controls/__snapshots__/ControlPanelTextArea.test.js.snap
RobDolares: What if we pull out name and translate before putting into the placeholder?
```

        // if placeholder missing, create one
        let { placeholder } = control
        const translatedName = i18n(name)
        if (!placeholder) {
            placeholder = i18n('creation.ocp.cluster.enter.value', [translatedName ? translatedName.toLowerCase() : ''])
        }
```

KevinFCormier: On line 23 it is not translated either, but these must be translated in practice. So - perhaps there is preprocessing on the control data that runs certain fields through translation? (And which wouldn't be at play when you create these components directly for testing.) Can you please try to verify that? If so, then what is currently in this snapshot would be the expected behaviour I guess. 

@jeswanke may know...

KevinFCormier: Here it is:
https://github.com/stolostron/console/blob/main/frontend/src/components/TemplateEditor/utils/initialize-control-data.js#L106

frontend/src/components/TemplateEditor/controls/__snapshots__/ControlPanelTextInput.test.js.snap
KevinFCormier: Not an issue - translation is done at a higher level and does not take effect when you instantiate controls directly.

PR #2272, opened 2022-11-16
Default networking type

Randy424:
regarding: https://github.com/stolostron/backlog/issues/26524

"onImageChange" runs on release image selection, and changes the default active value for networkType based on the image version  selected. 

OCP 4.12+  defaults  to OVNKubernetes
....OpenShiftSDN for below 4.12


KevinFCormier (commented):
Looks good! Just wondering if we should have `OVNKubernetes` as the first choice and as initially selected, before the user chooses a release image, since as new OCP releases continue to be cranked out, it will be the default for most cases.

Randy424:
That makes sense!

KevinFCormier (approved):

PR #2276, opened 2022-11-17
Add backup label to submariner broker

vthapar:
Add `'cluster.open-cluster-management.io/backup': 'submariner'` label to SubmarinerBroker resource so it gets backed up during backup hub operation.

Fixes https://github.com/stolostron/backlog/issues/26876

Signed-off-by: Vishal Thapar <5137689+vthapar@users.noreply.github.com>


sridhargaddam:
The [issue](https://github.com/stolostron/backlog/issues/26876) talks about labelling the Broker as shown below.

```
cluster.open-cluster-management.io/backup: "" should be added to broker resource.
```

Whereas this PR is adding the label 
```
cluster.open-cluster-management.io/backup': 'submariner'
```
Is this intentional?

vthapar:
Yes. The label in issue is what the documentation mentions. The value in label doesn't matter from backup perspective, it only checks if label exists. I decided to use label "submariner" after noticing a similar practice in other modules. e.g. The value for similar label on managedcluster and clusterset is set to "cluster". I felt this would allow to better track/check submariner objects being backed up when have more than just one object.

PR #2289, opened 2022-11-19
Fix to allow upgrading single nodepool

fxiang1:

https://issues.redhat.com/browse/ACM-2122

- Fix case when users unchecks the control plane, the nodepools should not be disabled if they are within the n-2 version range
- Fix crash when patch has errors
- Fix issue where control plane and nodepool version are not compared correctly
- Fix incorrect image used when only upgrading nodepool ie. control plane unchecked


chenz4027 (approved):
lgtm

PR #2296, opened 2022-11-21
migrate topology to patternfly

jeswanke:
does https://github.com/stolostron/backlog/issues/19360

what changed:
1) deleted the 20 files that did our topology, moved the shapes and icon files into new topology folder
2) package.json--remove packages we used to use and include patternfly topology
3) package-lock.json--deduped so there is just one version of d3-selection
4) test-shots.ts --fixed bug and changed pasted snapshots to be commented out
5) channel changer for subscription application:
    a) moved active/allchannels into its own state 
    b) move control into pf topology toolbar
6) and of course the new Topology.tsx, etc files that implement the pf topology


frontend/tsconfig.json
```diff
@@ -15,7 +15,7 @@
         "noImplicitAny": true,
         "strict": true,
         "strictBindCallApply": true,
-        "strictFunctionTypes": true,
+        "strictFunctionTypes": false,
```
KevinFCormier, line 18: @jeswanke What is your reason for changing these settings?

KevinFCormier: (both `strictFunctionTypes` and `target`)

fxiang1 (approved):
Looks great John! Thanks!

frontend/tsconfig.json
jeswanke: ...patternfly's demo has strict turned off--they're willing to fix, but it won't be time for 2.7--alternatively we can disable whatever ts errors occur

...the patternfly layout requires us to extend the BaseLayout and override its methods--which isn't supported in es5 but is supported in es2015

all in all, using patternfly topology requires a closer integration then table or button

fxiang1 (dismissed):
lgtm

PR #2302, opened 2022-11-22
added conditional to cover undefined response

Randy424:

- We should not reference array.filter on undefined response values.
- Additional check for undefined value to avoid error

Screenshot 2022-11-22 at 12 25 43 PM


KevinFCormier:
@Randy424 Do we want to just return empty data if we encounter a problem, or should we return an error code?

Also, did you hit this in normal operation, or did you have to tweak something to encounter it?

Randy424:
@KevinFCormier This issue does not really come up in ordinary use of the api. So I think we can hang up this PR for now.

PR #2307, opened 2022-11-23
Fix and align ACM table filter cache

chenz4027:

FYI @dhaiducek 

https://issues.redhat.com/projects/ACM/issues/ACM-2068


KevinFCormier (changes requested):
What happens if two different tables use the same ID for their filters? As a developer, my intuition would tell me that the IDs only need to be unique for a given table.

I found a few issues.

1. It seems `initialFilters` is not taking precedence. There is a feature on the GRC Overview page on the Policy violations card where if you click the legend for violations, you get taken to the policies table with a filter applied. If I visit the table first and filter by something else, like namespace, then my filter based on violations will not work.
2. If the JSON in the local storage is not well-formed, the table will crash.
3. If I filter by namespaces in the policies table, then the namespace gets deleted, the filter gives me 0 results, and it still counts the namespace filter as being present, even though it is not shown.

image

chenz4027:
Thanks @KevinFCormier 

What happens if two different tables use the same ID for their filters? As a developer, my intuition would tell me that the IDs only need to be unique for a given table.
Should we make it unique then?


It seems initialFilters is not taking precedence. There is a feature on the GRC Overview page on the Policy violations card where if you click the legend for violations, you get taken to the policies table with a filter applied. If I visit the table first and filter by something else, like namespace, then my filter based on violations will not work.
Yep I actually made it on purpose. Was debating whether we should make initialFilter first or cached filter first. Discussed with Dale and he looked into the code and he wasn't sure whether GRC update the URL for the table filters (we weren't able to find any code that would do that). Given that I move cache first over preset. However, I can move the sequence so that if there's any initial Filter, we would always render that instead of cache. Or maybe we render both?

If the JSON in the local storage is not well-formed, the table will crash.
Yep that's correct, I wonder what's the best approach for this issue? Ideally, user interactions should be okay as the cache are stored properly. However, if the user modify the cache in developer mode, I don't think we have a way to prevent it?


If I filter by namespaces in the policies table, then the namespace gets deleted, the filter gives me 0 results, and it still counts the namespace filter as being present, even though it is not shown.
Ok this is for sure an issue, will look into it.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -353,7 +355,35 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
     }
     const initialSort = props.initialSort || defaultSort
     const initialSearch = props.initialSearch || ''
-    const initialFilters = props.initialFilters || {}
+
+    let existingFilter: any = {}
+
+    const cachedFilter = id && localStorage[id]
+    if (cachedFilter) {
+        existingFilter = JSON.parse(cachedFilter)
```
chenz4027, line 363: retest

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -779,13 +809,10 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
     const clearSearchAndFilters = useCallback(() => {
         clearSearch()
         clearFilters()
-        for (const filter of filters) {
-            const cachedFilters = filter.cachedFilters
-            if (cachedFilters) {
-                cachedFilters.length = 0
-            }
+        if (id) {
+            localStorage.removeItem(id)
```
chenz4027, line 813: prefix ids

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1192,15 +1222,29 @@ function TableColumnFilters<T>(props: {
                 /* istanbul ignore next */
                 if (filter.options.find((option) => option.value === selection)) {
                     filterId = filter.id
-                    const cachedFilters = filter.cachedFilters
-                    if (cachedFilters) {
-                        if (!cachedFilters.includes(selection)) {
-                            cachedFilters.push(selection)
+                    const filterCache = JSON.parse(localStorage.getItem(id) as string)
```
chenz4027, line 1225: make it a common component

dhaiducek:
LGTM! Thanks for the updates, @chenz4027! And for reminding me why we have the initial table filtering in GRC! 😬 

(I had forgotten why we had it and was wondering just now whether it should be removed, but I think it's helpful to be able to construct a shareable URL that filters the GRC table for users.)

PR #2308, opened 2022-11-23
Automation workflows

Randy424:
- Added request to workflow API to pull down workflow templates
- Added radio buttons for choosing between template types
- Changes some function names to "*Template" to avoid using the, now more meaningful "Job", keyword...

WIP: 
- Cluster curator currently provides the status url link we surface to the console when running automation. The URL for workflow jobs is bad.

Example: 
 - lastTransitionTime: "2022-11-23T20:34:22Z"
    message: https://ansible-tower-web-svc-tower.apps.collective.aws.red-chesterfield.com/#/jobs/workflow/74928
    reason: ansiblejob_url
    status: "True"
    type: prehookjob-78949 

**URL should not contain "/Jobs"**

I've updated the Jira epic with this issue, hopefully CLC squad can get it squared away.

- Testing

- Still need to apply changes to our function names and doc strings more broadly to avoid misunderstanding with the "Job" vs "Workflow" keywords

See screenshots:
Screenshot 2022-11-23 at 12 58 04 PM

Screenshot 2022-11-23 at 12 58 12 PM

Screenshot 2022-11-23 at 12 58 24 PM


frontend/src/resources/cluster-curator.ts
```diff
@@ -56,6 +56,7 @@ export interface ClusterCurator {
 export interface ClusterCuratorAnsibleJob {
     name: string
     extra_vars?: Record<string, string>
+    type?: string
```
KevinFCormier, line 59: Let's be specific on the type here. In my PR for the operator detection, this is how I set up the types: https://github.com/stolostron/console/pull/2313/files#diff-628bf0cfb854ec47885ef6e2f08a9a20d41dab04cc0c9a37ac1333aafdf93be4

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -595,18 +631,38 @@ function EditAnsibleJobModal(props: {
             onClose={() => props.setAnsibleJob()}
         >
             <AcmForm>
+                <span style={{ display: 'inline-flex' }}>
+                    <Radio
+                        name={'job-template'}
+                        id={'job-template'}
+                        label={t('Job template')}
+                        isChecked={filterForJobTemplates}
+                        className={classes.radio}
```
KevinFCormier, line 640: Please use a `<FormGroup>` with `isInline` property instead of introducing custom CSS.

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -580,13 +592,37 @@ function EditAnsibleJobModal(props: {
...
+        },
+    })
+    const classes = useStyles()
     const [ansibleJob, setAnsibleJob] = useState<ClusterCuratorAnsibleJob | undefined>()
+    const [filterForJobTemplates, setFilterForJobTemplates] = useState(true)
     useEffect(() => setAnsibleJob(props.ansibleJob), [props.ansibleJob])
+
+    const newTemplateSelection = (jobName: string | undefined) => {
+        if (ansibleJob) {
+            ansibleJob.type = filterForJobTemplates ? 'Job' : 'Workflow'
```
KevinFCormier, line 613: Don't mutate original state. Mutate the copy.

KevinFCormier: Also, does this belong in `newTemplateTypeSelection`? Why are we updating the name in both functions?

KevinFCormier (changes requested):
A few inline comments.

I was also wondering, once we have confirmed terminology, should we change these labels when you change the radio selection, so it is apparent something has changed?
image

Also, the placeholder text says "Enter" but that is not strictly correct. You can search for a template name, but you can't manually enter the name.

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
Randy424: The name update in `newTemplateTypeSelection` wipes the name field away as we change template types. I guess I don't really need to pass in a jobName for this function, so I will remove that argument.

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
Randy424: It seems like `isInline` is not a prop on `<FormGroup>` or on `<Radio>` So I've applied a `<Split>` instead. Please let me know if I'm mistaken!

KevinFCormier: I meant `<FormGroup>` from `@patternfly/react-core`, not `@material-ui/core`.

frontend/public/locales/en/translation.json
```diff
@@ -2718,19 +2718,19 @@
     "template.create": "Create Ansible template",
     "template.create.config.wizard.title": "Configure Ansible job template name and Ansible credential",
     "template.create.destroy": "Destroy",
-    "template.create.destroy.wizard.title": "Configure Ansible job templates to run for your cluster at destroy",
+    "template.create.destroy.wizard.title": "Configure automation templates to run for your cluster at destroy",
```
KevinFCormier, line 2721: "Automation template" = ClusterCurator, so it is a template that contains lists of Ansible templates (both Ansible job templates and Ansible workflow job templates) from which Ansible jobs will be created before or after various cluster lifecycle events. So I don't think this is the correct term here.

KevinFCormier: I thought you wanted to address terminology separately, so if that is the case, please revert all changes to translation.json other than the new strings. I know this unfortunately affects tests as well.

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -612,8 +667,16 @@ function EditAnsibleJobModal(props: {
...
-                    {props.ansibleTowerTemplateList
-                        ? props.ansibleTowerTemplateList?.map((name) => (
+                    {filterForJobTemplates
+                        ? props.ansibleTowerTemplateList
+                            ? props.ansibleTowerTemplateList?.map((name) => (
+                                  <SelectOption key={name} value={name}>
+                                      {name}
+                                  </SelectOption>
+                              ))
+                            : undefined
```
KevinFCormier, line 677: Sonar is complaining about the nested ternaries here, and I agree.

How about doing something like this early in the render function:
```
const { ansibleTowerTemplateList = [], ansibleTowerWorkflowTemplateList = [] } = props
```

frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
```diff
@@ -588,21 +619,45 @@ function EditAnsibleJobModal(props: {
...
-                            copy.name = name as string
-                            setAnsibleJob(copy)
-                        }
+                        newTemplateSelection(name)
                     }}
                     variant={SelectVariant.typeahead}
-                    placeholder={t('template.modal.name.placeholder')}
+                    placeholder={t(
+                        filterForJobTemplates
+                            ? 'template.modal.name.placeholder'
```
KevinFCormier, line 658: Move the function call inside the ternary. `t` should always be called with a string literal.

frontend/public/locales/en/translation.json
```diff
@@ -2740,7 +2740,7 @@
     "template.information.description": "To create a sequence of events, add multiple jobs. Drag and drop the job templates to reorder the sequence.",
     "template.information.title": "Basic information",
     "template.job.duplicate.error": "The job name already exists in this hook. The job list cannot contain duplicates.",
```
KevinFCormier, line 2742: Looks like this validation string is used for both jobs and workflows?

KevinFCormier (changes requested):
Please see new comments.

KevinFCormier (approved):

PR #2309, opened 2022-11-23
ACM-2605 Skip machine pools if worker node count != 3

KevinFCormier:
https://issues.redhat.com/browse/ACM-2065


KevinFCormier:
/cc @Randy424 

This is the 2.6 version ready for when the stream opens. Just adding skipMachinePools when size != 3.

PR #2319, opened 2022-11-30
initial commit, color scale changes

Randy424:
regarding: https://github.com/stolostron/backlog/issues/26879

Screenshot 2022-11-30 at 12 09 30 AM
Screenshot 2022-11-30 at 12 12 16 AM
Screenshot 2022-11-30 at 12 12 00 AM

### DIFF
--- a/frontend/src/routes/Governance/overview/PolicyViolationSummary.tsx
+++ b/frontend/src/routes/Governance/overview/PolicyViolationSummary.tsx
@@ -96,11 +96,7 @@ export function ViolationsCard(props: {
                         link: props.pending > 0 ? `${NavigationPath.policies}?violations=pending` : undefined,
                     },
                 ]}
-                colorScale={[
-                    'var(--pf-global--danger-color--100)',
-                    'var(--pf-global--success-color--100)',
-                    'var(--pf-global--warning-color--100)',
-                ]}
+                colorScale={['var(--pf-global--danger-color--100)', '#BBB', 'var(--pf-global--warning-color--100)']}
             />
         </Card>
     )
--- a/frontend/src/routes/Governance/overview/PolicySetViolationSummary.tsx
+++ b/frontend/src/routes/Governance/overview/PolicySetViolationSummary.tsx
@@ -74,11 +74,7 @@ export function ViolationsCard(props: {
                         link: props.compliant > 0 ? `${NavigationPath.policySets}?violation=no-violation` : undefined,
                     },
                 ]}
-                colorScale={[
-                    'var(--pf-global--danger-color--100)',
-                    'var(--pf-global--success-color--100)',
-                    'var(--pf-global--warning-color--100)',
-                ]}
+                colorScale={['var(--pf-global--danger-color--100)', '#BBB', 'var(--pf-global--warning-color--100)']}
             />
         </Card>
     )
--- a/frontend/src/routes/Governance/policy-sets/components/PolicySetDetailSidebar.tsx
+++ b/frontend/src/routes/Governance/policy-sets/components/PolicySetDetailSidebar.tsx
@@ -75,7 +75,7 @@ function renderDonutChart(clusterComplianceSummary: { compliant: string[]; nonCo
                     <ChartLegend
                         data={legendData}
                         labelComponent={<ChartLabel />}
-                        colorScale={['var(--pf-global--success-color--100)', 'var(--pf-global--danger-color--100)']}
+                        colorScale={['#BBB', 'var(--pf-global--danger-color--100)']}
                     />
                 }
                 labels={({ datum }) => `${datum.x}: ${datum.y}`}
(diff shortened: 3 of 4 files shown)


frontend/src/routes/Governance/overview/PolicyViolationSummary.tsx
```diff
@@ -96,11 +96,7 @@ export function ViolationsCard(props: {
                         link: props.pending > 0 ? `${NavigationPath.policies}?violations=pending` : undefined,
                     },
                 ]}
-                colorScale={[
-                    'var(--pf-global--danger-color--100)',
-                    'var(--pf-global--success-color--100)',
-                    'var(--pf-global--warning-color--100)',
-                ]}
+                colorScale={['var(--pf-global--danger-color--100)', '#BBB', 'var(--pf-global--warning-color--100)']}
```
Randy424, line 99: A note on #BBB: This CSS color class corresponds to "--pf-chart-color-black-200", however, the "pf" version of this class does not seem to be one of the built in classes for the donut chart component, as using the key renders a solid black color. So I think I am forced to use the CSS class directly.

Randy424:
@heyethankim @KevinFCormier For your consideration...
It looks like some of these decisions are still being [discussed](https://docs.google.com/presentation/d/1ZiDAXhEUE-PaX408mEkC70O4sCb78oxQ3kX5SrpibUo/edit?disco=AAAAjjNsatU)?

heyethankim:
> @heyethankim @KevinFCormier FYI - For your consideration...
> It looks like some of these decisions are still being [discussed](https://docs.google.com/presentation/d/1ZiDAXhEUE-PaX408mEkC70O4sCb78oxQ3kX5SrpibUo/edit?disco=AAAAjjNsatU)?

I am going to meet with designers from PatternFly today and get a consensus at the Design Playback tomorrow.

heyethankim:
> get a consensus at the Design Playback tomorrow.

@Randy424 

Shared the doc in the Design Playback today. It was well received. Here are the recommendations in the doc. Sho mentioned this should be treated it as a defect. 

image
image

KevinFCormier (commented):
Hey Randy,

Let's try to use the chart colors as specified rather than finding a variable that currently results in the same colour, because they could change in the future. I tried changing `var(--pf-global--danger-color--100)` to `var(--pf-chart-color-red-100)` and found that it works when you run as a dynamic plugin. This is because OCP is including the correct stylesheet.

For our local development purposes, you can modify `frontend/src/index.tsx` to have the following imports (I moved stylesheets to the top to keep them together):
```
/* Copyright Contributors to the Open Cluster Management project */
/* istanbul ignore file */

import '@patternfly/react-core/dist/styles/base.css'
import '@patternfly/patternfly/patternfly-charts.css'
import '@patternfly/patternfly/patternfly-charts-theme-dark.css'
import { Fragment, StrictMode, Suspense } from 'react'
import ReactDOM from 'react-dom'
import 'regenerator-runtime/runtime'
import App from './App'
```

In terms of organization, it looks like we have the same `colorScale` values repeated a few times. If you look at the [PF example for themeColors](https://www.patternfly.org/v4/charts/themes#multi-color-ordered), I think we want to do something like that. Centralize them somewhere (maybe in AcmDonutChart? or elsewhere if they are useful in other contexts) and we should see the color variables referenced only once.

I see in some cases, the items appear in a different order. Might be worth looking if that should be standardized, for example, success might always be first or last.

frontend/src/ui-components/AcmCharts/AcmDonutChart/AcmDonutChart.tsx
```diff
@@ -137,3 +137,23 @@ export function AcmDonutChart(props: {
...
+        'var(--pf-chart-color-gold-300)',
+        'var(--pf-chart-color-black-100)',
+    ],
+    dangerImportantModerateInfo: [
+        'var(--pf-chart-color-red-100)',
+        'var(--pf-global--warning-color--100)',
+        'var(--pf-chart-color-gold-300)',
+        'var(--pf-global--info-color--100)',
+    ],
+}
```
KevinFCormier, line 159: Much better! There is still repetition in here I'd like to see factored out. ie: introduce a constant for `failure`, `success`, and so on.

Also, please double-check the colours against the latest from @heyethankim 
- should warning be --pf-chart-color-gold-300 for "Pending" or any intermediate status?
- should the insights colors be called Critical, Important, Moderate, Low? The prescribed colors should all be from pf-chart I think?

heyethankim: > * should warning be --pf-chart-color-gold-300 for "Pending" or any intermediate status? **YES**

heyethankim: > * should the insights colors be called Critical, Important, Moderate, Low? The prescribed colors should all be from pf-chart I think?

This is what I shared. If you could swap the "global colors" with the "chart colors," that will be great.  
image

PR #2331, opened 2022-12-02
ACM-2105 - Added datalist cell for Ansible job type

Randy424:

### DIFF
--- a/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
+++ b/frontend/src/routes/Infrastructure/Automations/AnsibleAutomationsForm.tsx
@@ -261,6 +261,9 @@ export function AnsibleAutomationsForm(props: {
                         ))}
                     </ChipGroup>
                 )}
+                <FlexItem style={{ fontWeight: 'bold' }}>
+                    {ansibleJob.type === 'Job' ? t('Job template') : t('Workflow job template')}
+                </FlexItem>
             </Flex>,
         ]
     }


KevinFCormier:
What about using a Label? And I think it should probably come before the variables, maybe even before the name. Or maybe something like these in OCP:
image

JT or WJT, and on hover it would tell you which type?

Thoughts @imjoyjean ?

Randy424:
JT, WJT?

KevinFCormier:
> JT, WJT?

Job template, Workflow job template

OCP uses those little badges for resource name abbreviations

Randy424:
Note that a bug has been found related to the datalist edit feature. No need to make a new issue:
https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=16125&view=detail&selectedIssue=ACM-2286

Randy424:
@KevinFCormier correct me if I'm wrong here, but I don't think those labels are defined well in patternfly. They are custom.

Icon class reference: https://github.com/openshift/console/blob/64228873e86aa335169c8821294aba8eeec1ec1a/frontend/public/components/utils/resource-icon.tsx#L30

CSS : https://github.com/openshift/console/blob/64228873e86aa335169c8821294aba8eeec1ec1a/frontend/public/components/utils/resource-icon.tsx#L30

The dropdown you referenced: https://github.com/openshift/console/blob/15589a466ddc839c7876c05bda01308c4a4f497a/frontend/public/components/resource-dropdown.tsx

I think I'l try the[ PF native label](https://www.patternfly.org/v4/components/label) if that's ok

Randy424:
Screenshot 2022-12-02 at 3 19 44 PM

KevinFCormier:
@Randy424 Yeah, I guess it's not something directly from PF. Here's the component itself:
https://github.com/openshift/console/blob/master/frontend/public/components/utils/resource-icon.tsx

So yeah, Label sounds good. Seeing it, I wonder if the words "Job" and "Workflow" would look better?

Randy424:
Screenshot 2022-12-02 at 4 24 12 PM


Screenshot 2022-12-02 at 4 22 47 PM

Randy424:
/unhold
Approval received from @imjoyjean

Randy424:
FYI @dtthuynh

PR #2332, opened 2022-12-02
Fixes ACM-2269 where expandable titles in hypershift cluster details …

o-farag:
…in dark mode had bad contrast

Hypershift cluster details had bad contrast in dark mode:

Before:
image

After: 
image

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/OnboardingModal.css
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/OnboardingModal.css
@@ -5,3 +5,7 @@
 .pf-c-modal-box__footer {
     padding: 0px;
 }
+
+:root:where(.pf-theme-dark) .pf-c-expandable-section__toggle-text {
+    color: #FFFFFF;
+}


KevinFCormier:
@o-farag @fxiang1 

This problem was introduced by the OnboardingModal, which itself does not work properly in dark mode:
image

We must avoid custom CSS as much as possible. These hard-coded text colours are not the right way to go. In addition, the styles in `OnboardingModal.css` are not automatically scoped to the component. They are affecting toggles and modals globally, like this one @o-farag has pointed out, and for example this modal:

image

For imported CSS files, you must use selectors to scope the style appropriately, for example by using a custom class name on your element and including that in the selector. Or you can use `makeStyles` to get a style for your component only.

If you remove the colors from the divs in the modal, it adjusts the font color automatically for dark mode:
image

Will need some investigation to see if the colours you were trying to use correspond to some PF colours that can be automatically switched with the theme.

dtthuynh:
This affects https://issues.redhat.com/browse/ACM-2229 as well.

o-farag:
Oh I see, I wasn't aware of this. Thanks for the explanation, I'll have to look into it more. Closing the PR for now

fxiang1:
Thanks @KevinFCormier ! will keep that in mind.

PR #2346, opened 2022-12-06
Fix add nodepool subnet id and onboarding model text

fxiang1:
https://issues.redhat.com/browse/ACM-2118
https://issues.redhat.com/browse/ACM-2230

- Disable the add nodepool subnet id if auto populated to prevent users from editing(Nodepools should be in the same subnet) Otherwise, if there's no existing nodepool the subnet id input will be enabled.
- Updated the onboarding modal text, removed 'All Clusters'

image


chenz4027 (approved):
/lgtm

PR #2347, opened 2022-12-07
Adding translations for ja, zh, ko

rh-tchuang:
Translations for Japanese, Simplified Chinese and Korean is completed.


KevinFCormier:
@rh-tchuang Can you please update the filenames as follows (you should find there are already files in these locations from your last drop):

`frontend/public/locales/en/ja/translationja.json` -> `frontend/public/locales/ja/translation.json`
`frontend/public/locales/en/ko/translationko.json` -> `frontend/public/locales/ko/translation.json`
`frontend/public/locales/en/zh_ch/translationzh_cn.json` -> `frontend/public/locales/zh/translation.json`

Note we are using `zh` instead of `zh_cn`. This matches how it is set up in OpenShift console, which I believe offers these same 3 languages:
image

See also https://github.com/openshift/console/tree/master/frontend/public/locales

For the DCO message, we require the --signoff line on the commit. If this is not practical, it may make more sense for you to just notify us when translations are updated, and we can pull from your fork.

CC @RobDolares

PR #2358, opened 2022-12-08
Adding translations for ja, zh, ko - fixed

rh-tchuang:
- Adding translations for Japanese, Chinese and Korean
- Changed macrolanguage zh
- Changed exported filename format


KevinFCormier:
Superceded by https://github.com/stolostron/console/pull/2361

PR #2362, opened 2022-12-08
Fix Hypershift view log link and create cluster catalog text

fxiang1:
- https://issues.redhat.com/browse/ACM-1910
- https://issues.redhat.com/browse/ACM-2129

- Changed the view log link for Hypershift clusters to go to the control plane pods page
- Fixed create cluster catalog BM control plane text based on usability study


chenz4027 (approved):
/lgtm
Seems like the token issue won't be resolved soon. Will approve for now

PR #2366, opened 2022-12-08
does not create a new placement config

jeswanke:
https://issues.redhat.com/browse/ACM-2103

automatically generated placement rule name was -1, etc which already exists with an editted sub app


chenz4027 (dismissed):
lgtm, thanks John!

PR #2370, opened 2022-12-09
ACM-2195 Use annotation instead of installAttemptsLimit

KevinFCormier:
https://issues.redhat.com/browse/ACM-2195


Randy424 (approved):
/lgtm

PR #2372, opened 2022-12-09
update topology legend

jeswanke:
CleanShot 2022-12-09 at 15 25 31@2x


KevinFCormier (changes requested):
Strings in the diagrams should be externalized.

PR #2374, opened 2022-12-09
Added provider type to AWS and Host Inventory create cluster pages

o-farag:

Add control plane type to the create cluster catalog as per https://issues.redhat.com/browse/ACM-2357

image

image


o-farag:
@fxiang1 I noticed that only AWS and Host Inventory have the created catalog cards, while the rest look like this
image

Should I add it to these pages as well?

I also hard coded the types into the title for AWS and Host Inventory because they have their own pages, not sure if that was a good way for me to do it. Also, the types seem to be a bit too large since they're the same size as the title, I was thinking maybe a smaller text or something less aggressive would be better?

fxiang1:
> @fxiang1 I noticed that only AWS and Host Inventory have the created catalog cards, while the rest look like this: image
> 
> Should I add it to these pages as well?
> 
> I also hard coded the types into the title for AWS and Host Inventory because they have their own pages, not sure if that was a good way for me to do it. Also, the types seem to be a bit too large since they're the same size as the title, I was thinking maybe a smaller text or something less aggressive would be better?

@o-farag Right, only AWS and Host inventory has HostedCluster support currently so you only need to change those.
As for the type size, I think it should be fine since we want to make sure the user can clearly see the type and not miss it.

frontend/public/locales/en/translation.json
```diff
@@ -740,6 +740,8 @@
     "Control plane pods": "Control plane pods",
     "Control plane status": "Control plane status",
     "Control plane type": "Control plane type",
+    "Control plane type - AWS": "Control plane type - AWS",
```
fxiang1, line 743: @o-farag For the translations, maybe you can use a variable instead for the type so we can reuse the same translation. Something like this https://github.com/fxiang1/console/blob/main/frontend/src/components/TemplateSummaryModal.tsx#L185

fxiang1 (changes requested):

frontend/public/locales/en/translation.json
o-farag: Alright, changed it to use a variable instead.

fxiang1 (approved):
LGTM! Thanks Omar!

PR #2378, opened 2022-12-12
fix and prevent infinite loop

jeswanke:
this was causing infinite loop
CleanShot 2022-12-12 at 10 58 08@2x


KevinFCormier (commented):
These limits seem arbitrary. What happens if you have a topology that exceeds these limits? Is there a better way to prevent an infinite loop?

PR #2381, opened 2022-12-12
Capture cluster control data baseline snapshots

KevinFCormier:
Convert cluster creation tests to capture snapshots so that control data can be refactored without introducing regressions.


Randy424 (approved):
/lgtm

Cool idea.

dtthuynh:
very cool @KevinFCormier !

PR #2383, opened 2022-12-13
MGMT-12590 Use Jira for customer bug reports

mareklibra:
Depends on:
- [x] https://github.com/openshift-assisted/assisted-ui-lib/pull/1799


mareklibra:
The commit is moved to https://github.com/stolostron/console/pull/2407 , closing here.

PR #2388, opened 2022-12-13
Show updates only beside updatable resources

o-farag:

Work on the issue where `Upgrade available` would be displayed alongside non-upgradable resources which may cause confusion. See https://issues.redhat.com/browse/ACM-2331

Scenario 1 - not all nodepools are upgradable:

- Before: All Nodepools and CP show as upgradable even though only 1 nodepool is upgradable 
image

- After: Only upgradable nodepool shows as upgradable image

Only show updatable nodepool - main page still shows upgradable since either nodepools or hcp on the cluster page are upgradable:

Scenario 2 - Nodepool is not at latest version but cannot be upgraded past the version of the CP:

- Before: Nodepool would show upgradable, but would require CP upgrade first image

- After: Nodepool won't show upgradable unless its version is less than that of CP 
image

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
@@ -571,6 +571,7 @@ export function useClusterDistributionColumn(
                 hostedCluster={hostedClusters.find(
                     (hc) => cluster.namespace === hc.metadata?.namespace && cluster.name === hc.metadata?.name
                 )}
+                resource={'managedclusterpage'}
             />
         ),
     }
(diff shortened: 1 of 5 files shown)


chenz4027 (approved):
/lgtm
Thanks for the fix 👍

o-farag:
@chenz4027 I looked at the failed test case, it seems to be unrelated to these changes. I also tested it on the current main branch and the same failure occurs:
image

chenz4027:
Yeah @o-farag it was a time sensitive test that failed. @KevinFCormier has a fix for it: https://github.com/stolostron/console/pull/2429#issuecomment-1370047845
You will need to merge the latest main after the fix is in.

KevinFCormier:
@o-farag You probably don't need to merge main into this PR, because your code will be merged into main in order to run the tests. I would try resetting your branch back to before the merge commit and force pushing it.

chenz4027 (approved):
/lgtm

PR #2396, opened 2022-12-15
fix onChangeDisconnect function

RobDolares:
Signed-off-by: rhowingt@redhat.com <rhowingt@redhat.com>


RobDolares:
@KevinFCormier - Here is the other case.

PR #2401, opened 2022-12-16
Improve or remove messages regarding consumption of cloud resources

KevinFCormier:


KevinFCormier:
@berenss Might need a bit more finessing. One of the messages says that hibernating “blocks all cluster operations” but that is a strange way of putting it to me. Maybe we should still mention that virtual machines are stopped? What do you think.

berenss (dismissed):
Hi Kevin I like this new content. It is succinct, and I think the messaging is safe within the context of what a user should expect.
perhaps using 'management operations' more accurately describes what you mean
> "Moving to the Hibernating state blocks any management operations for the clusters

berenss (approved):
🚀

PR #2408, opened 2022-12-19
Fix ocplaunch link

fxiang1:

https://issues.redhat.com/browse/ACM-2299

- Switch hard coded url to getResource() so it can take advantage of the mce plugin proxy

### DIFF
--- a/frontend/src/lib/ocp-utils.ts
+++ b/frontend/src/lib/ocp-utils.ts
@@ -1,3 +1,5 @@
+import { ConfigMapApiVersion, ConfigMapKind, getResource } from '../resources'
+
 /* Copyright Contributors to the Open Cluster Management project */
 function api<T>(url: string, headers?: Record<string, unknown>): Promise<T> {
     return fetch(url, headers).then((response) => {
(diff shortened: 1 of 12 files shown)


frontend/src/lib/ocp-utils.ts
```diff
@@ -9,20 +11,27 @@ function api<T>(url: string, headers?: Record<string, unknown>): Promise<T> {
...
-    )
-        .then(({ data }) => {
+    const resourceResult = getResource({
+        apiVersion: ConfigMapApiVersion,
+        kind: ConfigMapKind,
+        metadata: {
+            name: 'console-public',
+            namespace: 'openshift-config-managed',
+        },
+    }).promise
```
KevinFCormier, line 21: @fxiang1 Since we only run inside the OpenShift console, you don't need to look up this URL. You can just use a link starting with /

I have removed some of these cases, but some still remain. The only thing is that it won't work for local development, but I think we should probably come up with something that is not used in production for that case.

fxiang1: Thanks @KevinFCormier ! I changed it to just use /. For local development I agree we should come up with something.

frontend/src/lib/ocp-utils.ts
```diff
@@ -8,22 +8,12 @@ function api<T>(url: string, headers?: Record<string, unknown>): Promise<T> {
...
-        .catch((error) => {
-            onError?.()
-            // eslint-disable-next-line no-console
-            console.error(error)
-        })
+export function launchToOCP(urlSuffix: string, newTab: boolean) {
+    if (newTab) {
+        window.open(`/${urlSuffix}`)
+    } else {
+        location.href = `/${urlSuffix}`
```
KevinFCormier, line 15: I believe this will result in an unnecessary page load. It would be smoother to use history.push from react-router-dom. But it looks like we are not currently using this anywhere in the code, so I would probably opt to remove it for now. I also noticed that `launchToOCP` is passed as a property of several components for seemingly no reason. Could we remove these?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/HypershiftClusterDetails.tsx
```diff
@@ -32,9 +32,7 @@ const HypershiftClusterDetails: React.FC = () => {
                             hostedCluster={hostedCluster}
                             nodePools={clusterNodePools}
                             clusterImages={clusterImageSets as CIM.ClusterImageSetK8sResource[]}
-                            launchToOCP={(url, newTab) =>
-                                launchToOCP(url, newTab, () => window.open(`${window.location.origin}/${url}`))
-                            }
+                            launchToOCP={(url, newTab) => launchToOCP(url, newTab)}
```
KevinFCormier, line 35: This could be expressed as `launchToOCP={launchToOCP}`, but as I mentioned - I would probably just import the function directly in `HostedClusterProgress` unless we're actually using an alternative implementation in some case?

PR #2409, opened 2022-12-19
Add the create cluster set link in the policy set wizard

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-2438


dhaiducek (approved):
LGTM!

Maybe we should consider dropping OWNERS files into the `wizards/Policy*` directories?

PR #2411, opened 2022-12-19
Only display the onboarding modal if there are no existing clusters(except local-cluster).

o-farag:

Closes https://issues.redhat.com/browse/ACM-2480

Before: Onboarding Modal would show even with multiple clusters
image

Now: Only shows when one cluster (local-cluster) exists

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
@@ -73,7 +73,9 @@ export default function ManagedClusters() {
 
     const onBoardingModalID = `${window.location.href}/clusteronboardingmodal`
     const [openOnboardingModal, setOpenOnboardingModal] = useState<boolean>(
-        localStorage.getItem(onBoardingModalID) ? localStorage.getItem(onBoardingModalID) === 'show' : true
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -73,7 +73,9 @@ export default function ManagedClusters() {
 
     const onBoardingModalID = `${window.location.href}/clusteronboardingmodal`
     const [openOnboardingModal, setOpenOnboardingModal] = useState<boolean>(
-        localStorage.getItem(onBoardingModalID) ? localStorage.getItem(onBoardingModalID) === 'show' : true
+        localStorage.getItem(onBoardingModalID)
+            ? localStorage.getItem(onBoardingModalID) === 'show'
+            : clusters.length === 1
```
fxiang1, line 78: @o-farag Can you also check if the one cluster on the list is `local-cluster`? If RBAC is enabled, the user might not have access to see `local-cluster` so if there's a cluster, it might be one that they already created.

fxiang1 (changes requested):
Request changes

fxiang1 (approved):
LGTM! Thanks Omar!

PR #2416, opened 2022-12-20
[WIP] Fix Hypershift cluster details line wrap issue

fxiang1:
https://issues.redhat.com/browse/ACM-2411

- Use ref to get the width from a parent component and set that to the NodePoolTable. However because ref won't trigger a re-render when changed the NodePoolTable width won't change when the user is resizing the browser. Will see if I can find a better solution.


fxiang1:
Issue fixed already. Closing.

PR #2418, opened 2022-12-20
Restore username route

KevinFCormier:
username route still used by search

https://github.com/stolostron/backlog/issues/27047


RobDolares (approved):
/lgtm

PR #2422, opened 2022-12-21
Refactor definedNamespace

chenz4027:
https://github.com/stolostron/backlog/issues/26955

### DIFF
--- a/frontend/src/routes/Applications/Overview.tsx
+++ b/frontend/src/routes/Applications/Overview.tsx
@@ -53,6 +53,7 @@ import {
 import { isLocalSubscription } from './helpers/subscriptions'
 import { getArgoDestinationCluster } from './ApplicationDetails/ApplicationTopology/model/topologyArgo'
 import { PluginContext } from '../../lib/PluginContext'
+import { get } from 'lodash'
 
 const gitBranchAnnotationStr = 'apps.open-cluster-management.io/git-branch'
 const gitPathAnnotationStr = 'apps.open-cluster-management.io/git-path'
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Applications/Overview.tsx
```diff
@@ -349,10 +349,9 @@ export default function ApplicationsOverview() {
                 .filter((argoApp) => {
                     const resources = argoApp.status ? argoApp.status.resources : undefined
                     let definedNamespace = ''
-                    resources &&
-                        resources.forEach((resource: any) => {
-                            definedNamespace = resource.namespace
-                        })
+                    if (resources) {
```
KevinFCormier, line 352: `resources` has type `any`, so we shouldn't really make any assumptions about its type. However, it appears to be understood that it is an array. In the old code, we assumed it was an array. In the new code, we assume it is an array AND that it has elements. 

Try this in the console:
`if ([]) console.log("truthy")`

We should at least check that `resources` has 1 element before we try to access it.

chenz4027: Sounds good, thank @KevinFCormier , will make the change :)

KevinFCormier: Also, could these resources be cluster-scoped and not have a namespace? The previous code would not have handled that either, but it would have taken the value from the last resource instead of the first.

chenz4027: Yeah they could be. For now the namespaces do exist, but I guess they could be possible to not have namespace. Given that, maybe for each was better?

chenz4027: Oh I can use get, and set it to undefined by default.

If no definedNamespace, it will use other spec

```                    definedNamespace ? definedNamespace : argoApp.spec.destination.namespace```

KevinFCormier (approved):

PR #2425, opened 2022-12-22
Allow copied yaml to be edited in the create policy wizard

JustinKuli:
Previously, if you copied in YAML from another policy you had already created (ie copied from the "Edit Policy" page), you were unable to edit many of the fields. Now, you can edit those fields as needed.

### DIFF
--- a/frontend/src/components/SyncEditor/process.test.js
+++ b/frontend/src/components/SyncEditor/process.test.js
@@ -1,0 +1,36 @@
+/* Copyright Contributors to the Open Cluster Management project */
+'use strict'
+
+import { processUser } from './process'
+
+describe('Test process.ts', () => {
+    afterAll(() => {
+        jest.resetAllMocks()
+    })
+
+    const policyYamlWithUid =
+        'apiVersion: policy.open-cluster-management.io/v1\n' +
+        'kind: Policy\n' +
+        'metadata:\n' +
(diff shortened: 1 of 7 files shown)


JustinKuli:
Finally ready for review! Hopefully I didn't do anything too unusual to get the test in...

/cc @mprahl @KevinFCormier @willkutler

frontend/src/components/SyncEditor/process.test.js
```diff
@@ -0,0 +1,36 @@
+/* Copyright Contributors to the Open Cluster Management project */
```
mprahl, line 1: Is there a particular reason that you didn't use TypeScript for this test file?

JustinKuli: Other test files in this directory and the TemplateEditor directory were just in js. But based on some TS test files I've found, it looks like I would just need to update the extension? I could do that if you want

mprahl: I prefer TypeScript but let's use JavaScript since that matches the other tests around it as you pointed out.

mprahl (approved):

willkutler (approved):

willkutler:
87.6%!! Congrats on beating sonarcloud @JustinKuli

jeswanke (approved):

PR #2427, opened 2022-12-22
Show alert for auto import failure

KevinFCormier:
https://github.com/stolostron/backlog/issues/24761
https://github.com/stolostron/backlog/issues/22815

- Validate kubeconfig for YAML syntax (could do more - check for specific values 
- Show error if automatic import fails
- Import code is hidden until the auto-import-secret is deleted; due to a backend change, the secret is only deleted after the import is successful or fails
image
image


RobDolares (approved):
/lgtm

PR #2438, opened 2023-01-05
Handle policy sets that are not placed on a cluster

mprahl:
When there are no placement decisions associated with the policy set or the policies within it, the console would crash when viewing the details of the policy set.


KevinFCormier (commented):
I would suggest making `status` an optional property of PlacementDecision so that TypeScript can catch this type of error.

https://github.com/stolostron/console/blob/main/frontend/src/resources/placement-decision.ts#L15

PR #2444, opened 2023-01-06
Update node/npm version in README prerequisities

jtomasek:
ACM now requires node 18 and npm 8


KevinFCormier (approved):
Thanks for updating that!

PR #2364, opened 2022-12-08
Hide cluster claim name for standalone clusters w/o a clusterpool and…

lennysgarage:
… hosted control planes
Signed-off-by: Jonathan Marcantonio <jmarcant@redhat.com>

This PR is for hiding the Cluster Claim name row in the clusterOverview page when the cluster is a standalone cluster with no cluster pool or is a hosted control plane.

Before:
Screenshot from 2022-12-08 11-57-34

After:
Screenshot from 2022-12-08 11-54-58

Addresses:
 - https://issues.redhat.com/browse/ACM-1947

PR #2445, opened 2023-01-06
ACM-1947: Hide cluster claim name for hosted clusters and hub

lennysgarage:
Signed-off-by: Jonathan Marcantonio <jmarcant@redhat.com>

This PR is for hiding the Cluster Claim name row in the clusterOverview page when the cluster is a standalone cluster with no cluster pool or is a hosted control plane. This is for patching a missed part of https://github.com/stolostron/console/pull/2364, where the hosted cluster was still being shown. 
Screenshot from 2023-01-06 10-57-29
Screenshot from 2023-01-06 10-57-56
As well removes the Cluster pool row when they are not from a cluster pool:
Screenshot from 2023-01-06 11-02-38

Addresses:
 - https://issues.redhat.com/browse/ACM-1947


KevinFCormier (commented):
We should never base logic on translated values. I am also not a fan of the hard-coded indices for splicing the arrays, and the various blocks of code that start from scratch for the left and right items. I know you were only following the pattern that had been established here.

I am proposing a refactored version this in this PR: https://github.com/stolostron/console/pull/2450

KevinFCormier:
Superseded by #2450

PR #2451, opened 2023-01-06
Add a warning when deleting a policy with shared placements

mprahl:
The first commit clarifies the wording in the delete policy modal .

The second commit adds a warning when deleting a policy with shared placements. When deleting a policy, you can select to delete the placements and placement bindings associated with it. These placements and placement bindings could be used by other policies, so a warning is added for this case so that the user is aware before actually deleting them.

image

Relates:
https://issues.redhat.com/browse/ACM-2412


KevinFCormier:
Approved. Will leave code review to @willkutler

willkutler (approved):

PR #2463, opened 2023-01-10
ACM-2690 Keep previous cached secret in case of syntax OR validation error

KevinFCormier:
https://issues.redhat.com/browse/ACM-2690

### DIFF
--- a/frontend/src/components/SyncEditor/process.ts
+++ b/frontend/src/components/SyncEditor/process.ts
@@ -190,7 +190,7 @@ const process = (
         if (secrets && !isEmpty(secrets)) {
             allSecrets = getMatchingValues(secrets, paths)
             allSecrets.forEach((value: { $p: string[] }) => {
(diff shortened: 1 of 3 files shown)


frontend/src/components/SyncEditor/process.ts
```diff
@@ -251,7 +256,7 @@ const process = (
         validate(validators, mappings, resources, validationErrors, syntaxErrors, protectedRanges)
     }
 
-    if (syntaxErrors.length !== 0 && validationErrors.length !== 0 && cachedSecrets && cacheFiltered) {
+    if ((syntaxErrors.length !== 0 || validationErrors.length !== 0) && cachedSecrets && cacheFiltered) {
```
KevinFCormier, line 259: @jeswanke This is the crucial part - validation errors are not assessed if there is a syntax error, so this code was not executed in that case.

jeswanke (approved):

PR #2464, opened 2023-01-10
Populate autoscale column for nodepools

o-farag:

https://issues.redhat.com/browse/ACM-2663

Autoscaling column shows `False` if autoscaling is disabled; the Min and Max size of the pool otherwise.
image
image


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/NodePoolsTable.tsx
```diff
@@ -86,17 +86,20 @@ const NodePoolsTable = ({ nodePools, clusterImages }: NodePoolsTableProps): JSX.
...
+            }
 
-        const min = autoscaling.min || 1
-        const max = autoscaling.max || 1
+            const min = autoscaling.min || 1
+            const max = autoscaling.max || 1
 
-        return `Min ${min} Max ${max}`
-    }, [])
+            return `Min ${min} Max ${max}`
```
KevinFCormier, line 99: This string needs to be externalized for translation.

o-farag: Done. Not sure if that was the best way to do it, let me know if there's a better way.

KevinFCormier: Probably better to use interpolation, like `t('Min {{min}} Max {{max}}', {min, max})`. When you commit, it should automatically add to translations.json and remove the old ones.

o-farag: Thanks, I tried something similar but couldn't get it to work

chenz4027 (approved):
/lgtm thanks Omar :)
/hold for @KevinFCormier to take another look

PR #2469, opened 2023-01-10
ACM-2286 Key ansible templates on index only (same template repeated allowed)

KevinFCormier:


KevinFCormier:
@jeswanke @RobDolares I propose that we override the test coverage requirement for this PR. Many of the uncovered lines are in code that is anticipating future support of scale and destroy operations for the ClusterCurator. That is not currently supported and is disabled.

RobDolares (approved):
/lgtm

PR #2471, opened 2023-01-11
Translation updated for Japanese, Korean and Chinese

rh-tchuang:
Translation updated for Japanese, Korean and Chinese

Signed-off-by: Terry Chuang <tchuang@redhat.com>


rh-tchuang:
closing PR and recreating to include sign-offs

PR #2504, opened 2023-01-23
ACM-2666 Reflect CP upgrade in UI right away

o-farag:
Issue: https://issues.redhat.com/browse/ACM-2666

Depends on https://github.com/stolostron/console/pull/2494

Peek 2023-01-23 13-28


o-farag:
/cc @chenz4027 
I checked the security hotspot, shouldn't be an issue since it doesn't take user input
Also, I made this PR with changes from https://github.com/stolostron/console/pull/2494 so not sure how merging it would work. Should I create another PR for this one when that one is merged?

o-farag:
https://github.com/stolostron/console/pull/2510

PR #2515, opened 2023-01-26
ACM-2914

o-farag:
Fixed issue where clusters with periods in their name would cause console crash on import

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -1497,6 +1497,7 @@
     "Instance type": "Instance type",
     "Internal server error": "Internal server error",
     "internet": "internet",
+    "invalidclustername.message": "The cluster name is invalid. Change the name to the RFC 1123 standard which consists of lower case alphanumeric characters, '-', and must start and end with an alphanumeric character.",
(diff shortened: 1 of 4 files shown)


fxiang1 (changes requested):
@o-farag in ACM the namespace for imported clusters must be the same as the cluster name. So I'm afraid we can't modify the namespace to a different value. Can you check the cluster name and display an error instead? ie. validate the cluster name against the suggested regex `[a-z0-9]([-a-z0-9]*[a-z0-9])?`

o-farag:
@fxiang1 I've changed it to that. Do you want something specific? At the moment, clicking import cluster just returns this error message:
image

I could change the error message and/or disable the import cluster dialogue all together instead

fxiang1:
@o-farag Thanks, I think that looks good! I do want to change the error message a bit to specify why the name is invalid.
Maybe we can use:
`The cluster name is invalid. Change the name to the RFC 1123 standard which consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character.`

fxiang1 (approved):
LGTM! Thanks Omar!

PR #2522, opened 2023-01-31
Include related resources tab in resource details page

zlayne:

https://issues.redhat.com/browse/ACM-1859

Screenshot 2023-01-31 at 9 16 53 AM


frontend/public/locales/en/translation.json
```diff
@@ -2541,6 +2541,7 @@
     "The private SSH key to use to access your cluster machines.": "The private SSH key to use to access your cluster machines.",
     "The public SSH key to use to access your cluster machines.": "The public SSH key to use to access your cluster machines.",
     "The pull secret that you obtained from the pull secret page on the Red Hat OpenShift Cluster Manager site. Use this pull secret to authenticate with the services that are provided by the included authorities, including Quay.io, which serves the container images for OpenShift Container Platform components.": "The pull secret that you obtained from the pull secret page on the Red Hat OpenShift Cluster Manager site. Use this pull secret to authenticate with the services that are provided by the included authorities, including Quay.io, which serves the container images for OpenShift Container Platform components.",
+    "The related resource search criteria matched too many resources, the results are truncated.": "The related resource search criteria matched too many resources, the results are truncated.",
```
KevinFCormier, line 2544: Comma-splice. Please have doc squad review new messages.

zlayne: @dockerymick Could I get your review on this new string? There are two instances using a similar string:
`The related resource search criteria matched too many resources, the results are truncated.`& 
`The search criteria matched too many resources, the results are truncated. Add more conditions to your search.`
For context these strings are displayed in a warning alert to notify users they are not seeing all of the search results as we have a 1000 item limit on all queries.

dockerymick: Thanks for the context! I recommend that you update the statements to read:

* Search result limit reached (1000 items per query). Your query results are truncated, add more filter conditions to your query.

zlayne: Thanks, I am thinking I will leave out the `1000 items per query` part as this could be customized, and later will most likely be changed. But will update the strings to:
`Search result limit reached. Your query results are truncated.` &
`Search result limit reached. Your query results are truncated, add more filter conditions to your query.`

The first string is used in a context where the user is not able to define the search parameters.

KevinFCormier: @dockerymick "Your query results are truncated, add more filter conditions to your query." is still a comma splice. Can we just use a period here instead of a comma?

dockerymick: @KevinFCormier thanks! sure, i am fine with that

dockerymick: Is there another way for users to view all of their resources? Or is "add more filter conditions to your query" the only solution?

dockerymick (approved):
The string updates look good to me !

PR #2526, opened 2023-02-01
Skip search query when OCP apps are filtered out

fxiang1:

https://issues.redhat.com/browse/ACM-2503

- Don't run OCP apps search query if they are filtered out
- Fix filter count for non default OCP apps


chenz4027 (approved):
/lgtm

Thanks Feng for the enhancement!

PR #2529, opened 2023-02-02
Mark more policies as externally managed

JustinKuli:
Policies are determined to be local or externally managed based on their managed fields. Previously, only exact values were checked, but the ArgoCD manager's name sometimes varies slightly, so some policies from Argo were being marked as "local". Now pattern matching is used to mark more policies correctly.

Refs:
 - https://issues.redhat.com/browse/ACM-2846

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

### DIFF
--- a/frontend/src/routes/Governance/common/util.tsx
+++ b/frontend/src/routes/Governance/common/util.tsx
@@ -286,9 +286,9 @@ export function getClustersSummaryForPolicySet(
 }
 
 export function resolveExternalStatus(policy: Policy) {
-  const knownExternalManagers = ['multicluster-operators-subscription', 'argocd-application-controller']
+  const knownExternalManagerPatterns = [/^multicluster-operators-subscription$/, /argocd/]
   const managedFields = policy.metadata.managedFields ?? []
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Governance/common/util.tsx
```diff
@@ -286,9 +286,9 @@ export function getClustersSummaryForPolicySet(
 }
 
 export function resolveExternalStatus(policy: Policy) {
-  const knownExternalManagers = ['multicluster-operators-subscription', 'argocd-application-controller']
+  const knownExternalManagerPatterns = [/multicluster-operators-subscription/, /argocd/]
```
mprahl, line 289: Could you make `multicluster-operators-subscription` be an exact match instead?

JustinKuli: Sure, but if they change the name of that, I get to say I told you so

mprahl (approved):

RobDolares (approved):
/lgtm

PR #2530, opened 2023-02-02
ACM-2661 Shows upgrade available under distribution version

o-farag:
Shows upgrade available under distribution version in details section
image

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ClusterDetails/ClusterOverview/ClusterOverview.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ClusterDetails/ClusterOverview/ClusterOverview.tsx
@@ -134,8 +134,13 @@ export function ClusterOverviewPageContent(props: {
     },
     distribution: {
       key: t('table.distribution'),
-      value: cluster?.distribution?.displayVersion && (
-        <DistributionField cluster={cluster} clusterCurator={clusterCurator} hostedCluster={hostedCluster} />
+      value: (
+        <DistributionField
+          cluster={cluster}
+          clusterCurator={clusterCurator}
+          hostedCluster={hostedCluster}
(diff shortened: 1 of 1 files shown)


fxiang1 (changes requested):
@o-farag I think we'll need `cluster?.distribution?.displayVersion &&` to prevent the `Upgrade available` link from displaying when the cluster is not ready.

I think this if statement:
https://github.com/stolostron/console/blob/main/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx#L397

is not triggering so the upgrade link is not displayed. Try looking there to see if you can figure out what's wrong.

o-farag:
@fxiang1 Not sure if I completely understand. Doesn't [this line](https://github.com/stolostron/console/blob/cb27a0713ca87cf89006d88bbf9480b1bf31668b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx#L151) prevent the `Upgrade available` link to be shown if `cluster?.distribution` doesn't exist?

As for https://github.com/stolostron/console/blob/main/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx#L397, I'm also not sure if I understand. This if statement is the same statement for all `Upgrade available` links i.e. in the nodepool table and managedcluster page etc and seems to work for me.

fxiang1:
@o-farag Sorry, you're right [this line](https://github.com/stolostron/console/blob/cb27a0713ca87cf89006d88bbf9480b1bf31668b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/DistributionField.tsx#L151) does prevent the `Upgrade available` link from being displayed and I missed the `resource={'hostedcluster'}` you added to `DistributionField`.

All looks good. Thanks!

fxiang1 (approved):

PR #2532, opened 2023-02-03
MGMT-13551 Move reconcilePause control to automationControlData

jtomasek:
`reconcilePause` control flag is currently only used by `onChangeAutomationTemplate` function inside `automationControlData`. The function relies on the control to be defined in the `controlData` otherwise an error occurs when trying to set `reconcilePause.active` value. Since the `reconcilePause` control was defined in `clusterDetailsControlData`, for the wizards (e.g. CIM or AI) which don't use this step the error occurred.

To fix this, we're moving the reconcilePause control to automationControlData which is only controlData which actually uses that flag.

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
@@ -386,11 +386,6 @@ export const clusterDetailsControlData = (t) => {
       type: 'hidden',
       active: false,
     },
-    {
-      active: '',
-      id: 'reconcilePause',
-      type: 'hidden',
-    },
   ]
 }
 
@@ -693,6 +688,11 @@ export const automationControlData = (t) => {
       id: 'toweraccess-destroy',
       type: 'hidden',
       active: '',
+    },
+    {
+      active: '',
+      id: 'reconcilePause',
+      type: 'hidden',
     },
   ]
 }
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/__snapshots__/ControlDataAI.test.js.snap
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/__snapshots__/ControlDataAI.test.js.snap
@@ -135,6 +135,11 @@ Array [
     "type": "hidden",
   },
   Object {
+    "active": "",
+    "id": "reconcilePause",
+    "type": "hidden",
+  },
+  Object {
     "comment": "Ensure these settings are correct. The saved cluster draft will be used to determine the available network resources. Therefore after you press Save you will not be able to change these cluster settings.",
     "disableEditorOnSuccess": true,
     "disablePreviousControlsOnSuccess": true,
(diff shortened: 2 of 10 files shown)


jtomasek:
Note that a [hive.openshift.io/reconcile-pause](https://github.com/openshift/hive/blob/40fcaecc31022b16fcf8b5669a7344240c7065cc/pkg/constants/constants.go#L137) annotation is used on `clusterDeployment` resource (in hive-template.hbs) and in the code it is only set to true when there is an automation selected. I assume it is required by the [cluster-curator-controller](https://github.com/stolostron/cluster-curator-controller/blob/main/CONTRIBUTING.md#controller)

Since we're allowing to use ansible automation in Hypershift which is not Hive based and uses hostedCluster resource instead of clusterDeployment, I wonder of there could be any potential issue with reconciliation not being paused in that case.

KevinFCormier:
@jtomasek Good catch. I agree with this fix.

I think you are also right that there is an issue for HyperShift. I suspect that the `install` actions will not work properly, because `ClusterCurator` cannot delay the provisioning of the cluster until after the pre-install hooks are run, and it may even be looking for a `ClusterDeployment` from which to remove the `hive.openshift.io/reconcile-pause` annotation. I would guess that the `upgrade` actions are fine, since the `ClusterCurator` controller is the one to actually initiate the upgrade. @mikeshng Can you comment on this?

dtthuynh:
> Note that a [hive.openshift.io/reconcile-pause](https://github.com/openshift/hive/blob/40fcaecc31022b16fcf8b5669a7344240c7065cc/pkg/constants/constants.go#L137) annotation is used on `clusterDeployment` resource (in hive-template.hbs) and in the code it is only set to true when there is an automation selected. I assume it is required by the [cluster-curator-controller](https://github.com/stolostron/cluster-curator-controller/blob/main/CONTRIBUTING.md#controller)
> 
> Since we're allowing to use ansible automation in Hypershift which is not Hive based and uses hostedCluster resource instead of clusterDeployment, I wonder of there could be any potential issue with reconciliation not being paused in that case.

Wait, we are allowing automation template with hypershift? Is this support stated somewhere?

mikeshng:
> @jtomasek Good catch. I agree with this fix.
> 
> I think you are also right that there is an issue for HyperShift. I suspect that the `install` actions will not work properly, because `ClusterCurator` cannot delay the provisioning of the cluster until after the pre-install hooks are run, and it may even be looking for a `ClusterDeployment` from which to remove the `hive.openshift.io/reconcile-pause` annotation. I would guess that the `upgrade` actions are fine, since the `ClusterCurator` controller is the one to actually initiate the upgrade. @mikeshng Can you comment on this?

Kevin you are correct. I gone through ClusterCurator controller code. The only place that gets and delete (but never add) the pause reconcile annotation is in the "install" path. The "upgrade" path doesn't seem to interact with the ClusterDeployment resource, it interacts with the managed cluster's ClusterVersion resource. If there is a prehook for the upgrade then it's blocked by the ClusterCurator controller as Kevin stated.

CC @xiangjingli @elgnay

elgnay:
The `install` curation does not work for HyperShift cluster because the cluster is not provisioned by using Hive. The curator-contoller has no way to delay the cluster provisioning as Kevin mentioned, while the 'upgrade' curation does work for HyperShift cluster.

KevinFCormier:
@elgnay Oh, but actually - I think in the UI the "traditional" upgrade is disabled for hosted cluster, and replaced with node pool upgrades. @fxiang1 can you confirm?

fxiang1:
@KevinFCormier Right, for Hypershift cluster upgrades we don't use hive or the curator. We just update the image value in either the `HostedCluster` or `NodePool` CR and let the Hypershift controller handle the upgrade.

PR #2533, opened 2023-02-03
[2.7] MGMT-13551 Move reconcilePause control to automationControlData

jtomasek:
`reconcilePause` control flag is currently only used by `onChangeAutomationTemplate` function inside automationControlData. The function relies on the control to be defined in the controlData otherwise an error occurs when trying to set reconcilePause.active value. Since the reconcilePause control was defined in clusterDetailsControlData, for the wizards (e.g. CIM or AI) which don't use this step the error occurred.

To fix this, we're moving the reconcilePause control to automationControlData which is only controlData which actually uses that flag.

Cherry-picked from 405b2654


KevinFCormier:
@crizzo71 Can we get approval to merge this in 2.7.1? This is the UI crash when selecting an automation template for AI clusters.

PR #2536, opened 2023-02-06
ACM-2411

o-farag:
Fix text overflow on smaller screens/larger zoom

image
image


o-farag:
/cc @fxiang1 
I commented out the if statements for the tests for the window resize and zoom, didn't know how to simulate window resizing and what to check for, and I don't think it's that important to test. Lmk if you still think I should add tests for it

fxiang1 (approved):
I don't think we need to test resizing the window so this looks good to me. Thanks @o-farag !

PR #2538, opened 2023-02-07
ACM-3214 

o-farag:
Filter out versions before 4.11 when adding nodepools via ui


fxiang1 (approved):
LGTM! Thanks Omar!

PR #2541, opened 2023-02-08
ACM-2863 Fix create application button in dark mode

lennysgarage:
Aligns the 'Create application' button so it works properly in dark mode.
Before:
Screenshot from 2023-02-08 16-00-13

After: 
Screenshot from 2023-02-08 15-59-06

The prop, isPlain was already being set by the useStyles() func above.

Addresses:
 - https://issues.redhat.com/browse/ACM-2863

Signed-off-by: Jonathan Marcantonio <jmarcant@redhat.com>

### DIFF
--- a/frontend/src/ui-components/AcmDropdown/AcmDropdown.tsx
+++ b/frontend/src/ui-components/AcmDropdown/AcmDropdown.tsx
@@ -170,7 +170,6 @@ export function AcmDropdown(props: AcmDropdownProps) {
           )
         }
         isOpen={isOpen}
-        isPlain={props.isPlain}
       />
     </TooltipWrapper>
   )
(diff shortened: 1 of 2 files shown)


frontend/src/ui-components/AcmDropdown/AcmDropdown.tsx
```diff
@@ -170,7 +170,6 @@ export function AcmDropdown(props: AcmDropdownProps) {
           )
         }
         isOpen={isOpen}
-        isPlain={props.isPlain}
```
fxiang1, line 173: @lennysgarage We can't remove this from the shared component AcmDropdown as it might affect other components using it. Can you find where in App UI we're using AcmDropdown for the create button and set the isPlain prop to `false`?

fxiang1 (changes requested):

frontend/src/ui-components/AcmDropdown/AcmDropdown.tsx
lennysgarage: Looks like it was being set in Overview.tsx, updated it now!

fxiang1 (approved):
LGTM! Thanks Jonathan!

PR #2548, opened 2023-02-10
Fix search issue in cluster list and application

TimothyAsirJeyasing:
Signed-off-by: Timothy Asir <tjeyasin@redhat.com>

### DIFF
--- a/frontend/src/routes/Home/Search/components/Searchbar.tsx
+++ b/frontend/src/routes/Home/Search/components/Searchbar.tsx
@@ -104,7 +104,12 @@ export function Searchbar(props: SearchbarProps) {
   }, [currentQuery, savedSearchLimit, savedSearchQueries])
 
   /** callback for updating the inputValue state in this component so that the input can be controlled */
-  const handleInputChange = (value: string) => {
+  const handleInputChange = (input: any) => {
+    // **Note: PatternFly change the fn signature
+    // From: (value: string, event: React.FormEvent<HTMLInputElement>) => void
+    // To: (_event: React.FormEvent<HTMLInputElement>, value: string) => void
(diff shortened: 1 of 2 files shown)


TimothyAsirJeyasing:
@GowthamShanmugam Please review

TimothyAsirJeyasing:
Screenshot from 2023-02-10 18-38-52
Screenshot from 2023-02-10 18-38-19

KevinFCormier:
@TimothyAsirJeyasing Thanks for the PR. I assume you are testing on OpenShift console 4.13?

@zlayne it occurred to me that this is related to something that was mentioned at the last PatternFly community meeting. As part of the final release of PF 4, they changed the signature for the event handler callbacks of some beta components to make them all consistent with the event as the first argument.

However, some versions of OpenShift console that we support will not get this update. So we need to try to support whether the arguments in either order. Can you please try to adjust this PR to do that?

zlayne:
@KevinFCormier @TimothyAsirJeyasing I would propose the following for backwards compatibility, please let me know what you think? Would work for both the Searchbar and AcmTable instances. 
```
  /** callback for updating the inputValue state in this component so that the input can be controlled */
  const handleInputChange = (input: any) => {
    // **Note: PatternFly change the fn signature
    // From: (value: string, event: React.FormEvent<HTMLInputElement>) => void
    // To: (_event: React.FormEvent<HTMLInputElement>, value: string) => void
    // both cases need to be handled for backwards compatibility
    const value = typeof input === 'string' ? input : (input.target as HTMLInputElement).value
    const delimiters = [' ', ':', ',']
    if (delimiters.indexOf(value) < 0) {
      // Delimiters are only used to enter chips - do not allow for entry in input
      setInputValue(value)
    }
  }
```

PR #2561, opened 2023-02-13
ACM-2112 Fix filter value clash

KevinFCormier:
Allow different filters to have options with the same value, e.g. cluster status/node status/add-on status might all use `Unknown` as a value.

Also fixes keyboard navigability for the filter select menu.

### DIFF
--- a/frontend/jest.config.ts
+++ b/frontend/jest.config.ts
@@ -48,7 +48,7 @@ const config: Config.InitialOptions = {
   ci: true,
   collectCoverage: true,
   coverageDirectory: './coverage',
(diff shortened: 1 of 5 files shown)


KevinFCormier:
@dtthuynh @Napoco @atifshafi Please note, this PR changes the test IDs for checkboxes in filter dropdowns. For one test case, I updated it to use getByRole instead. For another test case, I updated the test IDs to include the filter name. Just a heads up in case there are tests that need updating based on this change.

chenz4027 (approved):
/lgtm
Thanks Kevin!

zlayne (approved):

PR #2562, opened 2023-02-13
Fix case where users use same name different namespace for resources 2.8

fxiang1:

https://issues.redhat.com/browse/ACM-3337

- Add namespace to the model key to make it more unique for when users use the same resource name but different namespaces in their apps


chenz4027 (approved):
/lgtm
👍

PR #2570, opened 2023-02-14
ACM-3244 Missing breadcrumbs

jeswanke:


KevinFCormier (changes requested):
@jeswanke These breadcrumbs cause a full page reload when clicked, unlike the existing pages. We need to refactor the code so that these wizards use the same breadcrumb code as other pages.

Please also make sure strings match:
image

jeswanke:
now uses react-dom Link

KevinFCormier:
@jeswanke what I had in mind was instead of using `<WizardPage>`, use just `<Wizard>` directly, placing it in an `AcmPage` like our other wizard/editor pages.

jeswanke:
https://github.com/stolostron/console/pull/2570#issuecomment-1432147297

fixed and fixed

PR #2581, opened 2023-02-16
ACM-2870 & ACM-3946: Placement migration

chenz4027:
https://issues.redhat.com/browse/ACM-2870
https://issues.redhat.com/browse/ACM-3946


frontend/src/routes/Applications/ApplicationDetails/ApplicationTopology/model/applicationSubscription.js
```diff
@@ -205,38 +209,80 @@ const buildSubscriptionMaps = (subscriptions, modelSubscriptions) => {
           subscription.rules = []
         }
       })
+
+    // ditto for placements and placmentrules
```
fxiang1, line 213: @chenz4027 For this, I'm wondering if you can use the same `get(subscription, 'spec.placement.placementRef.name', '')` above, you can check the kind to determine which map to use. This way we won't duplicate the code and it won't be putting PlacementRules in placementsMap.

fxiang1 (changes requested):

frontend/src/routes/Applications/ApplicationDetails/ApplicationTopology/model/applicationSubscription.js
chenz4027: Ah thanks @fxiang1 , will work on it

chenz4027: Combined the loop to the same one. @fxiang1 I notice that the subscription.rules might be a bit confusing as it's now changed to look after placement decision resources so that I changed it to subscription.decisions. Subscription.placements will contain both placement and placement rule resources.

fxiang1 (approved):
LGTM! Thanks Maggie!

PR #2594, opened 2023-02-21
ACM-3663: Fix breadcrumbs

chenz4027:
https://issues.redhat.com/browse/ACM-3663


frontend/public/locales/en/translation.json
```diff
@@ -975,6 +976,8 @@
   "creation.valid.numeric": "Value must be integer.",
   "Credential name": "Credential name",
   "Credential name and namespace are predefined as below for HyperShift add-on": "Credential name and namespace are predefined as below for HyperShift add-on",
+  "Credential type": "Credential type",
+  "Credential Type": "Credential Type",
```
KevinFCormier, line 980: Can you please use only "Credential type"? Otherwise looks good.

chenz4027: Thanks Kevin! Will make the change

chenz4027: Updated :)

PR #2597, opened 2023-02-21
ACM-3493 editing the name of a policy placement

jeswanke:


frontend/src/wizards/WizardPage.css
```diff
@@ -0,0 +1,3 @@
+.wizard-page .pf-c-drawer__panel {
+  transition: none;
+}
```
KevinFCormier, line 3: Is this for consistency with temptifly? If so, might it be nicer to add it there instead? Or is it causing a problem?

frontend/src/routes/Governance/policies/EditPolicy.tsx
```diff
@@ -30,6 +30,7 @@ export function WizardSyncEditor() {
       resources={resources}
       schema={schema}
       filters={['*.metadata.managedFields']}
+      immutables={['PlacementBinding.0.subjects.0.*']}
```
JustinKuli, line 33: This makes it look way easier than I was expecting! I wonder if the entire PlacementBinding should be immutable... It is helpful to see the binding on the "Edit Policy" page - just to have a full view of what is necessary for the Policy - but I think we might want to discourage changing any of its fields because it's really easy to mess something up.

frontend/src/wizards/WizardPage.css
KevinFCormier: John says it animates even when it should just be open. No PF switch to disable.

frontend/src/routes/Governance/policies/EditPolicy.tsx
jeswanke: sounds good :)

PR #2602, opened 2023-02-22
ACM-3450 - Adding missing translation function use

Randy424:
Regarding: [ACM-3450](https://issues.redhat.com/browse/ACM-3450)


KevinFCormier:
Manually verified all tests passed. Bypassing broken SonarCloud.

PR #2605, opened 2023-02-23
ACM-3661

o-farag:
provide option to clear app filters

Before: No applications matching a filter would prevent user from deleting filter(s)
image

After:
image


fxiang1:
I'm good with the changes. Let's see if @KevinFCormier has any concerns.

KevinFCormier:
This is a consequence of an optimization that was put in to not bother performing search queries for application data if the user is filtering for only subscription apps. 

Another consequence of this is that we cannot tell if there are other types of apps (they don't show in the filter) if we are already filtering on subscription apps.

I'm wondering if we could tweak the optimization a bit so that:
1. We perform the query regardless of filtering.
2. We only wait for the search query to complete if we are not filtering out apps that come from search.
3. We use the `startPolling` and `stopPolling` functions in response to changes in the filters to manage when we need fresh data.

The downside to this is that we might see the empty state for a moment until search finishes, then we would see that we are filtering and can clear the filter to see more results. We could consider trying to load these queries upfront instead, so that as soon as the user accesses any ACM page, we get an initial query result for these, and the page would then wait until it has a result to load.

Changing the table so that we never say there are no items if a filter is applied would be my second choice here, but if we go with that option, I'm sure we can do that with a few lines of code - no need to duplicate a big chunk.

fxiang1:
Thanks @KevinFCormier, we can try that.

@o-farag Let me know if you want attempt what Kevin suggested. I can show you which files to look at.

o-farag:
https://github.com/stolostron/console/pull/2610

PR #2606, opened 2023-02-27
UI fix of NavItem highlighting

nirrozenbaum:
in ACM 2.5+, when opening the "clusters" menu tab (sub-item of Infrastructure), the opened URL does not end with /clusters, therefore the isActive member of NavItem would always return false.  In ACM 2.5 line 338's isActive check was the same as in 335 and 344, which was fixed in ACM 2.6+. The bug persists for NavItem's that are either a group or not children of a group. The suggested PR is a fix for the above.

steps to reproduce the bug:
1. Expand Infrastructure NavItem
2. Open clusters sub-NavItem
3. Close Infrastructure NavItem

Bugged behavior:
Infrastructure NavItem is not highlighted 

Wanted behavior:
Infrastructure NavItem is highlighted (same as all other cases, e.g., Welcome).

This bug exists also in upstream.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -332,7 +332,7 @@ function AppSidebar(props: { routes: (IRoute | IRouteGroup)[] }) {
                   key={route.title}
                   title={route.title}
                   isExpanded
-                  isActive={!!route.routes.find((route) => location.pathname.startsWith(route.route)}
+                  isActive={!!route.routes.find((route) => location.pathname.startsWith(route.route))}
                 >
                   {route.routes.map((route) => (
                     <NavItem key={route.route} isActive={location.pathname.startsWith(route.route)}>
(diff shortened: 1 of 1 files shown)


nirrozenbaum:
https://user-images.githubusercontent.com/19717747/221602432-c02e83c1-7d1e-4666-806b-d4ee5c4ec86b.mov

KevinFCormier:
@nirrozenbaum Thanks for the fix. For 2.7+, this code is only used for development. Do you feel its important to fix in 2.5 and 2.6?

nirrozenbaum:
@KevinFCormier personally I would fix it for consistent UI behavior.
if the underneath code is extremely good but users see UI bugs, they might get the wrong impression that code is buggy.
the fix is also very simple and straight forward.

up to you of course :)

PR #2613, opened 2023-02-28
ACM-3947 fix placement query apiversion

o-farag:
Placement api version was outdated (hardcoded) causing query to fail

Before:
image

After:
image


fxiang1 (approved):
LGTM! Thanks Omar!

PR #2616, opened 2023-03-01
Add hypershift diagram and split cluster create catalog

fxiang1:
https://issues.redhat.com/browse/ACM-2472

- Update to react-data-view 1.7.0 to pull in the expandable section feature in ItemView
- Split create cluster catalog between ones with creds and one without
- Add diagram for hypershift to provide a better explanation
- Remove tech preview badge for AWS

image

We pushed back on the design to have a HR as part of the expandable section toggle text so the compromise is to just have a HR above the expandable section, this way we don't need to have a complex custom css:
image


chenz4027 (approved):
/lgtm

Thanks Feng!

chenz4027 (approved):
/lgtm
Sure!

PR #2619, opened 2023-03-01
ACM-3787 change placement apiversion for appset delete

o-farag:
change placement apiversion for appset delete

Outdated version of placement apiversion meant placements were not deleted with appset


chenz4027 (approved):
/lgtm

PR #2620, opened 2023-03-01
ACM-3787 (2.7) fetch placement apiversion for appset delete

o-farag:
fetch placement apiversion for appset delete

Outdated version of placement apiversion meant placements were not deleted with appset


chenz4027 (changes requested):
Could we change it to PlacementApiVersionBeta so that we don't reference the api manually?

chenz4027 (approved):
/lgtm

chenz4027:
@o-farag 2.7.1 is code freeze now, we will need to wait until 2.7.2 is open
image

PR #2621, opened 2023-03-02
ACM-3447 update react-data-view for default string context

RobDolares:
related issues: 
https://issues.redhat.com/browse/ACM-3447
https://issues.redhat.com/browse/ACM-3480


frontend/src/lib/dataViewStrings.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const useDataViewStrings = (): DataViewStrings => {
+  const { t } = useTranslation()
+  return useMemo(
+    () => ({
+      backLabel: t('Back'),
+      cancelLabel: t('Cancel'),
+      noItemsString: (plural: string) => t('No {{plural}} yet', { plural }),
+      getStartedMessage: (article: string, singular: string) =>
+        t('To get started, create {{article}} {{singular}}.', { article, singular }),
```
KevinFCormier, line 15: We don't want to substitute translations into translations. I think you forgot to update this after we changed it in `react-data-view`. These two fields are both just strings.
image

frontend/src/routes/Credentials/CreateCredentialsCatalog.tsx
```diff
@@ -70,13 +73,15 @@ export function CreateCredentialsCatalog() {
...
       }
     >
-      <ItemView
-        items={cards}
-        itemKeyFn={keyFn}
-        itemToCardFn={(card) => card}
-        onBack={back(NavigationPath.credentials)}
-        onCancel={cancel(NavigationPath.credentials)}
-      />
+      <DataViewStringContext.Provider value={dataViewStrings || defaultStrings}>
```
KevinFCormier, line 76: There's no need to pass in `defaultStrings` here or in any of the spots where you've added `<DataViewStringContext.Provider>`. `dataViewStrings` is never undefined.

PR #2625, opened 2023-03-02
ACM-2210 console api endpoint

jeswanke:


backend/src/routes/ansibletower.ts
```diff
@@ -33,6 +36,11 @@ export function ansibleTower(req: Http2ServerRequest, res: Http2ServerResponse):
       return respondBadRequest(req, res)
     }
 
+    // whitelist of apis our ui calls
```
KevinFCormier, line 39: Please use the term "allow list" here instead.

backend/src/routes/ansibletower.ts
```diff
@@ -13,6 +13,9 @@ interface AnsibleCredential {
   token: string
 }
 
+// must match ansiblePaths in frontend/src/resources/utils/resource-request.ts
+const ansiblePaths = ['/api/v2/job_templates/', '/api/v2/workflow_job_templates/']
+
 export function ansibleTower(req: Http2ServerRequest, res: Http2ServerResponse): void {
   const token = getToken(req)
   if (!token) return unauthorized(req, res)
```
KevinFCormier, line 21: Are we going to check if the token is valid? That was the complaint of the defect.

KevinFCormier: And this was what I was getting at in terms of checking all our APIs to make sure we handle them consistently.

backend/src/routes/ansibletower.ts
```diff
@@ -13,6 +13,9 @@ interface AnsibleCredential {
   token: string
 }
 
+// must match ansiblePaths in frontend/src/resources/utils/resource-request.ts
+const ansiblePaths = ['/api/v2/job_templates/', '/api/v2/workflow_job_templates/']
```
KevinFCormier, line 17: Please be aware that once this PR is merged, backend changes will be required to access new Ansible Tower APIs @RobDolares. (Related to [ACM-3842](https://issues.redhat.com/browse/ACM-3842))

jeswanke:
for new ansible api calls we just need to add to that list--just don't want anyone using our console-api to use the /screwyouup path

PR #2627, opened 2023-03-03
ACM-2867 add rbac for emptystate add cred and create automation template

RobDolares:
related issue: https://issues.redhat.com/browse/ACM-2867

### DIFF
--- a/frontend/src/routes/Credentials/CredentialsPage.tsx
+++ b/frontend/src/routes/Credentials/CredentialsPage.tsx
@@ -291,6 +291,8 @@ export function CredentialsTable(props: {
               history.push(createBackCancelLocation(NavigationPath.addCredentials))
             },
             variant: ButtonVariant.primary,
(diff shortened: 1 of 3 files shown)


KevinFCormier (changes requested):
Thanks, Rob. This is a great start. I see the buttons for the empty states are working as expected.
image

We also need to handle the table action buttons.
image

It would be great if you can add the check for those as well, but try not to duplicate any code. If we ever need to tweak the checks, ideally we do that in one place.

Note that the default `view` ClusterRole does not allow viewing of Secrets. To test this, I created a new `view-secrets` ClusterRole and bound it to the user `e2e-view-ns` in the namespace `e2e-rbac-test-1`.

KevinFCormier (changes requested):
So it looks like you got the tableActionButton for Automation templates:
image


But not for Credentials:
image

PR #2629, opened 2023-03-03
Increase search items limit from 1k to 10k

zlayne:
/hold


jlpadilla:
@zlayne Could we use an env variable to change this limit?
We want to keep increasing this limit in the near future as performance allows. An ENV would help us test higher limits or we can advise customers to reduce it if we encounter any problems.

zlayne:
@jlpadilla I have included the changes to use `SEARCH_QUERY_LIMIT` as an env variable to manage the limit.

frontend/src/routes/Home/Search/SearchPage.tsx
```diff
@@ -131,19 +132,20 @@ function RenderSearchBar(props: {
...
-  }, [currentSearch])
+  }, [currentSearch, searchQueryLimit])
 
   const searchCompleteResults = useSearchCompleteQuery({
     skip: !currentSearch.endsWith(':') && !operators.some((operator: string) => currentSearch.endsWith(operator)),
     client: process.env.NODE_ENV === 'test' ? undefined : searchClient,
     variables: {
       property: searchCompleteValue,
       query: searchCompleteQuery,
+      limit: 10000,
```
jlpadilla, line 148: Should read from env. `useSearchQueryLimit()`

zlayne: updated.

jlpadilla (approved):
/lgtm

PR #2641, opened 2023-03-07
Fix search query polling bug for app table

fxiang1:
https://issues.redhat.com/browse/ACM-4117

- Fix polling logic for app table


chenz4027 (approved):
/lgtm

PR #2648, opened 2023-03-09
UserPreference CRUD auth should use console serviceaccount

zlayne:


KevinFCormier (changes requested):
@zlayne I just did a quick review, but I didn't notice authentication being used. Please see John's PR which just merged that protects all routes by checking for a logged in token:

https://github.com/stolostron/console/pull/2625

In addition, I think the user should only be able to work with their own `UserPreference` resource. Instead of having the frontend first fetch the username and then use that in the userpreference call, could we have the userpreference call just look this up and only work with the user's resource?

PR #2649, opened 2023-03-09
ACM-2361 check for ACM before enabling all addons

o-farag:
Tweak previous fix by checking for acm before enabling addons, and then enabling all addons


fxiang1 (approved):
LGTM! Thanks Omar!

PR #2663, opened 2023-03-14
ACM-3473 ACM-3998 ACM-4213 Support Argo apps without using resourceTracking labels

fxiang1:
https://issues.redhat.com/browse/ACM-4213
https://issues.redhat.com/browse/ACM-3998
https://issues.redhat.com/browse/ACM-3473

Since ArgoCD 2.2, Argo introduced app resource tracking using annotations which means users can turn off label tracking completely. This caused an issue for the topology since we were dependent on the tracking label being added to Argo app resources.

I investigated using the new Argo annotation resource tracking method. It seemed promising but still would not work in the case when there are multiple Argo instances and apps with the same names are used. So the option I went with is to just search for all resource kinds in the deployed namespaces. 

- Added multi-namespace support for Argo/AppSet apps
- Removed the use of Argp resourceTracking label
- Fix Argo related apps not displaying properly


chenz4027 (approved):
/lgtm
Thanks Feng for for the fixes!

PR #2664, opened 2023-03-15
CIM: Pass new props for Infraenv error updates

rawagner:
Requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1978


rawagner:
included in https://github.com/stolostron/console/pull/2695

PR #2668, opened 2023-03-15
ACM-3837 Use importing/importfailed for auto-import clusters

KevinFCormier:
https://issues.redhat.com/browse/ACM-3837

Also addresses https://issues.redhat.com/browse/ACM-4303


Randy424 (approved):
/lgtm

PR #2669, opened 2023-03-16
MGMT-13262: CIM/Hypershift: Enable hostPrefix and stop using podCIDR (deprecated)

rawagner:
Requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1982


rawagner:
included in https://github.com/stolostron/console/pull/2695

PR #2677, opened 2023-03-17
ACM-4186 Add condition for hosted cluster import failed

o-farag:
Importing hosted clusters with invalid name will no longer get stuck on importing

Before:

After:
image
image

### DIFF
--- a/frontend/src/resources/utils/get-cluster.ts
+++ b/frontend/src/resources/utils/get-cluster.ts
@@ -34,6 +34,7 @@ import {
   checkCuratorConditionDone,
   checkForRequirementsMetConditionFailureReason,
   checkForCondition,
+  getConditionReason,
 } from './status-conditions'
 
 export enum ClusterStatus {
@@ -65,6 +66,7 @@ export enum ClusterStatus {
   'draft' = 'draft',
   'running' = 'running',
   'upgradefailed' = 'upgradefailed',
+  'error' = 'error',
 }
 
 export const getClusterStatusLabel = (status: ClusterStatus | undefined, t: TFunction) => {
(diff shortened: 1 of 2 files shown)


frontend/src/resources/utils/get-cluster.ts
```diff
@@ -65,6 +66,7 @@ export enum ClusterStatus {
   'draft' = 'draft',
   'running' = 'running',
   'upgradefailed' = 'upgradefailed',
+  'error' = 'error',
```
KevinFCormier, line 69: Doesn't look like this is used?

o-farag: Oops, forgot to remove that

frontend/src/resources/utils/get-cluster.ts
```diff
@@ -1302,7 +1304,14 @@ export function getClusterStatus(
     )
 
     if (managedCluster && HostedClusterReadyStatus?.status === 'True') {
-      mcStatus = ClusterStatus.importing
+      const stuckImport =
+        checkForCondition('HubAcceptedManagedCluster', mcConditions) &&
+        getConditionReason('HubAcceptedManagedCluster', mcConditions) === 'Error'
+      if (stuckImport) {
+        mcStatus = ClusterStatus.importfailed
```
KevinFCormier, line 1311: Is there a message from one of the conditions that you could put into `statusMessage` so the user has some more information on what the problem is here?

o-farag: I could add this, is it too verbose?
image

KevinFCormier: I think it's much better than having no explanation for the failure You can see what the problem is if you read it carefully.

o-farag:
/cc @fxiang1 
Despite the discussion on Jira, I thought it would b e a good idea to also add a visual to show a stuck cluster import because user error is still possible when creating the CRs manually.

fxiang1 (approved):
Looks great Omar! Thanks for adding this fix!

PR #2684, opened 2023-03-21
Integrate Credentials with InfraEnv form

rawagner:
Requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1989


rawagner:

PR #2686, opened 2023-03-21
String updates from Oliver

KevinFCormier:


oafischer (approved):
Thanks for working those in as well!

PR #2693, opened 2023-03-21
Search autocomplete should have separate limit than search results

zlayne:


backend/config/SEARCH_RESULT_LIMIT
```diff
@@ -0,0 +1 @@
+10000
```
jlpadilla, line 1: This limit was 1,000 in 2.7. 
The main factor was slow performance calculating relationships for 10,000 results. We need to test that scenario before increasing. We could also explore UX alternatives because relationships aren't useful for a large result set.

zlayne: Correct, we have now bumped the limit from 1k to 10k in both 2.7 & 2.8.
I have not tested with a search result that contains 10k as of yet, but have tested with around 6500. We now use the searchResultRelatedCount query when expanding the related section and it has been returning in ~1sec. 

I am interested in understanding how useful customers think the related section is on the main search results page, as I agree it is not practical to show related results when there are a large number of search results.

jlpadilla (approved):
/lgtm

PR #2699, opened 2023-03-23
Set Placement as the default in the policy wizards

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-4063

### DIFF
--- a/frontend/src/routes/Governance/policy-sets/CreatePolicySet.test.tsx
+++ b/frontend/src/routes/Governance/policy-sets/CreatePolicySet.test.tsx
@@ -7,15 +7,9 @@ import { nockIgnoreRBAC, nockCreate, nockIgnoreApiPaths } from '../../../lib/noc
 import { waitForNocks, waitForText } from '../../../lib/test-util'
 import { NavigationPath } from '../../../NavigationPath'
 import { CreatePolicySet } from './CreatePolicySet'
-import {
-  mockPolicySets,
-  mockPolicy,
-  mockNamespaces,
-  mockPlacementRules,
-  mockManagedClusters,
-  mockPlacementBindings,
-} from '../governance.sharedMocks'
+import { mockPolicySets, mockPolicy, mockNamespaces, mockManagedClusters } from '../governance.sharedMocks'
 import userEvent from '@testing-library/user-event'
+import { Placement, PlacementBinding } from '../../../resources'
 
 function TestCreatePolicySet() {
   return (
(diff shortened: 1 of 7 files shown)


frontend/src/routes/Governance/policy-sets/CreatePolicySet.test.tsx
```diff
@@ -74,19 +68,66 @@ describe('Create Policy Page', () => {
       nockCreate(mockPolicySets[2]),
     ]
 
-    const placementRuleNock = [
-      nockCreate(mockPlacementRules[1], undefined, 201, { dryRun: 'All' }), // DRY RUN
-      nockCreate(mockPlacementRules[1]),
+    const mockPlacement: Placement = {
```
dhaiducek, line 71: Nit: I'd prefer these be kept in the `governance.sharedMocks.tsx` file to keep with existing conventions.

frontend/src/wizards/Policy/PolicyWizard.tsx
```diff
@@ -177,7 +176,7 @@ export function PolicyWizard(props: {
           existingClusterSetBindings={props.clusterSetBindings}
           bindingSubjectKind={PolicyKind}
           bindingSubjectApiGroup={PolicyApiGroup}
-          defaultPlacementKind={props.defaultPlacementKind ?? PlacementRuleKind}
+          defaultPlacementKind={PlacementKind}
```
dhaiducek, line 179: Optional/Future: It looks to me like GRC is the only set of wizards that uses the `PlacementSection` component, so we can get rid of this property entirely here and make changes there.

dhaiducek (commented):
Looks good--some minor comments to consider.

frontend/src/routes/Governance/policy-sets/CreatePolicySet.test.tsx
dhaiducek: So...do you want to add a Placement/PlacementBinding mock generation function to the sharedMocks file to use here? 🙂

dhaiducek (approved):
/hold in case the mock generation function should be implemented in this PR

frontend/src/routes/Governance/policy-sets/CreatePolicySet.test.tsx
mprahl: For the sake of time, I think I'll defer that for now. :smile:

PR #2704, opened 2023-03-24
update got pkg

RobDolares:
related issue: https://issues.redhat.com/browse/ACM-3171


RobDolares:
redundant due to fast-forwards

PR #2707, opened 2023-03-27
ACM-2666 Add condition for hcp upgrade true

o-farag:
The `hostedCluster?.status?.version?.desired.image` condition used before doesn't get updated right away anymore for some reason. Seems like `hostedCluster.spec.release.image` gets updated right away, so use that as another condition to show HCP upgrading
image


fxiang1 (approved):
LGTM! Thanks Omar!

PR #2708, opened 2023-03-27
ACM-4592 - Transparent template summary

Randy424:
Jira: [ACM-4592](https://issues.redhat.com/browse/ACM-4592)

Before:
Screenshot 2023-03-27 at 10 58 27 AM

After:
(Proposed color change)

### DIFF
--- a/frontend/src/components/TemplateSummaryModal.tsx
+++ b/frontend/src/components/TemplateSummaryModal.tsx
@@ -122,9 +122,10 @@ export default function TemplateSummaryModal(props: ITemplateSummaryModalProps) 
 function ComposableTable(props: { stage: string; curatorJobs?: { name: string; type?: string }[] }) {
   const { curatorJobs, stage } = props
   const { t } = useTranslation()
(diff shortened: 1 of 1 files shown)


frontend/src/components/TemplateSummaryModal.tsx
```diff
@@ -122,9 +122,10 @@ export default function TemplateSummaryModal(props: ITemplateSummaryModalProps)
 function ComposableTable(props: { stage: string; curatorJobs?: { name: string; type?: string }[] }) {
   const { curatorJobs, stage } = props
   const { t } = useTranslation()
+  // let theme = localStorage.getItem('theme')
```
KevinFCormier, line 125: Remove commented line

KevinFCormier (commented):
I'm not sure about this change, because we have other contexts where the table background will be darker, like in the modals:

image


Another option would be to remove the hard-coded padding values, and then it looks better with the dark table.
image

image

PR #2715, opened 2023-03-29
CIM: Move InfraEnv proxy configuration to details page

rawagner:
Requires https://github.com/openshift-assisted/assisted-ui-lib/pull/1994


rawagner:

PR #2717, opened 2023-03-29
ACM-4185 Show searchbar suggestions when operator is used in search query

zlayne:


frontend/src/routes/Home/Search/SearchPage.tsx
```diff
@@ -152,31 +151,40 @@ function RenderSearchBar(props: {
...
-          _.get(searchSchemaResults, 'data.searchSchema.allProperties', []),
+          _.get(searchSchemaData, 'searchSchema.allProperties', [
+            'name',
+            'namespace',
+            'label',
+            'kind',
+            'cluster',
+            'apigroup',
+            'created',
+            'Loading more...',
```
KevinFCormier, line 173: Looks like "Loading more..." was previously translated and now is not?

KevinFCormier: Or rather "Loading..." was previously.

zlayne: Added the `Loading more...` string to translations.

PR #2723, opened 2023-03-30
Translation update to address ACM-3016 and ACM-3018

rh-tchuang:
Updated Japanese, Korean and Simplified Chinese to address ACM-3016 and ACM-3018

Signed-off-by: Terry Chuang <tchuang@redhat.com>


RobDolares (approved):
/lgtm

PR #2740, opened 2023-04-04
CIM: Enable setting installation disk in scaleup modal

rawagner:
Requires https://github.com/openshift-assisted/assisted-ui-lib/pull/2006


rawagner:
included  in https://github.com/stolostron/console/pull/2778

PR #2743, opened 2023-04-04
Add CatalogSource selection during Submariner deployment in UI

aswinsuryan:


skitt:
@MaxBab, @vthapar, does this include all the configuration fields we need?

@aswinsuryan, AFAIK other components use a selection list for some of these; I’m not sure whether that’s appropriate here, but unless there’s a specific reason not to, we should use the same UX.

aswinsuryan:
> @MaxBab, @vthapar, does this include all the configuration fields we need?
> 
> @aswinsuryan, AFAIK other components use a selection list for some of these; I’m not sure whether that’s appropriate here, but unless there’s a specific reason not to, we should use the same UX.

@skitt was it in ACM UI or Openshift UI, I did a quick search in console repo but did not find a UI where we can configure these.

frontend/public/locales/en/translation.json
```diff
@@ -1223,6 +1223,7 @@
   "Enable policy": "Enable policy",
   "Enable search to display application statuses properly.": "Enable search to display application statuses properly.",
   "Enable this option to instruct Submariner not to access any external servers for public IP resolution.": "Enable this option to instruct Submariner not to access any external servers for public IP resolution.",
+  "Enable this option to use Custom Submariner Subscription.": "Enable this option to use Custom Submariner Subscription.",
```
KevinFCormier, line 1226:
```suggestion
  "Enable this option to use a custom Submariner subscription.": "Enable this option to use a custom Submariner subscription.",
```
Unless "Custom Submariner Subscription" is the name of a product or feature that is capitalized like this elsewhere, I would use lowercase mostly.

frontend/public/locales/en/translation.json
```diff
@@ -1252,6 +1254,10 @@
   "Enter the application set name": "Enter the application set name",
   "Enter the base DNS domain": "Enter the base DNS domain",
   "Enter the basic credentials information": "Enter the basic credentials information",
+  "Enter the catalog source": "Enter the catalog source",
+  "Enter the catalog source of a submariner subscription, if left blank the default values will be used.": "Enter the catalog source of a submariner subscription, if left blank the default values will be used.",
```
KevinFCormier, line 1258:
```suggestion
  "Enter the catalog source of a Submariner subscription. If left blank, the default values will be used.": "Enter the catalog source of a Submariner subscription. If left blank, the default values will be used.",
```
This is a comma-splice, so I suggest changing it to 2 sentences. And I would think Submariner should always be capitalized?

frontend/public/locales/en/translation.json
```diff
@@ -2843,6 +2855,7 @@
   "Use a Git repository": "Use a Git repository",
   "Use a Helm repository": "Use a Helm repository",
   "Use cluster labels and application placement rules to easily move workloads across clusters, even between multiple cloud providers.": "Use cluster labels and application placement rules to easily move workloads across clusters, even between multiple cloud providers.",
+  "Use Custom Submariner Subscription": "Use Custom Submariner Subscription",
```
KevinFCormier, line 2858: Same casing concerns here.

KevinFCormier (commented):
The code looks good, but I noted some issues with the UI strings. Perhaps @oafischer could review them as well. Thanks!

frontend/public/locales/en/translation.json
```diff
@@ -1518,9 +1528,11 @@
   "infraEnvCount_plural": "{{count}} infrastructure environments",
   "Infrastructure": "Infrastructure",
   "install": "Install",
+  "Install Plan Approval": "Install Plan Approval",
   "Install plan approved": "Install plan approved",
   "Install the operator": "Install the operator",
   "installing": "Installing",
+  "InstallPlanApproval determines whether subscription installation plans are applied automatically.": "InstallPlanApproval determines whether subscription installation plans are applied automatically.",
```
oafischer, line 1535:
```suggestion
  "InstallPlanApproval determines if subscription installation plans are applied automatically.": "InstallPlanApproval determines if subscription installation plans are applied automatically.",
```

frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetDetails.test.tsx
```diff
@@ -1065,6 +1071,7 @@ describe('ClusterSetDetails page', () => {
     await typeByTestId('awsAccessKeyID', mockManagedClusterNoCredentialsSecret.data!.aws_access_key_id)
     await typeByTestId('awsSecretAccessKeyID', mockManagedClusterNoCredentialsSecret.data!.aws_secret_access_key)
     await typeByTestId('global-net-cidr', mockManagedClusterNoCredentialsSubmarinerConfig.spec?.globalCIDR!)
+    await clickByLabel('Use Custom Submariner Subscription')
```
oafischer, line 1074: Check here for cases as well. Make sure it is consistent.

frontend/public/locales/en/translation.json
aswinsuryan: Changed to lower case.

frontend/public/locales/en/translation.json
aswinsuryan: Done.

frontend/src/routes/Infrastructure/Clusters/ClusterSets/ClusterSetDetails/ClusterSetDetails.test.tsx
aswinsuryan: Changed to lower case expect for the starting word.

aswinsuryan:
> > @skitt was it in ACM UI or Openshift UI, I did a quick search in console repo but did not find a UI where we can configure these.
> 
> In the main OpenShift UI, you can choose the update channel; I think the VolSync addon has something similar. @MaxBab had something specific in mind.

Will wait for @MaxBab to get his thoughts.

MaxBab:
@aswinsuryan 

Two questions, please.
1) This configuration block exists on a per each cluster section. Right?
2) Could we add a note for a user that `startingCSV` is an optional parameter?
Better to not use it to fetch the latest version of the channel.

aswinsuryan:
> * sts on a per each cluster section. Right?
Yes this is a per cluster configuration
> * Could we add a note for a user that `s`
I will add that.

@MaxBab Do you think some of these values should be pre-populated? Since this is for using custom versions I thought of having text boxes which allow filling any values should be useful than drop downs with pre-populated values.

MaxBab:
> > * sts on a per each cluster section. Right?
> >   Yes this is a per cluster configuration
> > * Could we add a note for a user that `s`
> >   I will add that.
> 
> @MaxBab Do you think some of these values should be pre-populated? Since this is for using custom versions I thought of having text boxes which allow filling any values should be useful than drop downs with pre-populated values.

@aswinsuryan 
This section is optional and by default will not be used during the deployment.
So, I don't think it should be pre-populated with values.
Here we agree.

I'm just saying, that if the user will choose to provide custom values, maybe it worth mention, near the name of the parameter, that it's optional.
So the user will know that he don't need to fill up this parameter if it's not really required for him.

aswinsuryan:
image

@MaxBab Added isRequired for other fields(*) expect for startingCSV

MaxBab (approved):
Great.
Thanks for the change.

aswinsuryan:
@KevinFCormier I have addressed the issuw with UI string and this PR is good to go.

frontend/public/locales/en/translation.json
```diff
@@ -1252,6 +1254,10 @@
   "Enter the application set name": "Enter the application set name",
   "Enter the base DNS domain": "Enter the base DNS domain",
   "Enter the basic credentials information": "Enter the basic credentials information",
+  "Enter the catalog source": "Enter the catalog source",
+  "Enter the catalog source of a Submariner subscription. If left blank, the default values will be used.": "Enter the catalog source of a Submariner subscription. If left blank, the default values will be used.",
+  "Enter the channel": "Enter the channel",
+  "Enter the channel of a submariner subscription.": "Enter the channel of a submariner subscription.",
```
KevinFCormier, line 1260: @aswinsuryan Still using lowercase for `submariner` here.

frontend/public/locales/en/translation.json
```diff
@@ -1269,6 +1275,10 @@
   "Enter the OpenStack cloud name to reference in the clouds.yaml": "Enter the OpenStack cloud name to reference in the clouds.yaml",
   "Enter the OpenStack credentials": "Enter the OpenStack credentials",
   "Enter the pull secret and SSH keys": "Enter the pull secret and SSH keys",
+  "Enter the SourceNamespace": "Enter the SourceNamespace",
+  "Enter the SourceNamespace  of the catalog source namespace of a submariner subscription. if left blank the default values will be used.": "Enter the SourceNamespace  of the catalog source namespace of a submariner subscription. if left blank the default values will be used.",
+  "Enter the startingCSV": "Enter the startingCSV",
+  "Enter the startingCSV of a submariner subscription": "Enter the startingCSV of a submariner subscription",
```
KevinFCormier, line 1281: And here.

frontend/public/locales/en/translation.json
```diff
@@ -2843,6 +2855,7 @@
   "Use a Git repository": "Use a Git repository",
   "Use a Helm repository": "Use a Helm repository",
   "Use cluster labels and application placement rules to easily move workloads across clusters, even between multiple cloud providers.": "Use cluster labels and application placement rules to easily move workloads across clusters, even between multiple cloud providers.",
+  "Use custom submariner subscription": "Use custom submariner subscription",
```
KevinFCormier, line 2858: And here.

aswinsuryan:
@KevinFCormier sorry missed it, I updated the remaining.

PR #2747, opened 2023-04-05
ACM-4616: Ansible log faliure fix

Randy424:
regarding: https://issues.redhat.com/browse/ACM-4616

### DIFF
--- a/frontend/src/resources/pod.ts
+++ b/frontend/src/resources/pod.ts
@@ -67,8 +67,8 @@ export async function getDeprovisionPod(namespace: string, name: string) {
   })
 }
 
-export async function getLatestHivePod(namespace: string) {
-  const response = listPods(namespace, [])
+export async function getLatestHivePod(namespace: string, name: string) {
+  const response = listPods(namespace, [getClusterDeploymentNameSelector(name)])
   return await response.promise.then((result) => {
     const latestJob = getLatest<Pod>(result, 'metadata.creationTimestamp')
     return latestJob
@@ -81,7 +81,7 @@ export async function getHivePod(namespace: string, name: string, status: string
   if (status === ClusterStatus.creating) {
     hiveJob = await getProvisionPod(namespace, name)
   } else if (status === ClusterStatus.provisionfailed || status === ClusterStatus.deprovisionfailed) {
-    hiveJob = await getLatestHivePod(namespace)
+    hiveJob = await getLatestHivePod(namespace, name)
   } else if (status === ClusterStatus.destroying) {
     hiveJob = await getDeprovisionPod(namespace, name)
   }
--- a/frontend/src/resources/pod.ts
+++ b/frontend/src/resources/pod.ts
@@ -67,7 +67,7 @@ export async function getDeprovisionPod(namespace: string, name: string) {
   })
 }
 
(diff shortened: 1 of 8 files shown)


Randy424:
Preparing live cluster testing ahead of review.

frontend/src/resources/pod.ts
```diff
@@ -34,7 +34,7 @@ export interface PodList extends IResource {
   items: Pod[]
 }
 
-export function listHivePods(namespace: string, labels?: string[]) {
+export function listPods(namespace: string, labels?: string[]) {
```
Randy424, line 37: This isn't a necessary change, but when using this function myself it seemed more appropriate to state exactly what the function was doing in it's name, not how it was being implemented elsewhere.

frontend/src/resources/pod.ts
```diff
@@ -80,10 +80,22 @@ export async function getHivePod(namespace: string, name: string, status: string
   /* istanbul ignore else */
   if (status === ClusterStatus.creating) {
     hiveJob = await getProvisionPod(namespace, name)
-  } else if (status === ClusterStatus.provisionfailed || status === ClusterStatus.deprovisionfailed) {
+  } else if (
+    status === ClusterStatus.provisionfailed ||
+    status === ClusterStatus.deprovisionfailed ||
+    status === ClusterStatus.posthookjob ||
+    status === ClusterStatus.posthookfailed
```
Randy424, line 87: I noticed that when posthook runs or fails we can't access the provision logs from the details page because these cases weren't here. The link component still renders however, it just goes nowhere. It made more sense to me to add these cases.

Randy424:
Update: Added code to also cover StatusField component.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ClusterDetails/ClusterDetails.test.tsx
```diff
@@ -1314,7 +1314,7 @@ describe('ClusterDetails', () => {
   test('overview page opens logs', async () => {
     const nocks: Scope[] = [nockListHiveProvisionJobs()]
     window.open = jest.fn()
-    await clickByText('View logs', 1)
+    await clickByText('View logs', 0)
```
Randy424, line 1317: Change in behavior in the log button requires change in this test, namely, when there are posthook jobs defined but no prehook jobs defined, we no longer render the prehook link. We are still rendering the documentation link when no hooks are defined as well.

frontend/src/ui-components/AcmProgressTracker/AcmProgressTracker.tsx
```diff
@@ -112,6 +113,7 @@ export function AcmProgressTracker(props: AcmProgressTrackerProps) {
               </TextContent>
               {step.link && (
                 <AcmButton
+                  id={step.stepID && `${step.stepID}-link`}
```
Randy424, line 116: Added an id to the link button to help with testing, as multiple view logs can be rendered at a time.

KevinFCormier (commented):
I am still reviewing, but I have left one inline comment so far. 

I am also wondering - with the way `linkUrl` and `linkCallback` are implemented, is it possible we will get 2 new tabs opened in some cases?

PR #2750, opened 2023-04-06
Re-enable prettier in check script

KevinFCormier:
- Within frontend and backend directories, run checks sequentially for easier reading of output and to make sure commands don't interfere with each other
- Add prettier back to the list of checks to catch files that are not linted


Randy424 (approved):
/lgtm

PR #2758, opened 2023-04-11
ACM-4784 Cluster details status should not use search for counts

zlayne:


chenz4027:
Overall looks good to me. I like the fact that we no longer rely on search to get the data, thanks Zack!

I'll try run this PR locally against a hub with multiple clusters just to verify tomorrow.

chenz4027:
Tried QE's env with multiple apps, seems like the number is a bit off.

image
image
image
image

image
image

Will ping @zlayne you the env

zlayne:
@chenz4027 I am still looking into the application count difference. But the policy violation counts now match the Governance overview table seen below. i think the counts we incorrect previously.
Screen Shot 2023-04-12 at 10 18 45 AM

zlayne:
Hey @chenz4027 I have added some logic here: https://github.com/stolostron/console/pull/2758/files#diff-e96d70676c0c78b80154b9d66d3d38dd7806707f8ef8de9f2ec4e4964f3f8766R140 to handle the AppSet count. Let me know what you think.

chenz4027:
@zlayne It works and looks great for managed cluster overview. Although seems like the general overview is still a bit off?
image
image

image
image

zlayne:
Hey @KevinFCormier, could we get your review here when you have a moment? Maggie and I have both committed and we would like another set of eyes.

PR #2789, opened 2023-04-19
ACM-5035 Fix mis-match hooks crash

fxiang1:
https://issues.redhat.com/browse/ACM-5035

- Revert the change to fix the crash. It's not obvious where the hooks mis-match is coming from.


chenz4027 (approved):
/lgtm

PR #2801, opened 2023-04-24
ACM-5134

chenz4027:


fxiang1 (approved):
Look great! Thanks Maggie!

frontend/public/locales/en/translation.json
```diff
@@ -734,6 +734,7 @@
   "clusterSetBinding.edit.select.placeholder": "Select namespaces",
   "clusterSetBinding.edit.title": "Edit namespace bindings",
   "clusterSets": "Cluster sets",
+  "ClusterSets failed to fetch. Check the Gitopscluster and Placement YAML for status errors.": "ClusterSets failed to fetch. Check the Gitopscluster and Placement YAML for status errors.",
```
KevinFCormier, line 737: @chenz4027 Did you have someone from doc review this string and is`Gitopscluster` the correct casing?

chenz4027: Thanks Kevin, nice catch. It should be GitOpsCluster instead as that's the correct kind for the resource.

chenz4027: Will open a separate PR for it

PR #2805, opened 2023-04-25
ACM-5063 Upgraded-clusters-appear-to-be-automation

jeswanke:


KevinFCormier (changes requested):
I have several concerns with this PR.
- You have filtered the `ClusterCurator` where it may have unintended impacts, like where it is passed to the `<DistributionField>` component or used in `get-cluster.ts`. Whether there are Ansible pre-/post- install/upgrade actions is only relevant for whether we display the link to the Automation template and whether we allow removal of the template.
- You have repeated the same filtering logic in FOUR places. This is very bad for maintainability. Perhaps you could build on the function here, or add more helper functions in this file: https://github.com/stolostron/console/blob/main/frontend/src/resources/cluster-curator.ts#L108
- The logic you have used in your filter is NOT correct and does not address the issue. See screenshot.

image

It would also be great if the individual row action `Remove automation template` does not appear when there is no automation template. Currently that action uses the same test for applicability as the `Update automation template` action, but removal could have the additional requirement that it does have a template set currently.

jeswanke:
all issues fixed

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -437,9 +436,20 @@ export function ClustersTable(props: {
...
+        title: t('Remove automation templates'),
+        click: (managedClusters: Array<Cluster>) => {
+          if (!managedClusters) return
+          setRemoveAutomationTemplates(managedClusters)
+        },
+        variant: 'bulk-action',
+      })
+    }
+    return actions
+  }, [modalColumns, hasAllAutomationTemplates, t])
```
KevinFCormier, line 452: @jeswanke This is not what I was asking for. Currently we do not filter bulk actions. This is so that you can select all and easily apply an action to all clusters that it is applicable for.

I was suggesting that you remove the action on the table row kebab only. This is what I meant by "individual row action"

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/ClusterActionDropdown.tsx
```diff
@@ -114,16 +114,20 @@ export function ClusterActionDropdown(props: { cluster: Cluster; isKebab: boolea
...
+                text: t('Remove automation template'),
+                click: () => setShowRemoveAutomationModal(true),
+                isAriaDisabled: true,
+                rbac: [
+                  rbacDelete(ClusterCuratorDefinition, cluster.namespace),
+                  rbacDelete(SecretDefinition, cluster.namespace),
+                ],
+              },
+            ]
+          : []),
```
KevinFCormier, line 130: This array is already filtered by the `clusterSupportsAction` function. If you use it, the logic can be centralized and reused where needed.

frontend/src/resources/utils/get-cluster.ts
```diff
@@ -507,6 +509,10 @@ export function getCluster(
   }
 }
 
+export function hasAutomationTemplates(clusterCurator?: ClusterCurator) {
+  return !!clusterCurator
```
KevinFCormier, line 513: Logic is still not correct.

KevinFCormier (changes requested):

PR #2819, opened 2023-04-27
Order policy history from the order in the status

JustinKuli:
The event timestamps are only precise to the second. The history in the policy status is specially ordered by the framework using additional information. This change makes the UI just match the ordering in the policy status.

Refs:
 - https://issues.redhat.com/browse/ACM-5149

### DIFF
--- a/frontend/src/routes/Governance/policies/policy-details/PolicyDetailsHistory.tsx
+++ b/frontend/src/routes/Governance/policies/policy-details/PolicyDetailsHistory.tsx
@@ -12,6 +12,7 @@ import { AcmEmptyState, AcmTable, AcmTablePaginationContextProvider, compareStri
 interface HistoryTableData {
   message: string
   timestamp: moment.MomentInput
+  index: number
 }
 
 export function PolicyDetailsHistory(props: {
@@ -37,10 +38,12 @@ export function PolicyDetailsHistory(props: {
       .filter((detail) => detail?.templateMeta?.name === templateName)
       .forEach((detail: PolicyStatusDetails) => {
         const history = detail.history ?? []
+        let i = 0
         history.forEach((status) => {
           statuses.push({
             message: status.message ?? '-',
             timestamp: status.lastTimestamp ?? '-',
+            index: i--, // decrement to make the sort work by default
           })
         })
       })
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Governance/policies/policy-details/PolicyDetailsHistory.tsx
```diff
@@ -37,10 +38,12 @@ export function PolicyDetailsHistory(props: {
       .filter((detail) => detail?.templateMeta?.name === templateName)
       .forEach((detail: PolicyStatusDetails) => {
         const history = detail.history ?? []
+        let i = 0
         history.forEach((status) => {
           statuses.push({
             message: status.message ?? '-',
             timestamp: status.lastTimestamp ?? '-',
+            index: i--, // decrement to make the sort work by default
```
yiraeChristineKim, line 46: this going to minus?

JustinKuli: Yeah, because this replaces sorting by the timestamp, "smaller" needs to be "older", and in the policy status, the first item in the history is the most recent... it's definitely a bit odd, and mostly I wish I had a better name for this field, rather than `index`.

I thought about providing a custom sort function on this, basically just reversing it, but that seemed actually more complicated than this workaround.

dhaiducek: Different idea: What if `sort` is removed entirely? Does it leave the list alone?

JustinKuli: I can try removing it, but I'm not sure what will happen in that case if the user clicks on the table header to do the sort (or to switch the sort direction).

dhaiducek: I was wondering that. What happens with the `index` implementation? Does the order change?

JustinKuli: With my current changes, the default ordering on the page matches what is in the policy status yaml. Users can click on the table header and reverse the order. Is that what you mean?

dhaiducek (approved):

yiraeChristineKim (changes requested):

PR #2822, opened 2023-04-28
Update openshift-assisted-ui-lib to monorepo based @openshift-assisted/ui-lib

jtomasek:


frontend/package.json
```diff
@@ -41,6 +41,8 @@
     "@mui/styles": "^5.11.1",
     "@octokit/rest": "19.0.3",
     "@octokit/types": "6.40.0",
+    "@openshift-assisted/locales": "^2.19.4",
+    "@openshift-assisted/ui-lib": "^2.19.4",
```
jtomasek, line 45: This needs to be replaced with 2.8.2-cim

PR #2832, opened 2023-05-03
Stay in PlacementRule mode when switch to Existing Policies

mprahl:
Before this commit, if you had a PlacementBinding that pointed to a PlacementRule when you were in "New placement" mode in the new policy wizard, select "Existing placement" would cause it to reset back to pointing to a Placement.

Relates:
https://issues.redhat.com/browse/ACM-4988


yiraeChristineKim (approved):

KevinFCormier:
/cc @chenz4027 @fxiang1 

I think this code may be shared with ALC, so I'm requesting review from Maggie & Feng.

fxiang1:
/lgtm
We are using Placements instead of PlacementRules in that part of the wizard so not sure how much that would affect us.

chenz4027:
/lgtm
Thanks Matt and Kevin!

dhaiducek (approved):

PR #2835, opened 2023-05-04
ACM-5281 Use serviceNetwork instead of serviceCIDR for HostedCluster

fxiang1:
https://issues.redhat.com/browse/ACM-5281

- Use serviceNetwork instead of serviceCIDR for HostedCluster as serviceCIDR is removed from version v1beta1

nonadvancednetwork

image


chenz4027 (approved):
/lgtm 
Thanks for the quick fix!

PR #2839, opened 2023-05-09
ACM-3084 - Empty page dark mode png

Randy424:
regarding: [ACM-3084](https://issues.redhat.com/browse/ACM-3084)


Randy424:
@KevinFCormier 
I tested this primarily on Standalone ACM, but also used storybook.
Two problems I'm running into with the SVG however:
1. I'm still checking local storage for dark or light theme, but that method does not give me the ability to re-render as the CSS cascades when hitting the "light mode switcher". This may not be a real problem because this switcher (so far as I know) does not exist in live MCE, but I wanted to point out what the current behavior was. 
2. Apparently storybook has issues with loading SVG as react components. You can read more about it [here.](https://stackoverflow.com/questions/54292667/react-storybook-svg-failed-to-execute-createelement-on-document). Not yet sure how to respond to the issue, but it does break our EmptyState storybook demo.

One other question I had was, what did you have in mind for CSS + SVG? Maybe it could fix our issues here.

Randy424:
/hold 
Waiting for more information from design

frontend/src/ui-components/AcmEmptyState/AcmEmptyState.tsx
```diff
@@ -1,12 +1,18 @@
...
+  EmptyStatePrimary,
+  EmptyStateVariant,
+  EmptyStateIcon,
+  Title,
+} from '@patternfly/react-core'
 import { ReactNode } from 'react'
-import { useTranslation } from '../../lib/acm-i18next'
-import emptyPagePng from '../assets/EmptyPageIcon.png'
-import emptyTablePng from '../assets/EmptyTableIcon.png'
 import Folder from '../assets/Folder.png'
```
KevinFCormier, line 12: Can also get rid of this import and the file.

KevinFCormier (commented):
@Randy424 From this folder:
https://github.com/stolostron/console/tree/2e7a5716b8c4f38854b158538817b6547e0ef0c3/frontend/src/ui-components/assets

I think we can also get rid of `IBM-Cloud.png` and if you update `AcmPageProcess` to use a trash can icon,  you can get rid of `resource-destroyed.png` as well.

frontend/src/ui-components/AcmEmptyState/AcmEmptyState.test.tsx
```diff
@@ -21,22 +21,13 @@ describe('AcmEmptyState', () => {
...
     expect(getByText('Empty state title')).toBeInTheDocument()
     expect(container.querySelector('button')).toBeNull()
     expect(await axe(container)).toHaveNoViolations()
   })
   test('renders with isEmptyTableState', () => {
-    const { getByText } = render(
-      <AcmEmptyState title="Empty state title" message="Empty state message" isEmptyTableState />
-    )
+    const { getByText } = render(<AcmEmptyState title="Empty state title" message="Empty state message" />)
     expect(getByText('Empty state title')).toBeInTheDocument()
```
KevinFCormier, line 31: Can remove this whole test.

frontend/src/ui-components/AcmPageProcess/AcmPageProcess.tsx
```diff
@@ -59,7 +59,7 @@ export function AcmPageProcess(props: AcmPageProccessProps) {
       <AcmPageCard>
         <EmptyState>
           {/* eslint-disable-next-line jsx-a11y/alt-text */}
-          <img src={DestroyedImage} role="presentation" className={classes.image} />
+          <TrashIcon size="xl" />
```
KevinFCormier, line 62:
```suggestion
          <EmptyStateIcon icon={TrashIcon} />
```

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1035,7 +1035,6 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
               <AcmEmptyState
                 title={t('No results found')}
                 message={t('No results match the filter criteria. Clear filters to show results.')}
-                showIcon={false}
```
KevinFCormier, line 1038: Why did we remove this? I believe this is used when you filter the table down to no results, and currently we don't show an icon. If we do show an icon, I think the magnifying glass would be better here.

KevinFCormier (commented):
Please see inline comments.

frontend/src/ui-components/AcmTable/AcmTable.tsx
Randy424: It will default to the cubes icon, but yeah, I agree. Magnifying glass is a better fit...

KevinFCormier (approved):

PR #2846, opened 2023-05-10
ACM-5427 Support both new and old versions of TimePicker

fxiang1:
https://issues.redhat.com/browse/ACM-5427

PF TimePicker is still in beta but looks like we've been using it for some time. The newer version of PF TimePicker that's in OCP 4.13 added an event param to the onChange function which is causing issues for the UI. As a plugin we have no control over which PF version we use.

- Check the param type to see which param to use


KevinFCormier:
/lgtm
Awesome! Glad this was caught.

PR #2853, opened 2023-05-11
ACM-5425 Warn AppSet with multiple sources is not supported

fxiang1:
https://issues.redhat.com/browse/ACM-5425

AppSet/Argo app has a new feature in TechPreview that supports deploying apps with multiple repos. However the app UI doesn't support this and crashes. This fix does the following:

- Fix the crash for AppSet and Argo apps on the table
- Fix the create AppSet wizard from crashing
- Warn users that the multiple sources feature is not supported
- Disable row actions for viewing/editing AppSets and viewing Argo apps
- Disable link to app details page as that crashes as well

This changes will be temporary as we're planning to support the multiple sources feature in 2.9.

image


fxiang1:
Putting on hold for design feedback.

chenz4027 (approved):
/lgtm
Thanks Feng!

PR #2875, opened 2023-05-19
ACM-2666 Added percentage beside hc upgrade status

o-farag:
Previously, the percentage completion when upgrading would not be displayed. Now show upgrade percentage beside the status.
image


o-farag:
@fxiang1 regarding the security hotspot, is it an issue since the message isn't defined by the user, it's defined in the CR? I could try using string manipulation instead if you think that's better

fxiang1:
@o-farag I think it should be ok since the input is not from users but rather from the API status and I don't believe users can manipulate the API status.

fxiang1 (approved):
LGTM! Thanks Omar!

PR #2882, opened 2023-05-25
ACM-5596-Base-domain-is-not-populated-in-the-form

jeswanke:
1  the ssh public key add to the yaml from the credential wizard was in wrong format causing syntax errors
2 when yaml has syntax errors, the form won't update from it as a precaution


Randy424 (approved):
/lgtm

PR #2912, opened 2023-06-12
ACM-2861 disable upgrading nodepool when pending

o-farag:
Disable nodepool upgrade available when nodepool is pending

Before:
image

After:
image


fxiang1 (approved):
LGTM! Thanks Omar!

PR #2916, opened 2023-06-13
Fix double query on run search click

zlayne:


frontend/src/routes/Home/Search/components/Searchbar.tsx
```diff
@@ -198,7 +199,7 @@ export function Searchbar(props: SearchbarProps) {
       )
       .map((currentValue) => (
         <MenuItem
-          isDisabled={currentValue.name.includes('Loading')}
+          isDisabled={currentValue.name.includes('Loading') || currentValue.disabled}
```
KevinFCormier, line 202: Is the pre-existing logic here compatible with translation?

zlayne: @KevinFCormier logic should now account for translation.

frontend/src/routes/Home/Search/search-helper.ts
```diff
@@ -96,6 +96,7 @@ export function formatSearchbarSuggestions(
         id: `id-${field}`,
         name: field,
         kind: suggestionKind,
+        disabled: field.includes(t('Loading')),
```
KevinFCormier, line 99: I don't think you should mix the special "Loading..." or "Loading more..." entries with the data, and then detect them based on translated strings.

How about adding a param to the `formatSearchbarSuggestions` that indicates if suggestions are still being loaded, and then it can add "Loading..." or "Loading more..." to the bottom of the list depending on whether there were 0 suggestions or >0 suggestions in `data`.

PR #2923, opened 2023-06-20
ACM-6035 Change AppSet topology to use placementdecision for placement

fxiang1:
https://issues.redhat.com/browse/ACM-6035

- Change AppSet topology to use placementdecision for placement


chenz4027 (approved):
/lgtm

PR #2930, opened 2023-06-23
ACM-6100 Backport time window fix to 2.7

fxiang1:
https://issues.redhat.com/browse/ACM-6100

- Backport time window fix to 2.7


fxiang1:
We cannot test with two different PF versions at the same time so more then 66% test coverage is not possible.

chenz4027 (approved):
/lgtm
Thanks Feng!

PR #2945, opened 2023-07-06
ACM-6271 Don't create placementrule when selecting an existing one

fxiang1:
https://issues.redhat.com/browse/ACM-6271

- Fix subscription app create wizard to not create a new placementrule when selecting an existing placementrule


chenz4027 (approved):
/lgtm
Thanks Feng for fixing it!

PR #2947, opened 2023-07-12
ACM-6301 Fix topology for pod node when replicaset is great than one

fxiang1:
https://issues.redhat.com/browse/ACM-6301

- Handle pod kind as a special case in computeStatus
- Fix a bug where replicaSet is being used as a string but really should be a number


chenz4027 (approved):
/lgtm
Thanks Feng!

PR #2951, opened 2023-07-14
[WIP] Remove ProxyAgent

KevinFCormier:


KevinFCormier:
Something is not quite right yet. It seems like the protocol of the proxied URL is being changed from `https` to `http` somewhere along the line.
image

KevinFCormier:
Closing; we will fully remove proxy support for Ansible instead, since AAP does not have this support.

PR #2957, opened 2023-07-17
ACM-6170: Throw error if  placement selected

chenz4027:
https://issues.redhat.com/browse/ACM-6170

image
image


fxiang1 (approved):
LGTM! Thanks for the fix Maggie!

PR #2964, opened 2023-07-26
ACM-6656 [release 2.8] Fix table search 

KevinFCormier:
https://issues.redhat.com/browse/ACM-6656

### DIFF
--- a/frontend/src/ui-components/AcmTable/AcmTable.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmTable.tsx
@@ -450,12 +450,15 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
 
   const classes = useStyles()
 
-  const setInternalSearchWithDebounce =
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const setInternalSearchWithDebounce = useCallback(
     process.env.NODE_ENV !== 'test'
       ? debounce((search: string) => {
           setInternalSearch(search)
         }, SEARCH_DEBOUNCE_TIME)
-      : setInternalSearch
+      : setInternalSearch,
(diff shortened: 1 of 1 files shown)


frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -450,12 +450,15 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
 
   const classes = useStyles()
 
-  const setInternalSearchWithDebounce =
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const setInternalSearchWithDebounce = useCallback(
```
KevinFCormier, line 454: @fxiang1 @chenz4027 When I was doing some cleanup in this file for 2.8, I took out the useCallback - I'm not sure why I did this. At the same time, I added cleanup to the next effect so that we are not leaking state updates (the call to `clear()`). When you combine these two changes, I think it is possible that the debounced function (which sets internalSearch) will be cleared before it executes if something is causing the page to render more frequently than every 500 ms. That shouldn't be the case (and if it is, I guess we have another bug) but if so, this would at least fix the table search.

Compare to 2.7: https://github.com/stolostron/console/blob/release-2.7/frontend/src/ui-components/AcmTable/AcmTable.tsx#L467

fxiang1: Sounds good @KevinFCormier! hopefully this will fix the table search issue.

chenz4027: Thanks Kevin!

fxiang1 (approved):

PR #2966, opened 2023-07-30
Submariner: add support for IBM ROKS

yboaron:
Enable Submariner deployment on IBM ROKS (managed Openshift service), similar to other Openshift managed services (ROSA and ARO) submariner-gateway will be deployed using K8S LB service and there is no need to run cloud prepare.


KevinFCormier (approved):

Jaanki:
/assing @yboaron 
/unassign @aswinsuryan 
iunassign @KevinFCormier

PR #2972, opened 2023-08-01
Add column controll in policies page

yiraeChristineKim:
Description: Add a draggable column management https://www.patternfly.org/v4/components/table/react-demos/#column-management. The result should match with https://docs.google.com/document/d/1-UfL1QbnOdoxUf62bHhX4CwbuBKgm8x-mheXuQCxpoM/edit. I added an extra feature "the column selections should be saved in local storage so that Users can see the same result after the browser is refreshed or closed and returned" which our team requested

ref: https://issues.redhat.com/browse/ACM-6563


frontend/src/routes/Governance/policies/PolicyContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+  selectedColIds: defaultColIds,
+}
+
+const PolicyContext = createContext<{ state: policyStateType; dispatch: React.Dispatch<any> }>({
+  state: initialState,
+  dispatch: () => null,
+})
+
+function PolicyProvider(props: any) {
+  const localState = JSON.parse(localStorage.getItem('policies')!)
```
KevinFCormier, line 37: I would recommend a more descriptive name for the local storage item.

KevinFCormier (dismissed):
@yiraeChristineKim This is a great start! I know Ethan was wanting to see this for a long time, so it's great that you had time to tackle it.

Overall, I would prefer to see this implemented in a generic way, so that it is not specific to policies and any table can opt in. It could even be integrated with `AcmTable` as an opt-in feature. In addition to `id`, we could add optional boolean properties like `optional` and `defaultExcluded` to columns to prescribe which columns can be disabled and which are not included by default. The order these are provided to the table also defines the default ordering. This would keep the column metadata all in one place. Currently, you have the required columns in a constants file that is directly referenced by the column management. The table implementation itself could add the button when the user enables it, so that there is less boilerplate and duplication of code required when other tables want to use this feature.

Of course, if you feel that is out of scope, this could be left to future work.

I think there are a few issues with the implementation just for Policies though. Please see my inline comments. In addition:
- I think dragging the columns in the modal is intended to change the order of the columns in the actual table. Currently they are only reordered (with order preserved) in the modal.
- The *Restore columns* button does not restore the original order, only the checkbox state.
- Specifically for policies, I think too many columns are required. I think *Name* definitely has to be present, but I think we should allow the user to disable any other columns.

yiraeChristineKim:
> @yiraeChristineKim This is a great start! I know Ethan was wanting to see this for a long time, so it's great that you had time to tackle it.
> 
> Overall, I would prefer to see this implemented in a generic way, so that it is not specific to policies and any table can opt in. It could even be integrated with `AcmTable` as an opt-in feature. In addition to `id`, we could add optional boolean properties like `optional` and `defaultExcluded` to columns to prescribe which columns can be disabled and which are not included by default. The order these are provided to the table also defines the default ordering. This would keep the column metadata all in one place. Currently, you have the required columns in a constants file that is directly referenced by the column management. The table implementation itself could add the button when the user enables it, so that there is less boilerplate and duplication of code required when other tables want to use this feature.
> 
> Of course, if you feel that is out of scope, this could be left to future work.
> 
> I think there are a few issues with the implementation just for Policies though. Please see my inline comments. In addition:
> 
> * I think dragging the columns in the modal is intended to change the order of the columns in the actual table. Currently they are only reordered (with order preserved) in the modal.
> * The _Restore columns_ button does not restore the original order, only the checkbox state.
> * Specifically for policies, I think too many columns are required. I think _Name_ definitely has to be present, but I think we should allow the user to disable any other columns.

I also thought about that too, column management is only a requested GRC page so temporarily it would be better smaller scope. In the future, if we decide to expand this feature, we can change it to an opt-in component. That is not hard. Also, regarding `optional` and `defaultExcluded`, At first I added in `IAcmTableColumn` but as I mentioned, it is only used in grc now. After I discuss with Ethan and other front associates, Can I create a new ticket for this?

KevinFCormier:
> Can I create a new ticket for this?

@yiraeChristineKim Actually there is already a longstanding epic for this:
https://issues.redhat.com/browse/ACM-5687

yiraeChristineKim:
Let me move to the opt-in scope!

yiraeChristineKim:
@KevinFCormier  About #Specifically for policies, I think too many columns are required. I think Name definitely has to be present, but I think we should allow the user to disable any other columns.
https://docs.google.com/document/d/1-UfL1QbnOdoxUf62bHhX4CwbuBKgm8x-mheXuQCxpoM/edit present default column so disabling other columns need to be discussed

KevinFCormier:
> @KevinFCormier  About **Specifically for policies, I think too many columns are required. I think Name definitely has to be present, but I think we should allow the user to disable any other columns.**
https://docs.google.com/document/d/1-UfL1QbnOdoxUf62bHhX4CwbuBKgm8x-mheXuQCxpoM/edit presents default column so disabling other columns need to be discussed

From the design doc, it says:
> The default setting is as follows. “Policy name,” “Namespace,” “Cluster violation,” “Source,” and “Automation” are checked & disabled. 

But in your code, you have also disabled hiding of the *Created* column, which as shown in the design doc is shown by default but can be removed.

I think it's worth reviewing why each of these columns can't be hidden with @heyethankim 

I think **Name** and **Namespace** are fair, since these guarantee uniqueness.
**Cluster violation** also makes sense, since this is the key functionality of policies.
But why not allow customers to hide the **Source** or **Automation** columns? I guess the argument might be that users should be aware if polices are externally managed or if automation will run, but what if a customer is simply not using either of these features and would prefer to preserve screen real eastate?

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
```diff
@@ -0,0 +1,213 @@
...
+          <DataList aria-label="Table column management" id="table-column-management" isCompact>
+            {items.map((policy) => (
+              <Draggable key={policy.id} hasNoWrapper>
+                <DataListItem aria-labelledby={`table-column-management-${policy.id}`} id={policy.id}>
+                  <DataListItemRow>
+                    <DataListControl>
+                      <DataListDragButton
+                        aria-label="Reorder"
+                        aria-labelledby={`table-column-management-${policy.id}`}
+                        aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
```
fxiang1, line 179: I would suggest we translate this to support accessibility in other languages.

fxiang1 (changes requested):

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
yiraeChristineKim: Good catch Thanks!

PR #2979, opened 2023-08-03
Remove old annotations from hc import

o-farag:
Remove old annotations from hosted cluster when importing
Old: image

Should be: 
image


fxiang1 (approved):
LGTM! Thanks Omar!

PR #2999, opened 2023-08-24
ACM-7063 Overview page should detect if Observability is installed

zlayne:
Related issue: https://issues.redhat.com/browse/ACM-7063

Overview page should detect if Observability is installed and either hide Alert & operator metrics if disabled show if enabled.


backend/src/routes/observabilityProxy.ts
```diff
@@ -0,0 +1,98 @@
...
+  const url = req.url
+
+  const headers: OutgoingHttpHeaders = { authorization: `Bearer ${token}` }
+  for (const header of proxyHeaders) {
+    if (req.headers[header]) headers[header] = req.headers[header]
+  }
+
+  const observabilityProxyRoute = await jsonRequest(
+    process.env.CLUSTER_API_URL +
+      `/apis/route.openshift.io/v1/namespaces/open-cluster-management-observability/routes/rbac-query-proxy`,
```
KevinFCormier, line 61: Could the customer use a different namespace instead of `open-cluster-management-observability`? The documentation refers specifically to that namespace repeatedly, but I did not find any text that specifically says that namespace MUST be used. Is there an alternate way to find this route?

https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/observability/observing-environments-intro#enabling-observability-service

backend/src/app.ts
```diff
@@ -34,6 +35,11 @@ export const router = Router<Router.HTTPVersion.V2>({ maxParamLength: 500 })
 router.get('/readinessProbe', readiness)
 router.get('/livenessProbe', liveness)
 router.get('/ping', respondOK)
+router.all('/api/v1/query', observabilityProxy)
+router.all('/api/v1/query_range', observabilityProxy)
+router.get('/api/v1/targets', observabilityProxy)
+router.get('/api/v1/rules', observabilityProxy)
+router.get('/api/v1/label', observabilityProxy)
```
KevinFCormier, line 42: `/api` is used for k8s APIs... these don't seem to clash, but I wonder if it would be better to use a different route and rewrite in the proxy?

frontend/src/routes/Home/Overview/OverviewPageBeta.tsx
```diff
@@ -99,10 +101,58 @@ export default function OverviewPageBeta(props: { selectedClusterLabels: Record<
...
+      .promise.then((route: any) => {
+        if (route?.spec?.host) {
+          setGrafanaRoute(`${route?.spec?.tls?.termination ? 'https' : 'http'}://${route.spec.host}`)
+        }
+      })
+      .catch((err) => {
+        console.error('Error getting Observability Route: ', err)
+      })
+  }, [])
+
```
KevinFCormier, line 155: RBAC concerns here - not all users will be able to see these resources.

backend/src/routes/observabilityProxy.ts
zlayne: I was going based on the documentation you posted above. I guess they could use another namespace? The only way I could think of finding this Route otherwise would be to query for every Route and then filter them based on ownerReferences (if the owner is `kind: MultiClusterObservability & name: observability`)

backend/src/app.ts
zlayne: Reading through this documentation, I *think* the router matches static routes first and places lowest priority on the wildcard routes, so it shouldn't have clashes.

zlayne: https://www.npmjs.com/package/find-my-way#match-order

KevinFCormier: What I meant by "clashes" is that these routes take precedence over the wildcard, so this would be an issue if `query` or `rules`, for example, were k8s APIs that we want to access. We are lucky at the moment that there is no overlap, but in general I think it's bad practice to have 2 different backends serving stuff under `/api/v1`.

frontend/src/routes/Home/Overview/OverviewPageBeta.tsx
zlayne: Good point - Looks like Clusters page gets the grafana link via the `observability-controller` ClusterManagementAddOn resource. I can switch this to do the same. Should also make the page load faster by using recoil state.

backend/src/app.ts
zlayne: Gotcha, I can update this to use an `/observability` route instead of the current `api/v1/...`

backend/src/routes/observabilityProxy.ts
zlayne: FYI an update before the weekend.

I have tried to install Observability with a different namespace and am finding the Observability CRD never reaches healthy status as it cannot find the object storage secret (assuming due to it not being in the `open-cluster-management-obsevability` ns). 
Also, going through the obs repo, they have `open-cluster-management-observability` hardcoded all throughout the repo (see here: https://github.com/search?q=repo%3Astolostron%2Fmulticluster-observability-operator+open-cluster-management-observability&type=code).

I will ask on Monday if the namespace must be `open-cluster-management-observability` but as far as I can see it is.

zlayne:
@jlpadilla FYI if you would like to review - Changes to detect if Observability is installed, and if enabled, show the Alerts & Operator metrics.

frontend/src/routes/Home/Overview/OverviewPageBeta.tsx
KevinFCormier: Even via recoil, the resources are filtered based on the user's permissions, but I guess we have this issue anywhere we show the Grafana link currently.

PR #3002, opened 2023-08-24
Bumps @openshift-assisted/ui-lib to 2.9.5-cim

jkilzi:
This change removes the need to add @openshift-assisted/locales as a peer-dependency to the project. Now the ui-lib includes the locales as direct dependencies and a types-only package exposing shared models with the Assisted Installer back-end.


KevinFCormier:
Obsolete

PR #3004, opened 2023-08-25
Add prompt message when remediationAction is set to informOnly

zyjjay:
Modified 'getPolicyRemediation' to account for 'informOnly' 
Modified Policy Wizard to offer 'informOnly' as an option in the policy templates section

ref: https://issues.redhat.com/browse/ACM-6879


mprahl:
@yiraeChristineKim could you please review this PR since you're currently working in the same area of code?

frontend/src/wizards/Governance/Policy/PolicyWizard.tsx
```diff
@@ -488,6 +488,12 @@ export function PolicyWizardTemplates(props: { policies: IResource[] }) {
                 'Automatically runs remediation action that is defined in the source, if this feature is supported.'
               )}
             />
+            <Radio
+              id="informOnly"
+              label={t('InformOnly')}
+              value="informOnly"
+              description={t("Overrides parent policy's remediation action to 'Inform'.")}
```
mprahl, line 495:
```suggestion
              description={t("Reports the violation, which requires manual remediation. This cannot be overridden by the parent policy's remediation action.")}
```

frontend/src/wizards/Governance/Policy/PolicyWizard.tsx
```diff
@@ -488,6 +488,12 @@ export function PolicyWizardTemplates(props: { policies: IResource[] }) {
                 'Automatically runs remediation action that is defined in the source, if this feature is supported.'
               )}
             />
+            <Radio
+              id="informOnly"
+              label={t('InformOnly')}
```
mprahl, line 493:
```suggestion
              label={t('Inform Only')}
```

frontend/src/routes/Governance/common/util.tsx
```diff
@@ -467,23 +468,45 @@ export function getPolicyRemediation(policy: Policy | undefined) {
...
+      if (remediationSet.size > 1) {
+        return policy.spec.remediationAction.concat('/informOnly')
+      } else {
+        return 'informOnly'
+      }
+    } else {
+      return policy.spec.remediationAction
+    }
+  }
   return remediationAggregation
```
yiraeChristineKim, line 510: It could create "enforce/inform" or "informOnly/enforce" so I think we need to sort it

zyjjay: In lines 485-493, the condition statements should always make it so that the order is always `inform` -> `enforce` -> `informOnly`. I don't think `enforce` will ever precede `inform`, and similar with the other examples. I'm constructing a string from the set, and not using the set directly, so I don't think we need to sort it.

frontend/src/routes/Governance/common/util.tsx
```diff
@@ -561,3 +584,18 @@ export function formatDescriptionForDropdown(desc: string) {
...
+  const informOnlyPolicies = policyList.filter((policy) => {
+    return getPolicyRemediation(policy).includes('informOnly')
+  })
+
+  let policyNames = ''
+  for (const policy of informOnlyPolicies) {
+    policyNames += policy.metadata.name + ' '
+  }
+
+  return policyNames.trim().replace(' ', ', ')
```
yiraeChristineKim, line 600:
```suggestion
return informOnlyPolicies.reduce((result, policy) => result +', ' + policy.metadata.name, '')

```

frontend/src/routes/Governance/policies/Policies.tsx
```diff
@@ -559,7 +564,11 @@ export default function PoliciesPage() {
                 processing: t('policy.table.actions.enforcing'),
                 items: [...item],
                 emptyState: undefined, // there is always 1 item supplied
-                description: t('policy.modal.message.enforce'),
+                description:
+                  getInformOnlyPolicies(item) == ''
```
yiraeChristineKim, line 568: If you only use this function for these, isn't it better to change the function to return boolean?  `hasInformOnly` something like this

zyjjay: Oh yea, originally I was going to list out the names of the policies, but now the functionality is different so I will change that.

frontend/src/routes/Governance/common/util.test.tsx
```diff
@@ -34,3 +34,94 @@ describe('Test resolveExternalStatus', () => {
     expect(resolveExternalStatus(policy)).toEqual(expected)
   })
 })
+
+describe('Test getPolicyRemediation', () => {
+  test('getPolicyRemediation should be enforce when all template-remediations = enforce', () => {
+    const rootPolicy: Policy = {
```
yiraeChristineKim, line 40: Can you also one more test? root remediation: inform, policy template: enforce, informOnly

yiraeChristineKim:
How about filter, user can filter informOnly?

zyjjay:
> How about filter, user can filter informOnly?

Yes I made those changes here: https://github.com/stolostron/console/pull/3004/files#diff-91793b482c57f4aa67c4bc721e4790d4f39e60ec728e30172f06193f14ca9ae5R720

yiraeChristineKim:
I need a @mprahl  to approve. we need 2 reviewers

PR #3008, opened 2023-08-28
set create-via annotation to hypershift for HC import

rokej:
This is for story https://issues.redhat.com/browse/ACM-6547 to set `open-cluster-management/created-via: hypershift` annotation in ManagedCluster CRs the console creates to auto-import hosted clusters.


chenz4027 (approved):
/lgtm
Thanks Roke!

PR #3000, opened 2023-08-24
ACM-6066: Fix clusterset render issue and add alert

chenz4027:
https://issues.redhat.com/browse/ACM-6066

PR #3010, opened 2023-08-29
Fix clusterset render issue and add alert (#3000)

chenz4027:
https://issues.redhat.com/browse/ACM-6003


fxiang1 (approved):
Thanks Maggie!

PR #3014, opened 2023-08-31
Expose console page visit metric

zlayne:
PR will expose a metric: `acm_console_page_count`.
Metric will be used to track page visits across the RHACM UI. 
This PR will introduce the metric for pages: Overview, Overview beta, Search & Search details. 
To use the metric in other pages - the page will need to include the `usePageVisitMetricHandler` hook with a string param of the page name (which should be defined in the Page enum).

### DIFF
--- a/backend/src/routes/metrics.ts
+++ b/backend/src/routes/metrics.ts
@@ -19,6 +19,7 @@ export async function metrics(req: Http2ServerRequest, res: Http2ServerResponse)
     // POSTs originate from an ACM page - get the page from req url params and increase that pages count
     const page = req.url.split('?')[1]
     acm_console_page_count.labels({ page }).inc()
+    res.end(`Increased ${page} label count for metric acm_console_page_count`)
   } else {
     res.setHeader('Content-Type', register.contentType)
     await register.metrics().then((data) => res.end(data))
--- a/backend/src/routes/metrics.ts
+++ b/backend/src/routes/metrics.ts
@@ -2,10 +2,8 @@
 
 import { Http2ServerRequest, Http2ServerResponse } from 'http2'
 import Prometheus, { register } from 'prom-client'
-
-interface PageData {
-  page: string
-}
+import { catchInternalServerError } from '../lib/respond'
+import { getAuthenticatedToken } from '../lib/token'
 
 const acm_console_page_count = new Prometheus.Counter({
   name: 'acm_console_page_count',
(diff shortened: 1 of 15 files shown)


backend/src/routes/metrics.ts
```diff
@@ -0,0 +1,48 @@
...
+import Prometheus, { register } from 'prom-client'
+import { logger } from '../lib/logger'
+
+interface PageData {
+  page: string
+}
+
+// register.setDefaultLabels({
+//   // label to identify customer?
+// })
```
jlpadilla, line 13: Data should be anonymous. We'll know the cluster reporting this metric and can use it to find the customer account.

backend/src/routes/metrics.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const acm_console_page_visit_counter = new Prometheus.Counter({
+  name: 'acm_console_page_visit_counter',
```
jlpadilla, line 16: Change "counter" to "count" to be consistent with other ACM metrics.
```suggestion
  name: 'acm_console_page_count',
```

KevinFCormier (commented):
This looks very interesting @zlayne. With the current PR, the data is not being sent anywhere, right? And will it matter that we have 4 pods (2 MCE + 2 ACM) or is everything just additive?

I think it would be great if this was built into the `<AcmPage>` component, and we added a required prop for it in order to specify an identifier. (You'd have to refactor the beta vs. current Overview page a bit to distinguish the two, instead of using a common `AcmPage` component.) We could keep the hook as well if we wanted to track lower granularity, like particular tabs on a page or something like that. This could provide a lot of good data. I think we would at least want to count how many times the cluster list (as the default entry page) is loaded to compare how often users are using other areas of the console.

backend/src/routes/metrics.ts
```diff
@@ -0,0 +1,27 @@
+/* Copyright Contributors to the Open Cluster Management project */
+
+import { Http2ServerRequest, Http2ServerResponse } from 'http2'
+import Prometheus, { register } from 'prom-client'
+
+interface PageData {
+  page: string
+}
```
KevinFCormier, line 8: Is this type used somewhere?

frontend/src/hooks/console-metrics.ts
```diff
@@ -0,0 +1,18 @@
...
+  overviewFleet = 'overview-fleet',
+  search = 'search',
+  searchDetails = 'search-details',
+}
+
+export function usePageVisitMetricHandler(page: Pages) {
+  useEffect(() => {
+    postRequest(getBackendUrl() + `/metrics?${page}`, {})
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
KevinFCormier, line 17: `page` should be in the dependency list and the eslint-disable removed.

backend/src/routes/metrics.ts
zlayne: Removed

frontend/src/hooks/console-metrics.ts
zlayne: added page to dependency array

PR #3015, opened 2023-09-01
ACM-7078 | chore: Indicate deprecation of OCP 4.13 on RHV

robdolares:
related Issue: https://issues.redhat.com/browse/ACM-7078


frontend/src/routes/Applications/Overview.tsx
```diff
@@ -1041,7 +1041,9 @@ export default function ApplicationsOverview() {
             action={
               <>
                 {appCreationButton}
-                <TextContent>{viewDocumentation(DOC_LINKS.MANAGE_APPLICATIONS, t)}</TextContent>
+                <div>
+                  <ViewDocumentationLink doclink={DOC_LINKS.MANAGE_APPLICATIONS} />
+                </div>
```
KevinFCormier, line 1046: With the current changes, I'm not sure why you would need the new `<div>` as `<TextContent>` was moved inside `<ViewDocumentationLink>`, so the resulting DOM tree should be the same?

frontend/src/routes/Credentials/CreateCredentialsCatalog.tsx
```diff
@@ -33,19 +32,31 @@ export function CreateCredentialsCatalog() {
 
   const cards = useMemo(() => {
     const newCards: ICatalogCard[] = [
-      ...orderedProviders
-        // does not display AWS s3 card
-        .filter(([id]) => id !== Provider.awss3)
```
KevinFCormier, line 38: Why did you remove this filtering?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateClusterCatalog/CreateClusterCatalog.tsx
```diff
@@ -166,9 +165,25 @@ export function CreateClusterCatalog() {
...
               <br />
-              <a href={DOC_LINKS.CREATE_CLUSTER_PREREQ} target="_blank" rel="noopener noreferrer">
-                {t('View documentation')} <ExternalLinkAltIcon />
-              </a>
+              <ViewDocumentationLink doclink={DOC_LINKS.CREATE_CLUSTER_PREREQ} />
+            </>
+          ),
+        }
+      }
+      if (provider === Provider.redhatvirtualization) {
```
KevinFCormier, line 173: Use "else if" here.

frontend/src/routes/Credentials/CreateCredentialsCatalog.tsx
```diff
@@ -33,19 +32,31 @@ export function CreateCredentialsCatalog() {
...
+            <>
+              {t(
+                'Red Hat Virtualization is deprecated as a host platform for OpenShift 4.13 and will be removed in the next release.'
+              )}
+              <br />
+              <ViewDocumentationLink doclink={DOC_LINKS.RHV_DEPRECATION} />
+            </>
+          ))
+      }
+    })
```
KevinFCormier, line 59: Personally I would do this during the map rather than looping through again, and base it on provider rather than ID. Something like:
```
...(provider === Provider.redhatvirtualization ? {alertTitle: ..., alertVariant: ..., ...} : {})
```

KevinFCormier (changes requested):
Great work here! I left a few nit comments, and please resolve the code smells (they look valid and easy to fix).

I have not tried out the code but I would like to do so once you address the issues.

Thanks

frontend/src/routes/Credentials/CreateCredentialsCatalog.tsx
robdolares: I assumed since the S3 credential card has been moved to a secondary path behind AWS, it was no longer needed on the main catalog page.  Can revert

frontend/src/routes/Applications/AdvancedConfiguration.tsx
```diff
@@ -690,18 +690,7 @@ export default function AdvancedConfiguration() {
...
         </Split>
-        <TextContent>
-          <Text
-            component={TextVariants.p}
-            style={{
-              textAlign: 'right',
-              display: 'inline-block',
-              width: '100%',
-            }}
-          >
```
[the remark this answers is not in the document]
robdolares: unintentional - adding back

frontend/src/routes/Credentials/CreateCredentialsCatalog.tsx
KevinFCormier: This is the code that prevents it from appearing in the top level.

frontend/src/routes/Credentials/CreateCredentialsCatalog.tsx
```diff
@@ -44,10 +44,25 @@ export function CreateCredentialsCatalog() {
...
+                alertContent: (
+                  <>
+                    {t(
+                      'Red Hat Virtualization is deprecated as a host platform for OpenShift 4.13 and will be removed in the next release.'
+                    )}
+                    <br />
+                    <ViewDocumentationLink doclink={DOC_LINKS.RHV_DEPRECATION} />
+                  </>
+                ),
+              } as ICatalogCard)
```
KevinFCormier, line 60: Simplest fix for this TypeScript issue is to change line 50 to `alertVariant: 'warning' as const,` or you could also change this line to `} as Partial<ICatalogCard>)`... it's not really an issue in this case, but in general we should avoid casting to a certain type. An `ICatalogCard` must have `id` and `title` props, which this object does not.

robdolares: Thank you, I struggled with this one longer than I care to admit.  Updating.

frontend/src/routes/Infrastructure/Clusters/ClusterSets/components/GlobalClusterSetPopover.tsx
```diff
@@ -1,22 +1,13 @@
...
   return (
-    <Popover
-      bodyContent={
-        <>
-          <Trans i18nKey="learn.global.clusterSet" components={{ bold: <strong /> }} />
-          <TextContent>{viewDocumentation(DOC_LINKS.GLOBAL_CLUSTER_SET, t)}</TextContent>
-        </>
-      }
-    >
+    <Popover bodyContent={<ViewDocumentationLink doclink={DOC_LINKS.GLOBAL_CLUSTER_SET} />}>
```
KevinFCormier, line 10: The string was accidentally removed here. (`<Trans>` component)

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateClusterCatalog/CreateClusterCatalog.tsx
```diff
@@ -166,9 +165,24 @@ export function CreateClusterCatalog() {
...
-                {t('View documentation')} <ExternalLinkAltIcon />
-              </a>
+              <ViewDocumentationLink doclink={DOC_LINKS.CREATE_CLUSTER_PREREQ} />
+            </>
+          ),
+        }
+      } else if (provider === Provider.redhatvirtualization) {
+        card = {
+          ...card,
+          title: t('Red Hat Virtualization'),
```
KevinFCormier, line 175: `card` should already have `title` set... I think this line can be removed.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
```diff
@@ -718,6 +724,45 @@ function versionGreater(version, x, y) {
...
+          )
+        }
+        return {}
+      },
+    })
+  }
+
+  return controlData
+}
+
```
KevinFCormier, line 765: I have tried to eliminate "controlData.splice" as much as possible. Side-effects are hard to follow, and it is inefficient because we are creating controlData, then searching through it to find the control we want and replacing it with a different version. You can see we've also duplicated code here.

Please take a look at the `getControlDataRHV` function. I think we can add this `info` prop directly there. There is a 3rd argument which will be `t`.

KevinFCormier: Also, note we usually only start param names with underscore if we are not going to use those parameters but need to specify them because we are going to use positional parameters that come later in the list. So `_controlData` should not start with underscore.

KevinFCormier (changes requested):
Hi Robert,

Just looking at the code, I noted a few little things to fix and some refactoring that will just make it a bit cleaner. I will try it out fully once you've made these changes and we get feedback from design.

Thanks

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/controlData/ControlDataHelpers.js
```diff
@@ -718,6 +724,24 @@ function versionGreater(version, x, y) {
...
+  if (control.active && control.active.includes('4.13')) {
+    return (
+      <AcmAlert
+        variant="warning"
+        title={t('Deprecated host platform')}
+        message={t(
+          'Red Hat Virtualization is deprecated as a host platform for OpenShift 4.13 and will be removed in the next release.'
+        )}
+        noClose={true}
+        style={{ marginTop: '2em' }}
```
KevinFCormier, line 737:
```suggestion
        style={{ marginTop: '2em' }}
        isInline
```

I think the drop shadow looks strange in the context of a form. Wonder if we should use inline version?
image

PR #3020, opened 2023-09-05
ACM-7291: Add destroy op

chenz4027:
https://issues.redhat.com/browse/ACM-7291

image
image
image
image


frontend/public/locales/en/translation.json
```diff
@@ -1440,6 +1441,7 @@
   "hostCount_plural": "{{count}} hosts",
   "Hosted": "Hosted",
   "Hosted cluster requires a manual import.": "Hosted cluster requires a manual import.",
+  "Hosted clusters cannot be destroyed.": "Hosted clusters cannot be destroyed.",
```
fxiang1, line 1444: I think this is kind of misleading since users can still destroy the hosted cluster from the CLI. Maybe we can change this to say:
`Hosted clusters cannot be destroyed from the console. Use the individual cluster destroy option to see CLI instructions.`

fxiang1 (changes requested):

frontend/public/locales/en/translation.json
chenz4027: Thanks @fxiang1 , yeah that is much more user friendly. Will add that.

fxiang1 (approved):
Thanks Maggie!

PR #3022, opened 2023-09-07
ACM-7347: update binary name for hcp for mce 2.4

dtthuynh:
https://issues.redhat.com/browse/ACM-7347
https://issues.redhat.com/browse/ACM-7348
https://issues.redhat.com/browse/ACM-7368

1. Hypershift cli renamed to hcp, need to update the example commands for both AWS/KubeVirt
2. Fixes some inconsistencies between Kubevirt/AWS screens, such as "Prerequisite and Configuration" vs "Prerequisite"
3. Hypershift enabled by default, no longer need to mention (you can't get to this screen without enabling or if the operator is not healthy)


fxiang1 (approved):
Thanks David!

PR #3031, opened 2023-09-12
ACM-7442: hide warning for valid clusters

chenz4027:
https://issues.redhat.com/browse/ACM-7442
image
image


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -415,16 +415,17 @@ export function ClustersTable(props: {
...
-              >
-                <TextContent>{t('They will not be destroyed when you perform this action.')}</TextContent>
-              </Alert>
-            ),
+            alert:
+              unDestroyedClusters.length > 0 ? (
+                <Alert
+                  variant="danger"
+                  isInline
+                  title={t('You selected {{0}} clusters that cannot be destroyed', [unDestroyedClusters.length])}
```
fxiang1, line 423: I would suggest changing `clusters` to `cluster(s)` if possible since it's not always more than one cluster.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -415,16 +415,17 @@ export function ClustersTable(props: {
...
+                  style={{ marginTop: '20px' }}
+                >
+                  <TextContent>{t('They will not be destroyed when you perform this action.')}</TextContent>
```
fxiang1, line 426: Then we can change this to:
`The cluster(s) will not be destroyed when you perform this action.`

Since `They` suggest multiple clusters but it might only be one.

fxiang1 (changes requested):

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
KevinFCormier: You can use plural support by calling the interpolation variable `count`. When you run the tooling to update the translations file, it will automatically create 2 entries, one for singular and one for plural. Just edit them to only use the `s` for the plural case.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -415,16 +415,19 @@ export function ClustersTable(props: {
...
+              unDestroyedClusters.length > 0 ? (
+                <Alert
+                  variant="danger"
+                  isInline
+                  title={t('You selected {{count}} cluster that cannot be destroyed', {
+                    count: unDestroyedClusters.length,
+                  })}
+                  style={{ marginTop: '20px' }}
+                >
+                  <TextContent>{t('They will not be destroyed when you perform this action.')}</TextContent>
```
fxiang1, line 428: @KevinFCormier What about this message? Is there a way to generate a singular and plural message without using the interpolation `count` variable?

KevinFCormier: Yes, as long as the `count` variable is passed in, it does not actually need to be referenced in the string. See this one for example:
`"This application set uses placement \"{{placement}}\", which is not removable. This placement is shared by the following application set:_plural": "This application set uses placement \"{{placement}}\", which is not removable. This placement is shared by the following application sets:"`

I don't recall if the scripting picks up on this or not, but you can add the `_plural` key manually if not.

fxiang1: Ah nice! Thanks @KevinFCormier
@chenz4027 Can you fix it for this message too? Thanks!

fxiang1 (changes requested):

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
chenz4027: Sure thanks @fxiang1 @KevinFCormier , I'll update it.

chenz4027: Done.

fxiang1 (approved):
Look great! Thanks Maggie!

PR #3032, opened 2023-09-13
MGMT-15570: Enable BMC options in Add hosts for all platforms when there's a provisioning config

batzionb:
AI lib PR: https://github.com/openshift-assisted/assisted-installer-ui/pull/2323


rawagner:
included in https://github.com/stolostron/console/pull/3033

PR #3043, opened 2023-09-15
Fix incorrect docs link for nmstate config

rawagner:
AI lib PR https://github.com/openshift-assisted/assisted-installer-ui/pull/2364


rawagner:
included in https://github.com/stolostron/console/pull/3049

PR #3048, opened 2023-09-18
Disable resource actions & details view for global hub

zlayne:
Related issue: https://issues.redhat.com/browse/ACM-6872

disable search result row actions and links to search details view.


KevinFCormier (commented):
Users do not necessarily have permission to list CRDs. For example, on a shared environment like the `collective` cluster, we do not. Is there a way this can be determined regardless of user RBAC?

zlayne:
@KevinFCormier The only way I can think of would be to use the service account token - as the console role includes CRDs for list/watch. 
I guess this would include adding a new backend route (maybe specifically of global hub? `/globalhub`?) that is protected by the serviceaccount token.

frontend/src/atoms.ts
```diff
@@ -187,3 +187,13 @@ export function useSearchAutocompleteLimit() {
   const settings = useRecoilValue(settingsState)
   return useMemo(() => parseInt(settings.SEARCH_AUTOCOMPLETE_LIMIT ?? '10000'), [settings])
 }
+
+export function useIsGlobalHub() {
+  const crds = useRecoilValue(customResourceDefinitionsState)
+  return useMemo(
+    () =>
+      crds.filter((crd) => crd.metadata.name === 'multiclusterglobalhubs.operator.open-cluster-management.io').length >
```
jlpadilla, line 195: - Do we really need to fetch all CRDs? It would be more efficient to filter on the server side.
- `crds.find(...)` is better because it stops the loop once it finds a match.

frontend/src/components/LoadData.tsx
```diff
@@ -478,6 +484,44 @@ export function LoadData(props: { children?: ReactNode }) {
...
+    }
+  }, [globalHubRes, globalHubLoading, setIsGlobalHub])
+
+  // If all data not loaded (!loaded) & events data is loaded (eventsLoaded) && global values is loaded (globalValuesLoaded) -> set loaded to true
+  useEffect(() => {
+    if (!loaded && eventsLoaded && globalValuesLoaded) {
+      setLoaded(true)
+    }
+  }, [loaded, setLoaded, eventsLoaded, globalValuesLoaded])
+
```
KevinFCormier, line 524: These 2 should not need to be inside `useEffect`. 

Did you need to introduce `globalValuesLoaded` state? I think you can just use `loading` from the `useQuery`.

KevinFCormier: In general, you can update state without being inside an effect. The current render will be abandoned and restarted with the new state. This is more efficient than an effect, which will only happen after the full render completes.

https://react.dev/learn/you-might-not-need-an-effect

frontend/src/components/LoadData.tsx
```diff
@@ -478,6 +484,44 @@ export function LoadData(props: { children?: ReactNode }) {
     }
   }, [setSettings, setters, setLoaded])
```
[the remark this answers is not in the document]
zlayne: From the React docs the useState setter fns do not change: 
https://legacy.reactjs.org/docs/hooks-reference.html#usestate
```
React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.
```

frontend/src/components/LoadData.tsx
zlayne: I removed the useEffect from the 2nd instance mentioned above. Removing useEffect from the first instance causes too many rerenders.

I included `globalValuesLoaded` thinking it could be used were we to introduce other global values in the future. With only the global hub values being used now I can remove the state and use loading from useQuery.

frontend/src/components/LoadData.tsx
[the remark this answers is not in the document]
KevinFCormier: So just remove `setLoaded` then.

frontend/src/components/LoadData.tsx
zlayne: Was able to remove the effect now that globalValuesLoaded was removed.

PR #3054, opened 2023-09-21
Restore the table columns to the default view

mprahl:
There was a bug where the "Restore defaults" button would only restore the required columns and not the default columns. As part of this change, the existing defaultColIds was renamed to requiredColIds. There is a new optional prop of defaultColIds that allows you to specify the columns that should be shown on the first visit and when the "Restore defaults" button is clicked.

In the "Policies" table, the default columns are a union of required columns and "isFirstVisitChecked" columns.

Relates:
https://issues.redhat.com/browse/ACM-7584


JustinKuli (approved):
LGTM

PR #3055, opened 2023-09-21
ACM-7447: Change texts for hosted clusters

chenz4027:
https://issues.redhat.com/browse/ACM-7447


chenz4027:
@dockerymick All the texts have been addressed. You may notice that there are some discrepancies and that's because some texts, for example `Enable Hosted Control Plane feature for Amazon Web Services (AWS)` has been deleted by @dtthuynh 's previous PR. The env I gave you was outdated. I also kept "Add credential" because that will open a pop up modal. Aside from that, everything should align with your proposal. Thanks again for all the enhancements!

fxiang1 (approved):
Thanks Maggie!

PR #3057, opened 2023-09-22
Update the pod must exist policy template

mprahl:
This updates it to use an official Red Hat image as well as set the proper security constraints for restricted configurations.

Relates:
https://issues.redhat.com/browse/ACM-7630

### DIFF
--- a/frontend/src/wizards/Governance/Policy/stable/CM-Configuration-Management/policy-pod.yaml
+++ b/frontend/src/wizards/Governance/Policy/stable/CM-Configuration-Management/policy-pod.yaml
@@ -25,15 +25,20 @@ spec:
             - complianceType: musthave
               objectDefinition:
                 apiVersion: v1
-                kind: Pod # nginx pod must exist
+                kind: Pod # httpd pod must exist
                 metadata:
-                  name: sample-nginx-pod
+                  name: sample-httpd-pod
                 spec:
                   containers:
-                    - image: nginx:1.18.0
-                      name: nginx
-                      ports:
-                        - containerPort: 80
+                    - image: registry.redhat.io/rhel9/httpd-24:latest
+                      name: httpd
+                      securityContext:
+                        allowPrivilegeEscalation: false
+                        capabilities:
+                          drop:
+                            - ALL
+                        privileged: false
+                        runAsNonRoot: true
 ---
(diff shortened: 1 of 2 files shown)


gparvin (commented):
Do we test this outside of openshift anywhere?  I guess all of our stable policies are tested on openshift.
FYI: I don't see any doc updates here that are required: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/governance/governance#pod-policy

We would need to also update the policy on policy-collection https://github.com/open-cluster-management-io/policy-collection/blob/main/stable/CM-Configuration-Management/policy-pod.yaml
I think there should be other tests that specifically test that policy.

mprahl:
> Do we test this outside of openshift anywhere? I guess all of our stable policies are tested on openshift. FYI: I don't see any doc updates here that are required: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/governance/governance#pod-policy
> 
> We would need to also update the policy on policy-collection https://github.com/open-cluster-management-io/policy-collection/blob/main/stable/CM-Configuration-Management/policy-pod.yaml I think there should be other tests that specifically test that policy.

My thought is that I didn't want to recommend running a non-Red Hat image in the ACM console. The policy-collection one can stay the same in my opinion.

yiraeChristineKim (approved):

frontend/src/wizards/Governance/Policy/stable/CM-Configuration-Management/policy-pod.yaml
```diff
@@ -25,15 +25,20 @@ spec:
...
+                  name: sample-httpd-pod
                 spec:
                   containers:
-                    - image: nginx:1.18.0
-                      name: nginx
-                      ports:
-                        - containerPort: 80
+                    - image: registry.redhat.io/rhel9/httpd-24:latest
+                      name: httpd
+                      securityContext:
```
yiraeChristineKim, line 35: is it ok skip 
```
    ports:
    - containerPort: 80
 ```

mprahl: Yes, it just provides information and the container images should specify the ports they expose.

PR #3060, opened 2023-09-25
Fix incorrect docs link for nmstate config

rawagner:
backport of https://github.com/stolostron/console/pull/3043


rawagner:
included in https://github.com/stolostron/console/pull/3131

PR #3062, opened 2023-09-25
Remove unnecessary tooltips in grc table

yiraeChristineKim:
The kebab menu displays exactly the same text on the tooltips, which is unnecessary.

Ref: https://issues.redhat.com/browse/ACM-5348


mprahl:
@yiraeChristineKim I think the intention of the ticket was to remove all the tooltips in that menu but I'd check with Ethan.

mprahl:
I think a tooltip for when the item is disabled is still good to explain why but when it's enabled it's the same text which isn't useful.

mprahl (approved):

PR #3067, opened 2023-09-26
MGMT-15817 Show postinstall info for CIM/Nutanix clusters

rawagner:


frontend/src/lib/doc-util.tsx
```diff
@@ -62,6 +62,7 @@ export const DOC_LINKS = {
   ENABLE_OBSERVABILITY: `${DOC_HOME}/html/observability/observing-environments-intro#enabling-observability-service`,
   RHV_DEPRECATION: `${OCP_DOC_BASE_PATH}/4.13/html/release_notes/ocp-4-13-release-notes#ocp-4-13-rhv-deprecations`,
   ACCESSING_CONSOLE: `${DOC_HOME}/html/web_console/web-console#accessing-your-console`,
+  NUTANIX_POST_INSTALL: `https://access.redhat.com/documentation/en-us/assisted_installer_for_openshift_container_platform/2022/html-single/assisted_installer_for_openshift_container_platform/index#nutanix-post-installation-configuration_assembly_installing-on-nutanix`,
```
mareklibra, line 65: Can we reuse a constant similar to the `DOC_HOME`?

rawagner: we cannot, as the `DOC_HOME` references ACM docs, but this link is referencing AI docs

mareklibra: I know, the question was about having something similar to the DOC_HOME, so a constant referencing the root of AI documentation. But never mind, we can add that with a next such link.

mareklibra (approved):

PR #3069, opened 2023-09-26
ACM-7180 Add Containerfile for RHTAP

KevinFCormier:
https://issues.redhat.com/browse/ACM-7180


Containerfile.mce.rhtap
```diff
@@ -0,0 +1,38 @@
...
+      name="multicluster-engine/console-mce-rhel8" \
+      summary="multicluster-engine-console-mce" \
+      io.openshift.expose-services="" \
+      io.openshift.tags="data,images" \
+      io.k8s.display-name="multicluster-engine-console-mce" \
+      maintainer="['acm-component-maintainers@redhat.com']" \
+      description="multicluster-engine-console-mce"
+#     version="${CI_CONTAINER_VERSION}" \
+#     upstream-ref="${CI_CONSOLE_UPSTREAM_COMMIT}" \
+#     upstream-url="${CI_CONSOLE_UPSTREAM_URL}" \
```
KevinFCormier, line 38: @gurnben I tried to base this off of downstream as much as possible, as @Kyl-Bempah said it will eventually replace downstream, plus we have an OOM issue with the RHTAP build currently, and the downstream Dockerfile was structured a bit differently with fewer layers... so maybe this will solve it? 

But anyway, does it make sense to be applying these labels, and are there equivalent variables available for `CI_CONTAINER_VERSION`, `CI_CONSOLE_UPSTREAM_COMMIT`, and `CI_CONSOLE_UPSTREAM_URL`?

Containerfile.mce.rhtap
```diff
@@ -0,0 +1,38 @@
+FROM registry.redhat.io/ubi8/nodejs-18:1 AS builder
```
Kyl-Bempah, line 1: Does this line need to change to brew.registry.redhat.io?

Containerfile.mce.rhtap
gurnben: I think there may be some analagous vars - @scoheb are there any env vars in the build context we could leverage?

arewm: There are no such variables that can be substituted in RHTAP. These were injected into the Dockerfile previously because there was another process with took the Dockerfile as input, filling in the variable templates, and produced the _actual_ Dockerfile. 

Instead, the buildah task will add the following labels at build time: https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L154-L160

KevinFCormier: Thanks @arewm @scoheb 

I have removed these.

Randy424 (approved):
/lgtm

PR #3082, opened 2023-09-29
ACM-7565 Add new tooltip text for Hypershift

fxiang1:
https://issues.redhat.com/browse/ACM-7565

- Display a different text when viewing Hypershift cluster
- Also update console docs on how to add new resources


chenz4027 (approved):
/lgtm

PR #3088, opened 2023-10-03
Nutanix day2: ensure only nutanix agents are shown

rawagner:


rawagner:
included in https://github.com/stolostron/console/pull/3105

PR #3092, opened 2023-10-03
ACM-7784 [WIP] Change search query to use proper kind casing

fxiang1:
https://issues.redhat.com/browse/ACM-7784

Part of the performance optimization for Amadeus and Intesa. After discussing with search, they suggested using all lowercase kind names will have performance hits.

- Change kind name casing for App table


KevinFCormier:
@fxiang1 do you want to pursue this, or should we close?

fxiang1:
@KevinFCormier Let's close this. I don't think I will have any time to come back to this anytime soon. Thanks!

PR #3104, opened 2023-10-06
ACM-7797 Increase app page search query limit

fxiang1:
https://issues.redhat.com/browse/ACM-7797

This won't fix the performance issue but it will allow the customer to get all the applications. This might even make the performance issue worse since we're retrieving more data. For now, this will resolve the issue with not seeing all the apps but we need to work with search on a long term solution.

- Increase limit on search query
- Correct resource casing for search query


chenz4027 (approved):
Thanks Feng!

PR #3116, opened 2023-10-12
ACM-6976 Backport fix to disable AWS hosted cluster card when Hypershift featu…

fxiang1:
…re is disable 2.8

https://issues.redhat.com/browse/ACM-6976

- Add disable logic to AWS hosted cluster card
- Fix missing TP label
- Removed direct call to list multiclusterengine, used recoilstate instead. This helps unit testing as the nocks were being call randomly causing random test failures


chenz4027 (approved):
/lgtm

PR #3121, opened 2023-10-16
ACM-7943 Only use multi arch OCP images for hosted clusters

fxiang1:
https://issues.redhat.com/browse/ACM-7943

Turns out the x86 images were overwriting the multi arch images.

- Change to only use multi arch images


chenz4027 (approved):
/lgtm
Thanks Feng

PR #3125, opened 2023-10-18
ACM-8215 Fix cluster curator prehook fail for hosted clusters

fxiang1:
https://issues.redhat.com/browse/ACM-8215

- Add check for hostedclusters when curator pre/post hook fails

prehookfailed
image


fxiang1:
Prow bot not adding lgtm label. Merging manually.

PR #3132, opened 2023-10-20
ACM-7920 yaml-editor-not-works-well

jeswanke:
caused because when there are errors in the yaml pasted into the editor, the yaml is ignored
then when the user went to change anything outside of the editor, the editor would be reset with the form version

fixed by prevent form updates until yaml is fixed--alert user:

CleanShot 2023-10-20 at 12 03 14@2x

### DIFF
--- a/frontend/src/components/SyncEditor/SyncEditor.tsx
+++ b/frontend/src/components/SyncEditor/SyncEditor.tsx
@@ -665,7 +665,7 @@ export function SyncEditor(props: SyncEditorProps): JSX.Element {
       {editorHasErrors && (
         <Alert
           variant="danger"
-          title="Changes will be ignored until errors are fixed."
+          title={t('Form edits will be ignored until YAML syntax errors are fixed.')}
           actionClose={<AlertActionCloseButton onClose={() => setEditorHasErrors(false)} />}
         />
       )}
--- a/frontend/src/components/SyncEditor/SyncEditor.tsx
+++ b/frontend/src/components/SyncEditor/SyncEditor.tsx
@@ -5,7 +5,7 @@ import useResizeObserver from '@react-hook/resize-observer'
 import { CodeEditor, CodeEditorControl, Language } from '@patternfly/react-code-editor'
 import { global_BackgroundColor_dark_100 as editorBackground } from '@patternfly/react-tokens'
 import { RedoIcon, UndoIcon, SearchIcon, EyeIcon, EyeSlashIcon, CloseIcon } from '@patternfly/react-icons'
-import { Alert, AlertActionCloseButton, ClipboardCopyButton } from '@patternfly/react-core'
+import { Alert, ClipboardCopyButton } from '@patternfly/react-core'
 import { debounce, noop, isEqual, cloneDeep } from 'lodash'
 import { processForm, processUser, ProcessedType } from './process'
 import { compileAjvSchemas, ErrorType, formatErrors } from './validation'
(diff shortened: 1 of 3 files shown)


frontend/src/components/SyncEditor/SyncEditor.tsx
```diff
@@ -652,12 +653,22 @@ export function SyncEditor(props: SyncEditorProps): JSX.Element {
...
     editorRef?.current?.layout({ width, height })
     setShowCondensed(width < 500)
   })
 
   return (
     <div ref={pageRef} className="sync-editor__container">
+      {editorHasErrors && (
+        <Alert
+          variant="danger"
+          title="Changes will be ignored until errors are fixed."
```
KevinFCormier, line 668: - New string needs to be translated
- Text could be clearer. Would this be accurate: "Form edits will be ignored until YAML syntax errors are fixed"

jeswanke (commented):
yeah--I was hoping you'd provide the title :)

frontend/src/components/SyncEditor/SyncEditor.tsx
```diff
@@ -652,12 +653,22 @@ export function SyncEditor(props: SyncEditorProps): JSX.Element {
     if (variant === 'toolbar') {
       height -= 36
     }
+    if (editorHasErrors) {
+      height -= 36
+    }
```
KevinFCormier, line 658: Can we combine these two blocks to avoid repeating this hard-coded height adjustment?

frontend/src/components/SyncEditor/SyncEditor.tsx
```diff
@@ -652,12 +653,22 @@ export function SyncEditor(props: SyncEditorProps): JSX.Element {
...
     setShowCondensed(width < 500)
   })
 
   return (
     <div ref={pageRef} className="sync-editor__container">
+      {editorHasErrors && (
+        <Alert
+          variant="danger"
+          title={t('Form edits will be ignored until YAML syntax errors are fixed.')}
+          actionClose={<AlertActionCloseButton onClose={() => setEditorHasErrors(false)} />}
```
KevinFCormier, line 669: If you dismiss the alert, it actually stops ignoring changes, so your changes will be lost if you touch something in the form. And, if you continue editing without fixing the YAML syntax, the alert re-appears immediately. Should we just remove the close button?

KevinFCormier (commented):
@jeswanke One nit, plus a concern with having a close button on the alert.

Also, I notice that if you insert bad YAML in the review step, you can still submit the form, but I think the bad changes will just be ignored.  And perhaps that's OK. I suspect blocking the submit is not trivial, so I don't want to overcomplicate this if it adds risk.

frontend/src/components/SyncEditor/SyncEditor.tsx
jeswanke: so, I forgot to adjust the editorHasErrors value-- it's meant to be the height of the error alert so that the editor isn't pushed down past the bottom of the window and for that alert is should be 75---it's not in a place I can dynamically determine th height

frontend/src/components/SyncEditor/SyncEditor.tsx
jeswanke: yep--I was going to use it as a way to let the user throw away their editor changes but now it's just confusing

evelinec:
We're going determine to build our first RC (or FC) build tomorrow morning, any open PR needs to be merged asap in order to be in the build. Thanks.

PR #3134, opened 2023-10-23
chore(i18n): update translations

KevinFCormier:
Adding latest translations from Memsource project https://cloud.memsource.com/web/project2/show/XrwvlyAFpAVGbJpCv02dh4


frontend/public/locales/ja/translation.json
```diff
@@ -1521,19 +1529,23 @@
...
   "Increases resiliency with closely interconnected control plane and worker nodes.": "密接に相互接続されたコントロールプレーンとワーカーノードにより、回復力が向上します。",
   "Indicates if updates to any of the application resources are subject to a deployment time window.": "アプリケーションリソースの更新がデプロイメント時間枠の影響を受けるかどうかを示します。",
   "Indicates if updates to the subscription resources are subject to an active or blocked deployment time window.": "サブスクリプションリソースへの更新が、アクティブまたはブロックされたデプロイメント時間枠の影響を受けるかどうかを示します。",
   "Info": "情報",
   "Inform": "通知",
+  "Inform Only": "通知のみ",
   "Inform policies": "ポリシーの通知",
-  "Inform policy": "ポリシーの通知",
   "Inform/Enforce": "通知/実施",
+  "Inform/Enforce/InformOnly": "Inform/Enforce/InformOnly",
```
KevinFCormier, line 1546: @mprahl @zyjjay This is causing confusion for translators. It has been translated in some languages but not others.

Are these enum values from an API? It looks to me like they have slightly different casing than the values they map to. I also notice both "Inform Only" and "InformOnly" are in the list of strings. 

If we want it to be translated, we should avoid camel case and use proper casing, e.g. for this case `Inform/Enforce/Inform only`.

For the short term, if we want it to be translated, I am compiling a list of translation issues and can ask that they be translated in the languages that were missed. Or if they should not be, I can override them with the English versions for now.

Could you open up an issue to remove these from translation or address the consistency issues in a future release?

robdolares (approved):
/lgtm

PR #3138, opened 2023-10-24
MGMT-16030 Filter out non-nutanix agents

rawagner:


rawagner:
included in https://github.com/stolostron/console/pull/3138

PR #3141, opened 2023-10-24
ACM-8257 Fix issue with suggestion selection of marked text

zlayne:
related bug: https://issues.redhat.com/browse/ACM-8257

### DIFF
--- a/frontend/src/routes/Home/Search/components/Searchbar.tsx
+++ b/frontend/src/routes/Home/Search/components/Searchbar.tsx
@@ -364,7 +364,7 @@ export function Searchbar(props: SearchbarProps) {
 
   /** add the text of the selected item as a new chip */
   const onSelect = (event: React.MouseEvent<Element, MouseEvent> | undefined, itemId?: string | number) => {
-    const selectedText = (event?.target as HTMLElement).innerText
+    const selectedText = (event?.currentTarget as HTMLElement).innerText
     addChip(selectedText, itemId?.toString())
     event?.stopPropagation()
     focusTextInput()


KevinFCormier:
@zlayne Did you investigate if this behaves the same for all OCP versions (and therefore PF versions) that we run on? Just wondering if it's always been broken, or something that behaves a little differently depending on PF version.

zlayne:
@KevinFCormier Yes, we have discovered this has been present since the searchbar rewrite (2.7). The issue occurs when the user clicks directly on the marked/highlighted text as seen in the attached image (user clicked in the `Config` in `ConfigMap`).

Using the previous `event.target.innerText` returns the element where the event started. In this case, the innerText is `Config` not `ConfigMap`. (We use the <mark> element)
Using `event.currentTarget.innerText` returns which element the event was attached to or the element whose eventListener triggered the event. So this contains the entire text of the suggestion.

The issue was in how I was handling the JS event, and to my knowledge, the Patternfly version shouldn't matter.

Screenshot 2023-10-24 at 6 46 07 PM

KevinFCormier:
@zlayne Great! Thank you for confirming.

> @KevinFCormier Yes, we have discovered this has been present since the searchbar rewrite (2.7). The issue occurs when the user clicks directly on the marked/highlighted text as seen in the attached image (user clicked in the `Config` in `ConfigMap`).
> 
> Using the previous `event.target.innerText` returns the element where the event started. In this case, the innerText is `Config` not `ConfigMap`. (We use the element) Using `event.currentTarget.innerText` returns which element the event was attached to or the element whose eventListener triggered the event. So this contains the entire text of the suggestion.
> 
> The issue was in how I was handling the JS event, and to my knowledge, the Patternfly version shouldn't matter.
> 
> Screenshot 2023-10-24 at 6 46 07 PM

PR #3143, opened 2023-10-25
ACM-8345 Fix placement expression missing value

fxiang1:
https://issues.redhat.com/browse/ACM-8345

- Fix placement expression missing value

We are past the translation schedule so I'm just putting in this when an invalid expression is detected. Will work on something better in 2.10.
image


chenz4027 (approved):
/lgtm

I guess the error will just be that short string for now, I'm good with that.

PR #3150, opened 2023-10-30
Update Overview page toggle text

zlayne:
Update overview switch text from "Beta view" to "Fleet view".
Screenshot 2023-10-30 at 3 12 51 PM


jlpadilla (approved):
/lgtm

@KevinFCormier Is it too late to include this change for translation?

KevinFCormier:
> /lgtm
> 
> @KevinFCormier Is it too late to include this change for translation?

Checking with Nelson and Eveline. I can request another translation update later today if they are OK with that.

PR #3211, opened 2024-01-03
[WIP] GlobalHub search updates

zlayne:
Update managedhub search field to use camel case (managedHub)

### DIFF
--- a/frontend/src/routes/Home/Search/__snapshots__/searchDefinitions.test.tsx.snap
+++ b/frontend/src/routes/Home/Search/__snapshots__/searchDefinitions.test.tsx.snap
@@ -2057,7 +2057,7 @@ Array [
           "apiversion": "v1",
           "cluster": "testCluster",
           "kind": "Node",
-          "managedhub": "testManagedHub",
+          "managedHub": "testManagedHub",
           "name": "node.1",
         }
       }
--- a/frontend/src/routes/Home/Search/searchDefinitions.test.tsx
+++ b/frontend/src/routes/Home/Search/searchDefinitions.test.tsx
@@ -312,7 +312,7 @@ test('Correctly returns all resource definitions', () => {
 test('Correctly returns resource with managedHub column', () => {
   const testItem = {
     cluster: 'testCluster',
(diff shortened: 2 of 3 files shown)


jlpadilla:
@zlayne I'd like to add these changes.

1. Add a feature flag like `featureGlobalSearch=true` that we can use to toggle the feature in the console. Currently, Global Search enabled automatically when Global Hub is present, that’s great for GA, but risky for Dev Preview because it’s not possible to turn off.  Probably we should also [document this optional feature here.](https://github.com/stolostron/console/blob/main/README.md#optional-features)
2. Set the default `globalSearchAPIEndpoint="/federated"`

backend/src/routes/search.ts
```diff
@@ -30,7 +30,10 @@ export async function search(req: Http2ServerRequest, res: Http2ServerResponse):
     const searchService = `https://search-search-api.${mch?.metadata?.namespace || namespace}.svc.cluster.local:4010`
 
     const searchUrl = process.env.SEARCH_API_URL || searchService
-    const endpoint = process.env.globalSearchAPIEndpoint || '/searchapi/graphql'
+    const endpoint =
+      process.env.globalSearchFeatureFlag === 'enabled' && process.env.globalSearchAPIEndpoint
+        ? process.env.globalSearchAPIEndpoint
+        : '/searchapi/graphql'
```
jlpadilla, line 36: Missing the default `/federated` when global search is enabled.

```suggestion
      process.env.globalSearchFeatureFlag === 'enabled' ?
        process.env.globalSearchAPIEndpoint || '/federated'
        : '/searchapi/graphql'
```

jlpadilla: Maybe this could be simpler and more flexible.

```
const endpoint process.env.searchApiEnpoint || (process.env.globalSearchFeatureFlag === 'enabled' ? '/federated' : /.searchapi/graphql'
```

Note that is changes the name to `searchApiEndpoint` to make it more generic.

jlpadilla (approved):
/lgtm
/hold

PR #3223, opened 2024-01-16
ACM-9466 Optimize app count for cluster details page

fxiang1:
https://issues.redhat.com/browse/ACM-9466 
More info this slack thread https://redhat-internal.slack.com/archives/C04EYRKRXGF/p1705346214521019

- Removed spreading all apps into a new array and just add the length together
- Changed discovered Argo and OCP app queries to only get apps from the current cluster

### DIFF
--- a/frontend/src/lib/search.ts
+++ b/frontend/src/lib/search.ts
@@ -53,17 +53,24 @@ export function queryStatusCount(cluster: string): IRequestResult<ISearchResult>
   })
 }
 
-export function queryRemoteArgoApps(): IRequestResult<ISearchResult> {
+export function queryRemoteArgoApps(cluster?: string): IRequestResult<ISearchResult> {
+  const filtersArr = [
+    { property: 'kind', values: ['Application'] },
+    { property: 'apigroup', values: ['argoproj.io'] },
+  ]
+
+  if (cluster) {
+    filtersArr.push({ property: 'cluster', values: [cluster] })
+  } else {
+    filtersArr.push({ property: 'cluster', values: ['!local-cluster'] })
+  }
+
   return postRequest<SearchQuery, ISearchResult>(getBackendUrl() + apiSearchUrl, {
     operationName: 'searchResult',
     variables: {
       input: [
         {
-          filters: [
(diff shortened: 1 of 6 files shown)


frontend/src/lib/search.ts
```diff
@@ -72,6 +72,25 @@ export function queryRemoteArgoApps(): IRequestResult<ISearchResult> {
...
+            { property: 'cluster', values: [cluster] },
+          ],
+          limit: 20000,
+        },
+      ],
+    },
+    query: searchFilterQuery,
+  })
+}
+
```
KevinFCormier, line 93: @fxiang1 Overall the PR looks great. I think it would be cleaner to combine this new function with the original, making the `cluster` param optional, both in terms of reducing duplication and simplifying where it is used in the code. Same thing for `queryOCPAppResourcesForCluster`.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/StatusSummaryCount.tsx
```diff
@@ -156,8 +154,13 @@ export function StatusSummaryCount() {
   }, [applicationSets, placementDecisions, cluster?.name])
 
   const appsCount = useMemo(
-    () => [...applicationList, ...clusterDiscoveredArgoApps, ...clusterOcpApps, ...appSets, ...argoAppList].length,
-    [applicationList, argoAppList, clusterDiscoveredArgoApps, clusterOcpApps, appSets]
+    () =>
+      applicationList.length +
```
KevinFCormier, line 158: Nice optimization.

KevinFCormier (commented):
I left a suggestion for improvement, so I didn't LGTM so that it doesn't auto-merge. But if you want to merge it as-is, I am OK with that, too. Just let me know. Thanks.

frontend/src/lib/search.ts
fxiang1: Thanks for catching this @KevinFCormier, I will work on it.

fxiang1:
The suggested changes are done. @KevinFCormier can you review again? Thanks!

PR #3236, opened 2024-01-22
ACM-9536 Wrap search queries functions in useCallback

fxiang1:
https://issues.redhat.com/browse/ACM-9536

- Wrap search queries in useCallback so they don't get recreated during every render


chenz4027 (approved):
/lgtm

PR #3252, opened 2024-02-02
throttle events

jeswanke:
resolves ACM-8455

the magic of the asm UI architecture is that if there's change in kube, it will be automatically and instantly reflected in the ui. The downside of this is if the UI is monitoring lots of resources and events happen all at once-- or if a resource is thrashing it's results or both

to fix this problem, if an event passes into the throttler:
1. if that policy hasn't ever had an event, the event it just sent on--the magic lives on
2. however if the same policy sends another event within 10 seconds, the throttle holds onto it
3. if after 10 seconds nothing more happens, that event is sent on
4. however if lots of events happen. we use standard deviation to see if the compliance is thrashing (a big sd = thrashing)
    a. if thrashing, the compliance is set to noncompliant
5. this event is then sent if it doesn't match the compliance in the event queue---iow client doesn't know the current state of the compliance

there is a policy thrasher in eventUtils.ts:
that is accessed through event.ts:
import { WatchEvent, ServerSideEventData, **thrashEvents,** purgeEvents, throttleEvents } from '../lib/eventUtils'

export function startWatching(): void {
  ServerSideEvents.eventFilter = eventFilter

  for (const definition of definitions) {
    void listAndWatch(definition)
  }
  // **thrashEvents(cacheResource)** <<<<<<<<<<<<<< uncomment me
}

### DIFF
--- a/backend/src/lib/eventUtils.ts
+++ b/backend/src/lib/eventUtils.ts
@@ -1,0 +1,288 @@
+/* Copyright Contributors to the Open Cluster Management project */
+import { IResource } from '../resources/resource'
+
+type CompliantType = 'Compliant' | 'NonCompliant' | 'Pending'
+
+const POLL_INTERVAL = 10
+
+export interface IPolicy extends IResource {
+  metadata: {
+    name: string
+    namespace: string
+    resourceVersion: string
+    uid: string
+    annotations?: Record<string, string>
+  }
+  spec: unknown
+  status?: {
+    compliant?: CompliantType
+    status?: { clustername: string; clusternamespace: string; compliant?: CompliantType }[]
+  }
+}
+
+export interface WatchEvent {
+  type: 'ADDED' | 'DELETED' | 'MODIFIED' | 'BOOKMARK' | 'ERROR'
+  object: IResource
+}
+
+export interface SettingsEvent {
+  type: 'SETTINGS'
+  settings: Record<string, string>
+}
+
+export type ServerSideEventData = WatchEvent | SettingsEvent | { type: 'START' | 'LOADED' }
+
+export interface Compliance {
+  type: 'COMPLIANT' | 'DELETED' | 'MODIFIED' | 'BOOKMARK' | 'ERROR'
+}
+
+interface IThrottleCache {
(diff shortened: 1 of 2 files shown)


KevinFCormier (changes requested):
I have a number of concerns with this approach.

1. When resources are thrashing, it is indicative of a bug in controllers or a configuration problem, and it causes over-utilization of cluster resources that should be addressed. It is not ideal that this causes problems in the UI, but on the other hand, symptoms in the UI have often been the canary in the coal mine. Just last week I identified a problem with `ManagedCluster` resources thrashing because of symptoms seen in the UI: https://issues.redhat.com/browse/ACM-9688
2. If we decide that the UI should be resilient to thrashing, it should be implemented generally, because any resource thrashing can cause issues. This implementation is only focused on policies, and specifically on the compliance state. I am not 100% certain that compliance was what was changing on the policies for this customer case, and it's not clear to me exactly what happens if some other part of a policy is rapidly changing.
3. I am concerned about manipulating the compliance state based on statistical analysis. I modified your simulation code to try something out and found a potential bug. Suppose a policy is thrashing and then the issue is resolved, such that the resource immediately stops updating and the final update puts it in a compliant state. See the code I used below. The policy remained noncompliant in the UI, and the purgeEvents code never removed the throttle (nor sent a final update). So there is a memory leak. I know the backend is already unbounded in memory usage, but we don't want to make matters any worse. I also wonder if there is potential for a final update to be sent but the compliance to be altered based on statistics, rather than on the last seen value.
4. There is no connection of the throttleCache to a deletion event. Is it possible that purgeEvents might call cacheResource for a resource that has been deleted?

jeswanke:
ok, how about this:
1. The UI shows the resource is thrashing--- instead of indicating Compliant/NonCompliant we add Fault or Error to indicate a thrashing policy-- I bet that would have helped out in your case and would definitely help out here since all the field can report is 7.5gb of background data over an hour
2. To be cautious I would start with just policy -- it seems important that the bug mentions policy a lot and that's the. kind of thing that was made to change a lot
3. 3. and 4. are bugs that can bee fixed

PR #3260, opened 2024-02-06
update err output when failed get logs

zhiweiyin318:
story: https://issues.redhat.com/browse/ACM-9195

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -371,6 +371,7 @@
   "access.users": "Users",
   "access.usersInGroup": "Users in group",
   "access.usersInGroup.view": "View users in group",
+  "acm.logs.error": "Red Hat Advanced Cluster Management version 2.10 and newer requires the <code>cluster-proxy</code> and <code>managed-serviceaccount</code> add-ons to retrieve logs. Make sure that these add-ons are enabled on the hub cluster.",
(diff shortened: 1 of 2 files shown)


KevinFCormier:
@zhiweiyin318 Please reach out to the docs team for help with the error string. In addition, it should be set up for translation. I also notice that there was a documentation link included before. Is there a new link we could point to? Ideally, the link should actually be formatted as a link. The logs error is currently stored as a string, but probably we could store it as a ReactNode instead.

zhiweiyin318:
@KevinFCormier thanks the comment. 
The previous link `https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_sfor_kubernetes/2.10/html-single/release_notes/red-hat-advanced-cluster-management-for-kubernetes-release-note#non-ocp-logs` cannot address `non-ocp-logs` page.  Not sure if we still have `non-ocp-logs` page.
And from 2.10, the log function can work on all clusters, so do not need differentiate OCP and non-OCP cluster. But depend on the cluster-proxy and managed-serviceaccount addons.

@oafischer  could you help review the error string ? thanks
/assign @oafischer

oafischer (approved):

zhiweiyin318:
Thanks @KevinFCormier for helping update the PR. I appreciate it so much.

zlayne (approved):

PR #3276, opened 2024-02-08
Fix filtering of actions related to auto-scaling machine pools

KevinFCormier:
https://issues.redhat.com/browse/ACM-7813


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ClusterDetails/ClusterMachinePools/ClusterMachinePools.tsx
```diff
@@ -194,9 +194,7 @@ export function MachinePoolsTable() {
         if (machinePool.spec?.autoscaling) {
           actions = actions.filter((action) => action.id !== 'scaleMachinePool')
           actions = actions.filter((action) => action.id !== 'enableAutoscale')
-        }
-
-        if (machinePool.spec?.replicas) {
```
KevinFCormier, line 199: This test did not work if the spec had `replicas: 0` which is the case for SNO. But `autoscaling` and `replicas` are mutually exclusive, so better to just check for `autoscaling`.

Randy424 (approved):
/lgtm

PR #3281, opened 2024-02-12
Remove DELETED event after broadcast to current clients

KevinFCormier:
https://issues.redhat.com/browse/ACM-9883


fxiang1 (approved):
Thanks Kevin! This is great catch! Hopefully it will solve some of the performance issues we've been having :)

PR #3288, opened 2024-02-13
Add support for IBMPower Clusters

aswinsuryan:
Add support for IBMPower Clusters


yboaron:
Overall code LGTM, 
QQ: do we need to change also something on submariner-repos to support IBM power ?

yboaron (approved):

KevinFCormier:
@aswinsuryan added a hold so you can answer @yboaron's question before merging, but the changes look good to me. Please go ahead and remove the hold when you are ready to merge.

aswinsuryan:
> Overall code LGTM, QQ: do we need to change also something on submariner-repos to support IBM power ?

it is not required , we are following the generic path , where an existing gateway will be tagged.

PR #3296, opened 2024-02-16
ACM-8935 Fix HostedCluster upgrade not displaying in dropdown

fxiang1:
https://issues.redhat.com/browse/ACM-8935

- Re-enable upgrading HC from dropdown
- Load the HypershiftUpgradeModal when upgrade is selected from the dropdown menu


chenz4027 (approved):
/lgtm

PR #3313, opened 2024-02-21
ACM-9906 Default namespace to clusters for KubeVirt

fxiang1:
https://issues.redhat.com/browse/ACM-9906

The hostedclusters created from the UI is created in their own namespace whereas hostedclusters created from the CLI is in clusters namespace. When a managedcluster CR gets deleted, I think there's a CLC controller that goes and deletes the namespace with the same name as the managedcluster and because the hostedcluster CR is in that namespace the hostedcluster get deleted as well. For now we will just default the namespace to clusters.

- Default namespace to clusters for KubeVirt


chenz4027 (approved):
/lgtm

PR #3322, opened 2024-02-27
ACM-10148 Fix KubeVirt ui creation fails if clusters ns doesn't exist

fxiang1:
https://issues.redhat.com/browse/ACM-10148

- Add project to YAML template
- Ignore namespace already exists error

This is not the best solution, need to revisit https://github.com/stolostron/console/blob/main/frontend/src/lib/create-cluster.ts in the next release for a better fix. I'm hesitant to making too many changes this late in the releaase in create-cluster.ts as it involves the Agent HostedCluster as well.


chenz4027 (approved):
/lgtm

PR #3196, opened 2023-12-08
Add View Yaml link of a new relatedobject that name is *

yiraeChristineKim:
Ref: https://issues.redhat.com/browse/ACM-8782

PR #3328, opened 2024-02-29
Change unnamed related object from * to -

yiraeChristineKim:
We just want one thing to possibly match, using "*" makes it seem like all of the things are bad. Using "-" also matches what we do for some things when we don’t know/care about the namespace

Ref: https://issues.redhat.com/browse/ACM-8782

Related: https://github.com/stolostron/console/pull/3196

### DIFF
--- a/frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
+++ b/frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
@@ -255,14 +255,13 @@ export function PolicyTemplateDetails(props: {
           ) {
             return ''
           }
-          if (cluster && kind && apiVersion && name) {
-            const nameArg = name != '*' ? `&name=${name}` : ''
+          if (cluster && kind && apiVersion && name && name != '-') {
             const namespaceArg = namespace ? `&namespace=${namespace}` : ''
             return (
               <a
                 target="_blank"
                 rel="noopener noreferrer"
-                href={`${NavigationPath.resourceYAML}?cluster=${cluster}&kind=${kind}&apiversion=${apiVersion}${nameArg}${namespaceArg}`}
+                href={`${NavigationPath.resourceYAML}?cluster=${cluster}&kind=${kind}&apiversion=${apiVersion}&name=${name}${namespaceArg}`}
               >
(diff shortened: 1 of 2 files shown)


dhaiducek (dismissed):

frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
```diff
@@ -256,7 +256,7 @@ export function PolicyTemplateDetails(props: {
             return ''
           }
           if (cluster && kind && apiVersion && name) {
-            const nameArg = name != '*' ? `&name=${name}` : ''
+            const nameArg = name != '-' ? `&name=${name}` : ''
```
JustinKuli, line 259: Should we disable the View YAML link in the case where the name is a hyphen? I don't think the list of all those resources is particularly helpful, it could be confusing.

mprahl: I think it's useful to see all the objects that are "eligible" for matching the ConfigurationPolicy. Especially if the evaluation interval is long, the result might be out of date but you can still see the search results.

I don't have a strong opinion though, so we could remove it.

JustinKuli (approved):
LGTM!

frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
JustinKuli: It would be nice, if the list was reasonably filtered, so that it was stuff that is "eligible" for matching the ConfigurationPolicy. But what we found during testing was that there were several cases where the list was too big (ie all namespaces or all deployments in the cluster), and it was more confusing than useful. I think it is better to remove it for now, and see if we can add it back in specific instances where it is useful.

dhaiducek: @JustinKuli It's a good point. Would you open an issue on it so the idea doesn't get lost in this thread?

yiraeChristineKim:
Already cherry picked to 2.10 https://github.com/stolostron/console/pull/3328

PR #3329, opened 2024-02-29
Remove 'View YAML' link when resource is unnamed

JustinKuli:
In some situations, a policy will list a group of objects as related to the policy via the name `-` or `*`. In those cases, linking to a YAML list of all of those objects is not generally helpful, since that list is not filtered by other details in the policy.

Refs:
 - https://issues.redhat.com/browse/ACM-10202


JustinKuli:
https://github.com/stolostron/console/pull/3328 is similar but better overall, if we need this, I think we'll add it there

PR #3410, opened 2024-04-10
ACM-10981 Fix appsub edit crash when placementrule name is missing

fxiang1:
https://issues.redhat.com/browse/ACM-10981

- Fix appsub edit crash when placementrule name is missing


chenz4027 (approved):
/lgtm

PR #3430, opened 2024-04-16
Export search results as CSV with default headers

zlayne:
Exports search results as CSV with headers: `name, namespace, kind, cluster, created, label`

Searchbar buttons have been condensed and an export dropdown is now added to initiate the csv download.

Screenshot 2024-04-16 at 2 02 06 PM


robdolares (approved):
/lgtm

PR #3431, opened 2024-04-16
Update the policy table name Cluster Violations to Cluster Compliance

yiraeChristineKim:
Currently it is named "Cluster Violations." It is confusing because the naming is not neutral while it generates "Violations" and "Non violations"
Update the name: "Cluster violations" to "**Cluster compliance**." Accordingly we could get the results as "Compliant" and "Non-compliant."

+

There were multiple issues with the sonar cloud.
- Coverage issues
- Code Complexity 
- Deep code 
etc 
This PR will fix these issues

Ref: https://issues.redhat.com/browse/ACM-3338

### DIFF
--- a/frontend/src/routes/Governance/policies/Policies.tsx
+++ b/frontend/src/routes/Governance/policies/Policies.tsx
@@ -40,6 +40,10 @@ import {
   hasInformOnlyPolicies,
   getPlacementBindingsForResource,
   getPlacementsForResource,
+  getSource,
+  PolicySetList, 
+  resolveExternalStatus,
+  resolveSource 
 } from '../common/util'
 import { checkPermission, rbacCreate, rbacUpdate, rbacPatch } from '../../../lib/rbac-util'
 import { transformBrowserUrlToFilterPresets } from '../../../lib/urlQuery'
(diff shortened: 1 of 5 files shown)


frontend/src/routes/Governance/policies/Policies.tsx
```diff
@@ -406,7 +406,7 @@ export default function PoliciesPage() {
         id: 'add-to-set',
         title: t('policy.table.actions.addToPolicySet'),
         click: (item) => {
-          setModal(<AddToPolicySetModal policyTableItems={...item} onClose={() => setModal(undefined)} />)
+          setModal(<AddToPolicySetModal policyTableItems={[...item]} onClose={() => setModal(undefined)} />)
```
yiraeChristineKim, line 409: I guess ts has been upgraded so it caused an error. This change fix the error

frontend/src/routes/Governance/policies/Policies.tsx
```diff
@@ -133,26 +140,80 @@ export default function PoliciesPage() {
     checkPermission(rbacUpdate(PolicyAutomationDefinition), setCanUpdatePolicyAutomation, namespaces)
   }, [namespaces])
 
+  const handleAutomationCell = useCallback(
```
mprahl, line 143: Would it be better to have this in PolicyTableCell.tsx instead?

yiraeChristineKim: I tried I am not sure that is good way... because I have to add these   
```
 policyAutomations,
      canCreatePolicyAutomation,
      t,
      setDrawerContext,
      setModal,
      canUpdatePolicyAutomation,
      unauthorizedMessage,
```
as parameters 
and 
also added these to useMemo 
```
const policyColumns = useMemo<IAcmTableColumn<PolicyTableItem>[]>(
```

yiraeChristineKim: The functions in PolicyTableCell are independent

yiraeChristineKim: no you're right let me move

mprahl (approved):

PR #3442, opened 2024-04-19
Recoil Refactoring

KevinFCormier:


fxiang1 (approved):
LGTM! Thanks Kevin! Looks like internally useRecoilState actually calls useRecoilValue so hopefully we can get some performance improvements by calling useRecoilValue directly :)

PR #3453, opened 2024-04-24
ACM-11046 Refactor table filters

KevinFCormier:
The main goal of this PR is to make table filters accessible in other parts of the code by moving handling to a reusable hook. This hook will be used to adjust the search queries for the applications table based on current filters.

The change also makes all filters appear in the location bar as query params, so that links to filtered results can be shared. This removes the need to handle links to filtered pages on a case-by-case basis.

Fixes several other minor issues:
- When following add-on status links from the Overview page to the Cluster list, the "Get started with Multicluster Hub" modal would appear for each new status. Local storage was keyed by full URL, which is unnecessary.
- If you applied a filter like `?cluster=foo` to the Applications page, but there is no cluster named "foo" then no filter chip would appear, but the results would still be filtered, so you would see no results. Filters are now always validated against what is currently available.
- If you filtered by a value that is no longer present, the chip for that filter will show even if `showEmptyOptions` is false if it has been saved in local storage. For example, if a user filtered by add-ons that are in degraded status, and then resolved the problem with add-ons, they will now be able to remove the filter. Previously it may have been stuck on.

### DIFF
--- a/frontend/src/routes/Applications/Overview.tsx
+++ b/frontend/src/routes/Applications/Overview.tsx
@@ -43,7 +43,6 @@ import {
   IAcmRowAction,
   IAcmTableColumn,
 } from '../../ui-components'
-import { getResourceParams } from '../Home/Search/Details/DetailsPage'
 import { useAllClusters } from '../Infrastructure/Clusters/ManagedClusters/components/useAllClusters'
 import { getArgoDestinationCluster } from './ApplicationDetails/ApplicationTopology/model/topologyArgo'
 import { DeleteResourceModal, IDeleteResourceModalProps } from './components/DeleteResourceModal'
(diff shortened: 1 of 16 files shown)


frontend/src/routes/Applications/Overview.tsx
```diff
@@ -699,32 +697,32 @@ export default function ApplicationsOverview() {
         options: [
           {
             label: t('Application set'),
-            value: `${getApiVersionResourceGroup(ApplicationSetApiVersion)}/${ApplicationSetKind}`,
+            value: 'appset',
```
KevinFCormier, line 700: @fxiang1 @chenz4027 Because these values will now appear in the browser location bar as the user applies filters, I wanted to make them more human-readable. Did I miss updating them anywhere, or do you see any issues with this? Obviously saved filters will be lost on upgrade to this new format, but that's a one time thing that I'm not concerned about.

chenz4027: I think it should be okay  to just use 'appset'

fxiang1: Yep should be fine to change the values.

fxiang1 (approved):

zlayne (approved):

PR #3464, opened 2024-05-02
Submariner add support for ROKS on Satelite

yboaron:
When deploying Submariner on ROKS IBM managed openshift on IBM Satelite loadbalancer should be set to false.

ROKS on Satelite is identified by nodes labeled with 'node.kubernetes.io/instance-type' equals to 'upi'.


yboaron:
cc @aswinsuryan

aswinsuryan (approved):

PR #3467, opened 2024-05-06
ACM-11007 Fix crash when no PlacementRules are in templateObject

fxiang1:
https://issues.redhat.com/browse/ACM-11007

- Fix crash when no PlacementRules are in templateObject


chenz4027 (approved):
/lgtm

PR #3482, opened 2024-05-14
Remove unnecessary global search checks

zlayne:
The global search checks were not needed. These resources are only available in the hub and thus search will always function correctly.


robdolares (approved):
/lgtm

PR #3507, opened 2024-05-22
ACM-9431 - Auto import discovered ROSA clusters

Randy424:
Regarding: https://issues.redhat.com/browse/ACM-9431

To test:
1. In `multicloud/credentials`, setup OCM discovery credentials. You will need discoverable and importable ROSA and OCP clusters in your OCM environment.
2. Go to the discovery tab (`multicloud/infrastructure/clusters/discovered`).
3. Click the action menu (kebab menu) for your ROSA/OCP cluster and click `Import cluster`.
4. For ROSA, import mode "Import from Red Hat OpenShift Cluster Manager" should be set, the cluster information should be pre-filled. Credential, ClusterID and Api token fields should be present.
5. For OCP import mode "Enter your server URL and API token for the existing cluster" should be set. 
6. For Import cluster from the cluster list page (not from a discovered cluster), the import mode should be manual.


frontend/public/locales/en/translation.json
```diff
@@ -1556,8 +1558,9 @@
   "import.command.launchconsole": "Launch console",
   "import.command.pendingimport": "Cluster is pending import",
   "import.command.runcommand": "Run a command",
+  "import.credential": "Credential",
   "import.credential.explanation": "Credentials are temporarily stored in a secret until the cluster import succeeds or fails.",
-  "import.description": "You are provided a command to run after you submit the form.",
+  "import.credential.place": "Select your Open Cluster Management credential",
```
KevinFCormier, line 1563: Wrong OCM here - it should be "Red Hat OpenShift Cluster Manager"

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -81,6 +81,7 @@ enum ImportMode {
   manual = 'manual',
   token = 'token',
   kubeconfig = 'kubeconfig',
+  discoveryROSA = 'discoveryROSA',
```
KevinFCormier, line 84: I would perhaps call this something like `discoveryOCM` as it may support other cluster types eventually.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -636,15 +710,49 @@ const AutoImportControls = (props: { state: State; dispatch: Dispatch<Action> })
...
+        api_token: newToken,
+        cluster_id: clusterID,
+      }
+      discoverySecret.type = 'auto-import/rosa'
+      resources.splice(deleteCount ? secretIndex : 1, deleteCount, discoverySecret)
+      setCredentialSelection(credentialName || '')
+      dispatch({
+        type: 'setApiToken',
+        api_token: newToken,
+      })
```
KevinFCormier, line 749: Is this left over from when we had the API token as it's own field?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -636,15 +710,49 @@ const AutoImportControls = (props: { state: State; dispatch: Dispatch<Action> })
...
+      const newToken =
+        ocmCredentials.find((credential) => credential.metadata.name === credentialName)?.stringData?.ocmAPIToken ?? ''
+      const discoverySecret = cloneDeep(autoImportSecret)
+      discoverySecret.stringData = {
+        ...discoverySecret.stringData,
+        api_token: newToken,
+        cluster_id: clusterID,
+      }
+      discoverySecret.type = 'auto-import/rosa'
+      resources.splice(deleteCount ? secretIndex : 1, deleteCount, discoverySecret)
```
KevinFCormier, line 744: Rather than repeating this logic here and moving the code that finds the index, could we have this function simply return the auto import secret, and do the splicing in the switch inside of `onChangeImportMode`? It just keeps the code that calculates the index and the code that uses it closer together, and it keeps the cases more parallel.

KevinFCormier (commented):
Awesome work! It works great. I left a few comments about code organization. There are also a couple functional issues we might want to address.

1. Credential should include the namespace, as it does in the configuration screens. (See screenshots.)
2. I think we should avoid passing the credential and the cluster ID to the form for non-ROSA clusters. If the user changes to use the RHOCM import mode and sees the pre-filled information, they might expect it to work.

image

image

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
Randy424: Thanks. I realized that including the token in state in general was unnecessary.

Randy424:
Currently making these changes: https://github.com/stolostron/console/pull/3507#pullrequestreview-2074164426, will update thread when they're completed.

Randy424:
> Awesome work! It works great. I left a few comments about code organization. There are also a couple functional issues we might want to address.
> 
> 1. Credential should include the namespace, as it does in the configuration screens. (See screenshots.)
> 2. I think we should avoid passing the credential and the cluster ID to the form for non-ROSA clusters. If the user changes to use the RHOCM import mode and sees the pre-filled information, they might expect it to work.
> 
> image
> 
> image

@KevinFCormier 
I've made some changes to try and adopt the behavior of the discovery configuration wizard, but it looks like react-form-wizard does not support the kind of field validation I am trying to enforce. Namely, I'm not sure how I can prevent users from progressing in the wizard when the credential field is empty. The react-form-wizard exposes it's validation context to it's proprietary components, and if I want to use a proprietary function it needs a path that corresponds to the template data. I am using fields that don't have a place within the template but still need to block the users progression, as to mirror the validation of the other fields. It looks like the **import mode** field has this same issue, but we don't notice it because it always contains a value.

Randy424:
@KevinFCormier This is ready for re-review! Thanks.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -636,15 +736,46 @@ const AutoImportControls = (props: { state: State; dispatch: Dispatch<Action> })
...
+      case ImportMode.discoveryOCM:
+        return t('import.mode.discovery')
     }
     return '-'
   }
 
+  const getImportModeHelperText = (m: ImportMode) => {
+    switch (m) {
+      case ImportMode.manual:
+        return t('import.mode.manual')
```
KevinFCormier, line 748:
```suggestion
        return t('import.description')
```
### Before PR
image

### After PR
image

Randy424:
https://github.com/stolostron/console/compare/1168adddd0d4d57c57bc0a74407f92798ebea3f9..ccbacaaf70f287c5d2c4bd3ed1e1ee59941c726d#diff-8a926d9ebe0a8a54381ef56ca3f270e5206bd156cfc55418093676894c5a509dL177-L181

This was handing values to namespace, credential and clusterID during regular import (no discovery cluster).

PR #3508, opened 2024-05-22
Create service account secret if not found

KevinFCormier:
Followed these instructions:
https://docs.openshift.com/container-platform/4.16/nodes/pods/nodes-pods-secrets.html#nodes-pods-secrets-creating-sa_nodes-pods-secrets

We might want to look into using the TokenRequest API instead, but I'm not sure if we can get the certificates that way - though they can be accessed by `oc debug` on the console pod if not.


setup.sh
```diff
@@ -28,6 +28,20 @@ INSTALLATION_NAMESPACE_MCE=`oc get multiclusterengine -A -o jsonpath='{.items[0]
 
 SA=$(oc get serviceaccounts -n $INSTALLATION_NAMESPACE_MCE console-mce -o jsonpath='{.metadata.name}')
 SA_SECRET=$(oc get secrets -n $INSTALLATION_NAMESPACE_MCE -o json | jq -r "[.items[] | select(.metadata.annotations[\"kubernetes.io/service-account.name\"] == \"$SA\" and .type == \"kubernetes.io/service-account-token\")][0].metadata.name")
+if [[ -z "$SA_SECRET" ]]; then
```
GowthamShanmugam, line 31: it looks like SA_SECRET is getting string value "null" not actual null, so command check is failing

fxiang1 (approved):
LGTM! Thanks Kevin for finding a solution quickly :)

PR #3510, opened 2024-05-22
remove deprecated provider rhv from catalog

dtthuynh:
RHV is deprecated starting with OCP 4.13, no longer can be used with OCP 4.14.
In MCE 2.6+, the support matrix will start at min OCP 4.14 version, so RHV is no longer supported.


Randy424:
@KevinFCormier should we also scrub the RHV templates/template map data, provider id and credential creation support? (Basically everything that comes up when we search "redhatvirtualization" or "RHV")

KevinFCormier:
> @dtthuynh @KevinFCormier should we also scrub the RHV templates/template map data, provider id and credential creation support? (Basically everything that comes up when we search "redhatvirtualization" and "RHV")

Yes, I would think so. Unless @dtthuynh  had a good reason why we should not?

PR #3513, opened 2024-05-27
Add namspace field when it is not presented

yiraeChristineKim:
Description of problem:
I tried to paste the policy yaml from a community policy into the console's policy editor. By convention, we don't provide a namespace in these policies, so the policy wizard complained there was no namespace.  Adding the namespace though did not reflect the change in the yaml editor.  Editing the namespace in the yaml editor did not change the policy wizard value.  I expect the namespace value to be handled clearly.

Ref: https://issues.redhat.com/browse/ACM-11387

### DIFF
--- a/frontend/src/components/SyncEditor/process.ts
+++ b/frontend/src/components/SyncEditor/process.ts
@@ -5,6 +5,7 @@ import { getErrors, validate } from './validation'
 import { getMatchingValues, getUidSiblings, crossReference, getPathArray } from './synchronize'
 import { reconcile } from './reconcile'
 import { ChangeType } from './changes'
+import { Pair } from 'yaml/types'
 
 export interface ProcessedType {
   parsed: {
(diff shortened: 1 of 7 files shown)


KevinFCormier:
Since this modifies SyncEditor, I think @jeswanke should review.

yiraeChristineKim:
> Since this modifies SyncEditor, I think @jeswanke should review.

Thanks!!

frontend/src/components/SyncEditor/process.ts
```diff
@@ -118,10 +119,34 @@ export const processUser = (
...
 ) => {
   // get yaml, documents, resource, mapped
-  const documents: any[] = YAML.parseAllDocuments(yaml, { prettyErrors: true, keepCstNodes: true })
+  let documents: any[] = YAML.parseAllDocuments(yaml, { prettyErrors: true, keepCstNodes: true })
+  let createNs = false
+  if (autoCreateNs) {
+    const newDocs = documents.map((doc: YAML.Document) => {
+      if (doc.has('metadata') && !doc.hasIn(['metadata', 'namespace'])) {
+        createNs = true
+        const pair = new Pair('namespace', '')
```
mprahl, line 132: I'm wondering if the default should be `open-cluster-management-global-set` instead. What do you think @JustinKuli, @dhaiducek, and @gparvin ?

JustinKuli: Does it currently set it to an empty string, and then presumably refuse the request / warn the user if they try to apply it as-is? I think that in some way, it should tell the user that they need to set the namespace. I don't really think we should "recommend" `open-cluster-management-global-set`.

gparvin: `open-cluster-management-global-set` is really too long to use as a default I think.  There is another concern regarding using ACM namespaces as a default policy namespace too.  This namespace can be deleted in a partial uninstall scenario that keeps CRDs and would leave policy CRs (except in ACM namespaces).  If it was a policy we provide out of the box or through our automation it would be more acceptable.

PR #3521, opened 2024-05-30
ACM-11046 Indicate when application list is truncated

KevinFCormier:
https://issues.redhat.com/browse/ACM-11046

- An alert is displayed when the search results exceeds the search limits, and it specifies whether Argo CD or Flux/OpenShift/Default OpenShift apps (or both) are missing
- When counts are indeterminate, the table adds a `+` to the number, both in the pagination control and the filter badges
- Search queries are disabled when not needed and scoped down to the current filters and search text
- React Query developer console is available when running in standalone mode

image
image
image


fxiang1 (approved):
LGTM! Thanks Kevin! Really like the way you solved the apps limit popup issue :)

zlayne (approved):

PR #3524, opened 2024-05-31
Use Spanish (World) es_001 for Spanish

KevinFCormier:
Memsource rejects the file upload if the `--target-lang` option does not match one of the languages that is set up in our project template.


Randy424 (commented):
/lgtm

Randy424 (approved):

PR #3576, opened 2024-06-19
ACM-10168 Remove required condition for AppSub placement cluster label selector

fxiang1:
https://issues.redhat.com/browse/ACM-10168

Removing the required condition for AppSub placement cluster label selector since the user might want to select all the clusters from the ClusterSet, in this case no labels are required.


chenz4027 (approved):
/lgtm

PR #3579, opened 2024-06-19
ACM-10914 Switch to react-router-dom-v5-compat

KevinFCormier:
This PR switches the code to use react-router-dom version 6 APIs via the react-router-dom-v5-compat package and completely removes react-router-dom version 5.

Other notable changes:
* Updated dynamic plugin SDK packages for compatibility with OpenShift 4.15
* Back-levelled webpack to address compatibility issue with new SDK (see https://redhat-internal.slack.com/archives/C032NLNEE8G/p1709120178593309?thread_ts=1707842989.599359&cid=C032NLNEE8G)
* Added getRoutePathFunction so that NavigationPath enum can be used to generate paths relative to parent route without repeating portions of paths
* Created SubRoutesRedirect component for redirecting to a parent route while preserving route params
* LostChangesPrompt for navigating away from dirty forms is disabled
* All params are substituted into NavigationPaths using generatePath function for type safety rather than string replace (TypeScript checking verifies correct params are provided)
* All routes are lifted to the plugin entry points so that routing is centralized. This requires use of Outlets and outlet contexts. One exception is conditional redirects, which are done using a SubRoutesRedirect in a child component
* Known issue: https://issues.redhat.com/browse/OCPBUGS-34811
    * Location state is not currently preserved, so cancel/back buttons in wizards that have more than one referrer may direct to the wrong page currently


Randy424:
Chunky PR! Am currently working on reviewing this. I've tested manually on stand alone, but will test some more locally in the plugin mode.

jeswanke (approved):
ah Outlet cleans things up

Randy424 (approved):
/lgtm

PR #3590, opened 2024-06-24
ACM-12427 Convert applications list to use backend API

jeswanke:
https://docs.google.com/document/d/1mD4vRxZajQc43R5lse7RIdYbuH5RIkQxMysqO4Vhy9s/edit


backend/src/app.ts
```diff
@@ -98,6 +100,7 @@ export function start(): Promise<Http2Server | undefined> {
   if (eventsEnabled) {
     startWatching()
   }
+  startAggregating()
```
KevinFCormier, line 103: Since aggregating depends on watched resources, this should probably be moved inside the if block. `eventsEnabled` is true for the MCE pods, and false for the ACM pods.

backend/src/routes/events.ts
```diff
@@ -39,7 +39,24 @@ type ServerSideEventData = WatchEvent | SettingsEvent | { type: 'START' | 'LOADE
...
+  })
+}
+
+export async function isAuth(resource: IResource, token: string): Promise<boolean> {
+  let allowed = await canListClusterScopedKind(resource, token)
+  if (allowed) return true
+  allowed = await canListNamespacedScopedKind(resource, token)
+  if (allowed) return true
+  allowed = await canGetResource(resource, token)
+  return allowed
```
KevinFCormier, line 56: This looks correct for local-cluster resources, but it is not correct for managed cluster resources. For managed clusters, we need to check if the user can create ManagedClusterView in the managed cluster namespace.

Compare what the `e2e-view-ns` user can see on the weekly cluster.

For OCP apps, I think at this point we have a faked resource, usually the Deployment. Does this vary? For local-cluster OCP apps, we need to think about what happens if the user can see some but not all of the resources.

backend/src/lib/pagination.ts
```diff
@@ -0,0 +1,144 @@
...
+      items = items.map(({ transform, ...keepAttrs }) => keepAttrs)
+    }
+
+    const results: IResultListView = {
+      page: rpage,
+      items,
+      itemCount,
+      filterCounts,
+      emptyResult,
+      isPreProcessed: true,
```
KevinFCormier, line 139: Why is this part of the API? It seems hard-coded to always be true.

backend/src/routes/aggregators/applicationsArgo.ts
```diff
@@ -0,0 +1,161 @@
...
+      } as IResource)
+    }
+  })
+
+  return apps
+}
+
+// get argo apps from search api in three queries with a second apart
+async function getPagedRemoteArgoApps(options: RequestOptions) {
+  // a,i,n,etc are used with the hope that the distribution of names is evenly matched between queries
```
KevinFCormier, line 134: Most resource names permit a numeric character as the first character, including Application for Argo CD, so this scheme to ease the load on search will miss some applications  in its current form (See https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names; by attempting to create an Application with an uppercase character, I gleaned from the error message that they use RFC 1123)

I wonder if this is necessary? Perhaps we should leave it out for now or work with search to provide a real paging solution.

backend/src/routes/aggregators/applications.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  // Argo AppSets
+  const argoAppSets = getKubeResources('ApplicationSet', 'argoproj.io/v1alpha1')
+
+  // OCP Apps
+  const ocpApps = await getOCPApps(argoAppSet)
+
+  const filterCounts: FilterCounts = {}
+  const subscriptions = getKubeResources('Subscription', 'apps.open-cluster-management.io/v1')
+  const placementDecisions = getKubeResources('PlacementDecision', 'cluster.open-cluster-management.io/v1beta1')
```
KevinFCormier, line 64: The original frontend code looks at Channels, PlacementRules, and Placements as well... I wonder if we are missing something here.

KevinFCormier (commented):
Left a few inline comments for your records. We can discuss a bit more in person.

backend/src/lib/pagination.ts
KevinFCormier: (Future enhancement - for small data sets, frontend does the processing.)

backend/src/routes/events.ts
jeswanke: fixed

backend/src/routes/events.ts
```diff
@@ -39,14 +39,65 @@ type ServerSideEventData = WatchEvent | SettingsEvent | { type: 'START' | 'LOADE
...
+    await canAccess(
+      { kind: 'ManagedClusterView', apiVersion: 'view.open-cluster-management.io/v1beta1' },
+      'create',
+      token
+    )
+  ) {
+    return resources
+  }
+  return []
+}
```
KevinFCormier, line 77: This needs to check the specific namespace for the cluster that the resources come from.

backend/src/routes/events.ts
KevinFCormier: This needs to check if the resource can be created in the namespace for a specific managed cluster. For example, I created a `Role` and `RoleBinding` to grant `e2e-admin-ns` access to create a `ManagedClusterView` in the `weekly-managed` namespace. This should allow the user to see apps from that cluster, but your code is only checking if the user can create this resource in _any_ namespace.

KevinFCormier (commented):
Left a comment about the RBAC, which is not quite correct yet.

For the check failure, there is a TS issue. I would recommend updating `getBackendUrl` to return `string` only, rather than `string | undefined`:
```
export function getBackendUrl() {
  if (process.env.MODE === 'plugin') {
    const proxyPath = process.env.PLUGIN_PROXY_PATH
    const value = proxyPath ? `${proxyPath}${process.env.REACT_APP_BACKEND_PATH}` : ''
    return value
  }
  return process.env.REACT_APP_BACKEND_PATH ?? ''
}
```

We should also get the permissions changes for the ServiceAccount merged before we merge this code.

Thanks

backend/src/routes/events.ts
jeswanke: done

KevinFCormier (changes requested):
@jeswanke this is quite impressive overall. I have some more files to review and more testing to do, but I left some comments for a few bugs I found so far.

frontend/public/locales/en/translation.json
```diff
@@ -1073,7 +1072,6 @@
   "Decisions": "Decisions",
   "decoupled control plane and workers": "decoupled control plane and workers",
   "dedicated control plane nodes": "dedicated control plane nodes",
-  "Default OpenShift": "Default OpenShift",
```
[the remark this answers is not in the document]
fxiang1: John and I had a discussion about this and we thought users might be confused by this filter since it actually never appears on the Application table Type column. Maybe we need to word it better?

jeswanke:
new pr

backend/src/lib/search.ts
```diff
@@ -0,0 +1,126 @@
...
+  ['t*'],
+  ['u*'],
+  ['l*'],
+  ['m*'],
+  ['c*'],
+  ['p*'],
+  ['z*'],
+  ['d*', 'b*', 'g*', '0*', '1*', '2*', '3*', '4*', '5*'],
+  ['h*', 'k*', 'y*', 'v*', 'w*', 'f*', '6*', '7*', '8*', '9*'],
+]
```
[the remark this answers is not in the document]
KevinFCormier: Addressed in new PR https://github.com/stolostron/console/pull/3753

frontend/src/resources/utils/get-cluster.ts
```diff
@@ -482,7 +475,6 @@ export function getCluster(
     acmDistribution,
     microshiftDistribution: getMicroshiftDistributionInfo(managedCluster),
     acmConsoleURL: getACMConsoleURL(acmDistribution.version, consoleURL),
-    addons: getAddons(managedClusterAddOns, clusterManagementAddOns),
```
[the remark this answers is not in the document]
KevinFCormier: Addressed in new PR https://github.com/stolostron/console/pull/3753

frontend/public/locales/en/translation.json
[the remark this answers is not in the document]
KevinFCormier: Introduced "System" category in new PR https://github.com/stolostron/console/pull/3753

PR #3591, opened 2024-06-24
ACM-12311 Fix Kubevirt import error in import modal

fxiang1:
https://issues.redhat.com/browse/ACM-12311

This is caused by the UI trying to create a KlusterletAddonConfig CR in MCE only clusters. The KlusterletAddonConfig kind does not exist in MCE so we get the no resource found error. The fix is to ignore the error for MCE only clusters.


chenz4027 (approved):
/lgtm 
👍

PR #3596, opened 2024-06-25
ACM-12215-Provisioning-OpenShift-on-OpenStack-fails

jeswanke:


frontend/src/routes/Credentials/CredentialsForm.tsx
```diff
@@ -403,7 +413,7 @@ export function CredentialsForm(
         stringData.additionalTrustBundle = additionalTrustBundle
         break
       case Provider.openstack:
-        stringData['clouds.yaml'] = enforceCloudsYaml(cloudsYaml, cloud, osCABundle)
+        stringData['clouds.yaml'] = cloudsYaml
```
KevinFCormier, line 416: Is there a bug you are solving by moving this logic, rather than just fixing the YAML structure? I remember this had very specific behaviour, so I worry about changing it.

KevinFCormier (commented):
See 1 inline comment.

Also, can you please update the error messages that reference the cacert value being in the "cloud auth" key.

frontend/src/routes/Credentials/CredentialsForm.tsx
jeswanke: yes -- by doing it on line 406 it updates just the yaml file and not the wizard ui input for "clouds.yaml"

PR #3597, opened 2024-06-25
Fix download for international Spanish

KevinFCormier:
Need to use `es_001` as target language in place of `es`.


robdolares (approved):
/lgtm

PR #3599, opened 2024-06-25
Updated ja ko zh strings

KevinFCormier:
Adding latest translations from Memsource project https://cloud.memsource.com/web/project2/show/KDRbbxxUovqk2heUO0yKl2

ja, ko, zh ONLY

fr and es_001 are still being finalized


robdolares (approved):
/lgtm

PR #3601, opened 2024-06-25
ACM-10432 Remove required for AppSet labels

fxiang1:
https://issues.redhat.com/browse/ACM-10432

I'm removing the required asterisk from the Placement label selection because:

1. The Placement API actually doesn't require them to be filled ie. you can leave them blank and the Placement CR will still be created
2. I don't think there's a good way for us to validate the control being required


chenz4027 (approved):
/lgtm

PR #3625, opened 2024-07-04
ACM-9517 Fix AppSub Ansible jobs not displaying

fxiang1:
https://issues.redhat.com/browse/ACM-9517

The AppSub backend changed the way AnsibleJobs are stored in the Subscription CR. They now only store the last job that ran as oppose to all jobs. I updated the UI to support this new change.


chenz4027 (approved):
/lgtm

PR #3640, opened 2024-07-09
ACM-12051 - implement feedback mechanism for acm/mce 

robdolares:
related issue: https://issues.redhat.com/browse/ACM-12051


frontend/src/components/AcmFeedbackModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const AcmFeedbackModal = (props: FeedbackModalProps) => {
+  const { t } = useTranslation()
+  const i18nStrings = {
+    getSupport: t('Get support'),
+    cancel: t('Cancel'),
+    helpUsImproveHCC: t('Help us improve Red Hat Advanced Cluster Management for Kubernetes.'),
+    howIsConsoleExperience: t('What has your console experience been like so far?'),
+    openSupportCase: t('Support Case'),
+    problemProcessingRequest: t(
```
KevinFCormier, line 30: I think we can omit this one because we are not using any of the functionality that depends on callbacks.

frontend/src/components/AcmFeedbackModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+      'There was a problem processing the request. Try reloading the page. If the problem persists, contact'
+    ),
+    shareFeedback: t('Share feedback'),
+    somethingWentWrong: t('Something went wrong'),
```
KevinFCormier, line 34: I think we can omit this one because we are not using any of the functionality that depends on callbacks.

frontend/src/components/PluginContextProvider.tsx
```diff
@@ -69,16 +98,24 @@ export function PluginContextProvider(props: { children?: ReactNode }) {
...
     >
-      <div style={{ position: 'relative', height: '100%', width: '100%' }}>
-        <div style={{ position: 'absolute', height: '100%', width: '100%' }}>
-          <AcmToastProvider>
-            <AcmToastGroup />
-            <AcmTablePaginationContextProvider localStorageKey="clusters">
-              {props.children}
-            </AcmTablePaginationContextProvider>
-          </AcmToastProvider>
+      <AcmFeedbackModalProvider>
```
KevinFCormier, line 101: What is the purpose of this context provider? I do not see any consumers for the context, and removing it seemed to have no effect.

KevinFCormier (changes requested):
Hey @robdolares,

Thanks for this, it seems to work well. There are a few questions I have about the functionality.

1. Is the feedback button horizontal intentionally? 
image
2. One criteria does not seem to be met: "The feedback link is accessible, meaning it can be accessed via keyboard navigation and has appropriate labels for screen readers." This works in HCC, but I am not sure if it does in ACS. Any thoughts on this? We could open a minor defect to track this.
3. Nit: The feedback modal does not appear in standalone mode. It might be a bit tricky to refactor this, but I think we could probably find a way to do it that would also reduce boilerplate in other parts of the code. Let's leave this for now, but if I have some cycles I may request your review on a PR to do that.

For code quality, it looks pretty good. Please see my inline comments, and also, please take a look at the 3 Sonarcloud issues. I think they are all valid.

robdolares:
@KevinFCormier 
1. I have the feedback button as horizontal since that was ACS' implementation.  But I know the screen shot in the parent story has vertical so I made the change and we can lock that in or I can revert.
2. I am able to tab to the feedback button as well as tab through everything on the modal, but cannot trigger event on either external card (link) in the modal (only 'cancel' and 'X' work with space).  Will look more into this.
3. Can discuss further later.

frontend/src/components/PluginContextProvider.tsx
robdolares: removing

KevinFCormier:
> @KevinFCormier
> 
> 1. I have the feedback button as horizontal since that was ACS' implementation.  But I know the screen shot in the parent story has vertical so I made the change and we can lock that in or I can revert.

Looks like the rotation was present in the ACS implementation:
https://github.com/stackrox/stackrox/pull/6284/files#diff-6f98d028ca64ce6c035a0784ac21e5eaee2f26296fc19e7fd0509f737be05999R95

> 2. I am able to tab to the feedback button as well as tab through everything on the modal, but cannot trigger event on either external card (link) in the modal (only 'cancel' and 'X' work with space).  The demo on PF site has similar behavior.

I cannot tab to the feedback button. Can you elaborate on on how you did this? This works in HCC, but again I'm not sure if it does in ACS.

For the contents of the modal, that comes from a library so I'm not concerned. Although, if the modal itself is not usable by keyboard, then who cares if the button is? Perhaps it will be fixed in the PF5 version. Anyway, I won't hold up the PR for this.

> 3. Can discuss further later.

Yup.

At this point the check is failing (you forgot to re-run `npm run i18n:fix`) and there is still 1 Sonar issue to address.

PR #3643, opened 2024-07-10
AcmTable csv export button addition

Randy424:
Regarding: https://issues.redhat.com/browse/ACM-12099?filter=-1

PR 1/7

Added exportContent prop to AcmTable, enabling the component to consume a callback that returns the column value's export string. 

When used (see subsequent PR's for the implementation of this feature) the export button appears at the left end of the toolbar after the Manage columns button.

Screenshot 2024-07-10 at 2 15 24 AM

### DIFF
--- a/frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
@@ -17,7 +17,6 @@ import {
   DragDrop,
   Droppable,
   Draggable,
-  ToolbarItem,
 } from '@patternfly/react-core'
 import { IAcmTableColumn } from './AcmTable'
 import { useTranslation } from '../../lib/acm-i18next'
@@ -50,7 +49,7 @@ export function AcmManageColumn<T>({
   }
 
   return (
-    <ToolbarItem>
+    <>
       <ManageColumnModal<T>
         {...{
           isModalOpen,
@@ -67,7 +66,7 @@ export function AcmManageColumn<T>({
       <Tooltip content={t('Manage columns')} enableFlip trigger="mouseenter" position="top" exitDelay={50}>
         <Button isInline variant="plain" onClick={toggleModal} icon={<ColumnsIcon />} aria-label="columns-management" />
       </Tooltip>
-    </ToolbarItem>
+    </>
   )
 }
 
--- a/frontend/src/ui-components/AcmTable/AcmTable.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmTable.tsx
@@ -32,7 +32,7 @@ import {
   ToolbarItem,
   TooltipProps,
 } from '@patternfly/react-core'
-import { FilterIcon } from '@patternfly/react-icons'
+import { ExportIcon, FilterIcon } from '@patternfly/react-icons'
 import CaretDownIcon from '@patternfly/react-icons/dist/js/icons/caret-down-icon'
 import {
   expandable,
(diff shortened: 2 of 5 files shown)


KevinFCormier:
/hold
Reviewing...

frontend/package.json
```diff
@@ -95,6 +95,7 @@
     "uuid": "8.3.2",
     "validator": "13.7.0",
     "whatwg-fetch": "3.6.2",
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz",
```
KevinFCormier, line 98: I do not think this will work in our downstream builds because Cachito probably does not cache this package. They have not really provided a good explanation of why their package is not being published on npm.
https://github.com/SheetJS/sheetjs/issues/2667

This package does a lot more than we need, and while I'm usually in favour of using a library, since we already have CSV export functionality in our UI that doesn't depend on a package, I think we should share that code instead.

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
```diff
@@ -23,6 +23,14 @@ import { IAcmTableColumn } from './AcmTable'
 import { useTranslation } from '../../lib/acm-i18next'
 import ColumnsIcon from '@patternfly/react-icons/dist/js/icons/columns-icon'
 
+type PatternFlySpacingSetting = {
```
KevinFCormier, line 26: Don't duplicate type definitions when possible. We can just refer to `ToolbarItemProps['spacer']`. However, I have other suggestions that would avoid needing to add a prop for this at all.

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
```diff
@@ -50,7 +60,7 @@ export function AcmManageColumn<T>({
   }
 
   return (
-    <ToolbarItem>
+    <ToolbarItem spacer={toolbarItemSpacerSetting}>
```
KevinFCormier, line 63: I don't think the `<ToolbarItem>` component should be included in this component. I would move it up into the table. This component shouldn't be restricted to only appearing in a toolbar.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1162,8 +1217,45 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
               <AcmManageColumn<T>
                 {...{ selectedColIds, setSelectedColIds, requiredColIds, defaultColIds, setColOrderIds, colOrderIds }}
                 allCols={columns.filter((col) => !col.isActionCol)}
+                toolbarItemSpacerSetting={showExportButton ? { default: 'spacerNone' } : undefined}
```
KevinFCormier, line 1220: I think a more elegant solution than conditionally adding space to the right of this item would be to put the managed columns button and export dropdown in a group of their own:
```
<ToolbarGroup spaceItems={{ default: 'spaceItemsNone' }}>
```

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1162,8 +1217,45 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+                      id="export-search-result"
+                    >
+                      <ExportIcon />
+                    </DropdownToggle>
+                  }
+                  isOpen={isExportMenuOpen}
+                  isPlain
+                  dropdownItems={[
+                    <DropdownItem
+                      style={{ width: '10rem' }}
```
KevinFCormier, line 1248: What is this for? Seems to add space on the right of the "Export as CSV" text.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1162,8 +1217,45 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+                      key={'item.text'}
```
KevinFCormier, line 1249:
```suggestion
                      key='export-csv'
```

KevinFCormier (commented):
Looks great overall. Thank you for organizing into separate PRs! 

Please see inline comments.

frontend/src/ui-components/AcmTable/AcmTable.tsx
Randy424: Excellent solution, thank you.

frontend/src/ui-components/AcmTable/AcmTable.tsx
Randy424: It was for just that...we don't necessarily need it, so I've removed it but it was mostly to agree with the stylization happening on the search page export dropdown.

Randy424:
Running into a test failure on this component: src/routes/Infrastructure/Clusters/ManagedClusters/components/DownloadConfigurationDropdown.test.tsx
It's a undefined error caused from attempting to read a ClusterStatus enum from the get-cluster page...

...It's reproducible locally, but I don't yet see how we have caused this issue.

Randy424:
Looks like it's related to my use of the createDownloadFile() from the utils library: frontend/src/resources/utils/utils.ts 
Strange.

Randy424:
I think I fixed the issue, but the testing machine pools are shutting down?
Will test again on Monday :/

PR #3644, opened 2024-07-10
Update search SDK and apollo version

zlayne:
graphql 16.4.0 -> 16.9.0
@apollo/client 3.5.10 -> 3.10.8

Updated search-sdk using `graphql-codgen`

Removed `searchResultCountAndRelatedCount` query as it was not used.
Added `searchResultItemsAndRelatedItems` query for Apps (was previously using `searchResultRelatedItems`). They are using both search result items and related items from the query. And the `searchResultRelatedItems` query *should* only return the related items (and with these changes it will only return related items now).


zlayne:
FYI @jlpadilla

sonar-project.properties
```diff
@@ -2,7 +2,7 @@ sonar.projectKey=open-cluster-management_console
 sonar.projectName=console
 sonar.organization=open-cluster-management
 sonar.sources=frontend/src,backend/src
-sonar.exclusions=node_modules/**/*,frontend/node_modules/**/*,frontend/src/atoms.tsx, frontend/src/lib/nock-util.ts, frontend/src/ui-components/**/*.stories.tsx
+sonar.exclusions=node_modules/**/*,frontend/node_modules/**/*,frontend/src/atoms.tsx,frontend/src/lib/nock-util.ts,frontend/src/ui-components/**/*.stories.tsx,frontend/src/routes/Home/Search/search-sdk/search-sdk.ts
```
KevinFCormier, line 5: Should we really permanently exclude this file from Sonar scanning? Is this a test coverage issue or something else?

zlayne: My thinking was that this file is auto-generated from graphql-codegen. The coverage is low on this file mainly due to the type definitions not being very well covered, but the other unit tests in the console repo call the queries defined in this file.

KevinFCormier: OK, I see.

PR #3651, opened 2024-07-11
Remove Compliance wording messages from GRC UI pages

yiraeChristineKim:
Policies table
- Change column "Cluster compliance" back to "Cluster violations"
Policy details > Results > individual cluster policy template details (View details link)
- In Template details, change "Compliant" field name to Violations; and the result should be mirror the previous page with either "With violations" or "Without violations"
In Related resources table, change the "Compliant" column header; to "Violations" with values of "With violations" or "Without violations", same reasoning as above.
Ref: https://issues.redhat.com/browse/ACM-11787


dhaiducek (approved):
/hold for other reviews, but this LGTM!

mprahl (approved):

PR #3663, opened 2024-07-16
[ACM-12778] Elevate search link in left hand navigation

zlayne:
Related story - https://issues.redhat.com/browse/ACM-12778 

- Elevates Search link in left-hand nav - as seen in image.
- Removes search link from clusters page header.

Screenshot 2024-07-16 at 4 08 54 PM


KevinFCormier (commented):
Can you please link this to a story? At the very least, QE will need to be aware of these changes.

Up until now, the routes have matched the left nav structure. What do you think about removing "home" from the search route and putting a redirect in place on the old routes for a few releases?

zlayne:
@KevinFCormier I was considering changing the link, and think the redirect would work fine. But was thinking in that case we would also want to change the code structure and bring all search files out of the home folder.

KevinFCormier:
> @KevinFCormier I was considering changing the link, and think the redirect would work fine. But was thinking in that case we would also want to change the code structure and bring all search files out of the home folder.

Yeah, that's a nice to have, but not visible to a user, so I leave that up to you.

zlayne:
@KevinFCormier I have added a [redirect](https://github.com/stolostron/console/pull/3663/files#diff-2f4a9251f09ce29560829ade50db3768935c283db33f0c19718e4abba93c9da1R20), and moved the search folder out of home in the file structure.
Let me know if these changes make sense to you, thanks.

PR #3681, opened 2024-07-21
ACM-12098: Helper function for export of SubRows and returning CSV safe strings

Randy424:
Related to ACM-12099: Here I'm adding some functionality to the export feature that will make exporting SubRows possible and will make exporting strings with comma's a seamless experience.

**_Merge this commit ahead of the  "Implementation" commits!_**


KevinFCormier (changes requested):
Just 1 inline comment.

Randy424:
@KevinFCormier I rebased on main, so another review is needed!

PR #3693, opened 2024-07-25
ACM-12797 - LaunchLogs uses default path for installer pod logs

Randy424:
Regarding: [ACM-12899](https://issues.redhat.com/browse/ACM-12899)

https://github.com/user-attachments/assets/ed099cca-aaee-4221-bb4a-de02d57d0c3d

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/HiveNotification.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/HiveNotification.tsx
@@ -107,11 +107,7 @@ export function launchLogs(cluster: Cluster, configMaps: ConfigMap[]) {
     const response = getHivePod(cluster.namespace, cluster.name, cluster.status)
     response.then((job) => {
       const podName = job?.metadata.name || ''
-      const containerName = podName.includes('uninstall') ? 'deprovision' : 'hive'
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/HiveNotification.tsx
```diff
@@ -107,11 +107,8 @@ export function launchLogs(cluster: Cluster, configMaps: ConfigMap[]) {
     const response = getHivePod(cluster.namespace, cluster.name, cluster.status)
     response.then((job) => {
       const podName = job?.metadata.name || ''
-      const containerName = podName.includes('uninstall') ? 'deprovision' : 'hive'
-      podName &&
-        window.open(
-          `${openShiftConsoleUrl}/k8s/ns/${cluster.namespace}/pods/${podName}/logs?container=${containerName}`
-        )
+      const containerName = podName.includes('uninstall') ? '?container=deprovision' : ''
```
KevinFCormier, line 110: Can you check if we can remove the container name for deprovision as well?

Randy424: It looks like the default for the uninstall pod also works.

PR #3771, opened 2024-08-16
ACM-13246 App topology display more data for VM apps

fxiang1:
https://issues.redhat.com/browse/ACM-13246

- Display almost the same data as ArgoCD console

vm_app_topo

2vm_app_topo

**Limitations**

- Not possible to match the PodDisruptionBudget resource to a node in the topology due to lack of data in the resource ie. labels or VM name in metadata.name
- Not possible to match the ControllerRevision for starting the VM due to lack of data in the resource ie. labels or VM name in metadata.name


frontend/src/routes/Applications/ApplicationDetails/ApplicationTopology/model/computeRelated.js
```diff
@@ -103,7 +106,22 @@ export const addDiagramDetails = (resourceStatuses, resourceMap, isClusterGroupe
         continue
       }
       //look for pod template hash and remove it from the name if there
-      const { nameNoHash, deployableName } = getNameWithoutPodHash(relatedKindList[i])
+      let { nameNoHash, deployableName } = getNameWithoutPodHash(relatedKindList[i])
+
```
chenz4027, line 110: Do we want to add a check to see if relatedKindList[i] exist?

fxiang1: Thanks @chenz4027 ! I don't think we need this check because this is part of a for loop. If `relatedKindList` is empty then it will just exit the loop.
image

PR #3773, opened 2024-08-19
ACM-13295-Not-able-to-undo-in-YAML-editor

jeswanke:
unfortunately using setYaml on the monaco editor has the unfortunate side effect of messing up the undo stack

to avoid that, monaco prefers adding text with an executeEdits()

we don't want to do that everytime the user types a character because it would mess up the stack even more

but since the bug is only that it doesn't do it with a full paste

so when a paste occurs, this change now looks at the pasted yaml and if it's missing the namespace it will add it

if we want to make this more generic, instead of a boolean (autoAddNs) we could make that a function that's called when a paste occurs which then fixes up the pasted yaml

### DIFF
--- a/frontend/__mocks__/react-monaco-editor.tsx
+++ b/frontend/__mocks__/react-monaco-editor.tsx
@@ -117,7 +117,10 @@ const MonacoEditor = (props: {
         past: ['past'],
       },
       forceTokenization: () => {},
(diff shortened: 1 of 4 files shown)


yiraeChristineKim:
This is great Thanks!

PR #3803, opened 2024-08-27
 ACM-13003 Fast cluster page load

jeswanke:


backend/src/lib/server-side-events.ts
```diff
@@ -274,25 +286,78 @@ export class ServerSideEvents {
...
+    // send packets of resources
+    const sending = start
+    do {
+      sending.push(...clusters.splice(0, 200))
+      sending.push(...agents.splice(0, 200))
+      sending.push(...infos.splice(0, 200))
+      sending.push(...policies.splice(0, 200))
+      sending.push(...addons.splice(0, 400))
+      sending.push({ id: '999999', data: { type: 'EOP' } }) // END OF PACKET
+    } while (clusters.length || policies.length || addons.length)
```
KevinFCormier, line 350: Missing a check for whether `agents` and `infos` are empty?

backend/src/lib/server-side-events.ts
```diff
@@ -274,25 +286,78 @@ export class ServerSideEvents {
...
+    do {
+      sending.push(...remainder.splice(0, 2000))
+      sending.push({ id: '999999', data: { type: 'EOP' } }) // END OF PACKET
+    } while (clusters.length || policies.length || addons.length)
```
KevinFCormier, line 354: Wrong condition

frontend/src/components/LoadData.tsx
```diff
@@ -237,8 +237,10 @@ export function LoadData(props: { children?: ReactNode }) {
   const setAgentMachinesState = useSetRecoilState(agentMachinesState)
   const setIsGlobalHub = useSetRecoilState(isGlobalHubState)
 
-  const { setters, caches } = useMemo(() => {
+  const { setters, mappers, caches } = useMemo(() => {
     const setters: Record<string, Record<string, SetterOrUpdater<any[]>>> = {}
+    const mappers: Record<string, Record<string, { setter: SetterOrUpdater<Map<string, any[]>>; keyBy: string[] }>> = {}
+    // const recorders: Record<string, Record<string, { setter: SetterOrUpdater<Map<string, any>>; keyBy: string[] }>> = {}
```
KevinFCormier, line 243: Was this abandoned, or an idea for another kind of atom?

backend/src/lib/server-side-events.ts
```diff
@@ -274,25 +286,78 @@ export class ServerSideEvents {
...
+          agents.push(event)
+          break
+        case 'ManagedClusterInfo':
+          infos.push(event)
+          break
+        case 'ManagedClusterAddOn':
+          addons.push(event)
+          break
+        case 'Application':
+          break
```
KevinFCormier, line 323: I think we should probably do this filtering elsewhere, so that we watch the resources so they can be used in the paginated API, but we do not create events for them. And it needs to be coordinated with removing the atoms from the frontend.

backend/src/lib/server-side-events.ts
```diff
@@ -274,25 +286,78 @@ export class ServerSideEvents {
...
     })
 
-    let lastEventID = 0
-    if (req.headers['last-event-id']) {
-      const last = Number(req.headers['last-event-id'])
-      if (Number.isInteger(last)) {
-        const cookies = parseCookies(req)
-        if (cookies['watch'] === instanceID) {
-          lastEventID = last
-        }
```
KevinFCormier, line 284: Need to do some research into this. I'm a bit concerned that by removing this we may be losing the ability for the event stream to reconnect. But it's hard to say how often that happens or how robust it is.

PR #3808, opened 2024-08-28
Update color values for Donut chart

zlayne:
Updating color values to pf5 values - https://docs.google.com/document/d/1sZubxAtA5Nf2dmXoK7febe-9EF_88Ztq3hxK7FymOgs

FYI @imjoyjean


KevinFCormier:
Ah, interesting. I has similar changes in progress here for PF5: https://github.com/stolostron/console/pull/3809/commits/df6366355c23bc35647344264c643d7378d659eb

But it looks like yours actually tweak some colors and change low to black instead of blue.

PR #3811, opened 2024-08-29
[ACM-10130] Enhanced fleet view summary section

zlayne:
Screenshot 2024-08-29 at 9 47 35 AM


KevinFCormier:
@zlayne Looks good to me. Please consider whether the console threat model needs an update due to new connections.

zlayne:
> @zlayne Looks good to me. Please consider whether the console threat model needs an update due to new connections.

Thanks for the review. I think the Kubernetes resources data flow would capture this, as we are using the default Prometheus Route exposed in `openshift-monitoring` namespace.

PR #3812, opened 2024-08-29
Fix the percentages for cluster compliance in the overview

mprahl:
This approach makes the AcmDonutChart more general like it used to be. This also fixes the plural issue of "1 Violations".

Relates:
https://issues.redhat.com/browse/ACM-13695


yiraeChristineKim (approved):

yiraeChristineKim:
so quick

PR #3820, opened 2024-09-03
Set the PolicyTemplateDetail card to be expanded by default

yiraeChristineKim:
Ref: https://issues.redhat.com/browse/ACM-13734

### DIFF
--- a/frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
+++ b/frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
@@ -21,7 +21,6 @@ export function PolicyTemplateDetails() {
   const urlParams = useParams()
   const kind = urlParams.kind ?? ''
   const { clusterName, template } = useTemplateDetailsContext()
-  const isCertPolicy = kind === 'CertificatePolicy'
   const [relatedObjects, setRelatedObjects] = useState<any>()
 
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
```diff
@@ -231,7 +230,7 @@ export function PolicyTemplateDetails() {
               id={'template-details-section'}
               title={kind + ' ' + t('details')}
               leftItems={descriptionItems}
-              defaultOpen={isCertPolicy}
+              defaultOpen
```
dhaiducek, line 233: Should this be set to `true` instead? I'm not seeing where this might be defined.

mprahl: I had to look this up too but this is magic syntax that is the equivalent of `defaultOpen={true}`.

mprahl (approved):

frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
yiraeChristineKim: Just define  `defaultOpen`  == true 
Sorry for confusing you

PR #3833, opened 2024-09-06
Allow upgrade and channel select for MOS only when ARO

KevinFCormier:
- Disable upgrade and channel select for OSD and ROSA Classic
- Update distribution version to show "ARO" for ARO

Related to https://issues.redhat.com/browse/ACM-10135

image


KevinFCormier:
Added a hold as I'm waiting for @berenss to confirm if the distribution name "ARO" is OK to use.

robdolares (approved):
/lgtm

berenss:
Yes, confirmed, LGTM. Please ship it. 🚀

PR #3837, opened 2024-09-09
ACM-14157-Fix-ACM-performance-issues

jeswanke:
Enhances performance by converting array filters and finds into a map get

With these changes, the gigantic pages are very peppy (press play):

CleanShot 2024-09-10 at 12 24 44

**The biggest change was for ManageClusterAddons need to be associated with their clusters**
So rather then store addons in a recoil list, they are now stored as a map:
   cluster-> addons[]

This is done in: 
   atoms.ts which introduces the mapper function
   LoadData.ts which stores addons in these cluster->addons

Then then rather then:
    addons = list.filter(clustername)
It's
    addons = map.get(clustername)

**Now that addons are maps, we don't need the flag I added to useAllClusters.tsx to ignore addons for applications**
useAllClusters(excludeUnclaimed?: boolean, excludeAddons?: boolean) -> useAllClusters(excludeUnclaimed?: boolean)
useAllClusters.tsx
ApplicationOverview.tsx
ApplicationDetails.tsx
Overview.tsx

**useClusterDistributionColumn problem**
Even though useAllClusters is fast, it's still cpu intensive, but useClusterDistributionColumn was calling it multiple times--fixed in:
ManagedClusters.tsx
ClusterSetManageResources.tsx
RemoveAutomationModal.tsx
UpdateAutomationModal.tsx

**Polish the cluster list wehn there are lots of clusters**
AcmInlineStatus.tsx straightens out the status in the cluster list
ManagedClusters.tsx # useClusterLabelsColumn collapses all labels if cluster list has more than 10 clusters:
CleanShot 2024-09-10 at 12 05 55@2x

**highlights text search in cluster list**
ManagedClusters.tsx:
CleanShot 2024-09-10 at 12 09 46@2x


KevinFCormier (changes requested):
John, could you please review the Sonar issues. I'm not concerned with the cognitive complexity one, but most of the others are valid and minor little things to clean up. Otherwise it looks good.

PR #3838, opened 2024-09-09
ACM-14118 - fix for application advanced configuration export value

Randy424:
Regarding: https://issues.redhat.com/browse/ACM-14118

Before: 
image-2024-09-06-19-52-49-198

After: 
Screenshot 2024-09-09 at 12 52 05 PM

### DIFF
--- a/frontend/src/routes/Applications/AdvancedConfiguration.tsx
+++ b/frontend/src/routes/Applications/AdvancedConfiguration.tsx
@@ -465,11 +465,13 @@ export default function AdvancedConfiguration(props: AdvancedConfigurationPagePr
             tooltip: t(
               'Displays the number of remote and local clusters where resources for the subscription are deployed. Click to search for all related clusters.'
             ),
-            exportContent: () => {
+            exportContent: (resource) => {
               const clusterCount = {
                 localPlacement: false,
                 remoteCount: 0,
               }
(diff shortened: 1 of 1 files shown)


frontend/src/routes/Applications/AdvancedConfiguration.tsx
```diff
@@ -465,11 +465,13 @@ export default function AdvancedConfiguration(props: AdvancedConfigurationPagePr
...
               'Displays the number of remote and local clusters where resources for the subscription are deployed. Click to search for all related clusters.'
             ),
-            exportContent: () => {
+            exportContent: (resource) => {
               const clusterCount = {
                 localPlacement: false,
                 remoteCount: 0,
               }
+              // mutates clusterCount object
+              getSubscriptionClusterCount(resource, clusterCount, true)
```
Randy424, line 474: `getSubscriptionClusterCount` is used in the above template to update the clusterCount object. In my first implementation, I had not realized the function had a mutative effect on it, so I've applied it here in the export callback, as to mirror it's affects, without using the link that it returns, as that is not necessary for the exporter.

KevinFCormier: Thanks for including the explanatory comment.

PR #3849, opened 2024-09-12
ACM-14056-application-overview-taking-long-time-to-show-flux

jeswanke:
the backend used take 5-10 minutes to refresh because we didn't overtax the search api getting apps for a telco environment which can be >100k apps

however this rate can be vastly improved by realizing:
1. that a lot of these apps are the system apps acm injects into each cluster
2. every cluster gets the exact same apps (3800x33=125k)

so this change treats the system apps differently:
1. system apps are refreshed separately and slower so that we're not wasting time on a bunch of static system apps, 
3.  and because we're spending less time on system apps, we can go back to searching argo/flux etc in 6 groups instead of one letter at a time
4. with 6 groups * 2 (argo pass, ocp pass) * 10s, every user app is now refreshed at a maximum of 2 minutes


backend/src/lib/clusters.ts
```diff
@@ -0,0 +1,19 @@
...
+  const managedClusters = getKubeResources('ManagedCluster', 'cluster.open-cluster-management.io/v1')
+  const clusterDeployments = getKubeResources('ClusterDeployment', 'hive.openshift.io/v1')
+  const hostedClusters = getKubeResources('HostedCluster', 'hypershift.openshift.io/v1beta1')
+  return [...managedClusters, ...clusterDeployments, ...hostedClusters].reduce((clusterMap, cluster) => {
+    if (cluster.metadata.name) {
+      clusterMap[cluster.metadata.name] = cluster
+    }
+    return clusterMap
+  }, {} as ClusterMapType)
+}
```
KevinFCormier, line 19: Only clusters that are under management have search running on them. I'm not sure why we would need to include `ClusterDeployment` or `HostedCluster` resources when assembling the list of cluster names. Is there something I'm missing?

backend/src/routes/aggregators/applicationsOCP.ts
```diff
@@ -125,8 +168,68 @@ export async function getOCPApps(applicationCache: ApplicationCacheType, argAppS
...
+
+function getNextClusterNameChuck(applicationCache: ApplicationCacheType): string[] {
+  // if no cluster name chucks left, create a new array of chunks
+  if (clusterNameChunks.length === 0) {
+    const clusterMap = getClusterMap()
+    const clusterNames = Object.keys(clusterMap)
+    if (clusterNames.length > 0) {
+      const chunks = clusterNames.reduce((chunks: string[][], chunkName, index) => {
+        const cindex = Math.floor(index / REMOTE_CLUSTER_CHUNKS)
+        chunks[cindex] = ([] as string[]).concat(chunks[cindex] || [], chunkName)
```
KevinFCormier, line 192: Two suggestions here:

1. Variable naming. A chunk appears to be an array of cluster names that itself does not have a name. I think a better name for `chunkName` would be `clusterName`.
2. This line is overly convoluted. Would one of the following work? They are concise and avoid the TypeScript cast.

```suggestion
        chunks[cindex] = [...(chunks[cindex] ?? []), clusterName]
        chunks[cindex] = (chunks[cindex] ?? []).concat(clusterName)
```

backend/test/routes/aggregator.test.ts
```diff
@@ -73,7 +73,7 @@ describe(`aggregator Route`, function () {
 
     // fill in application cache from resourceCache and search api mocks
     aggregateKubeApplications()
-    await aggregatSearchAPIApplications(10)
+    await aggregatSearchAPIApplications(1)
```
KevinFCormier, line 76: Can we please fix the typo in the name of this function?

KevinFCormier (changes requested):
I tested this out and it works great. I see application updates almost immediately.

I did leave several comments on code maintainability and readability, though. Please take a look. Thanks.

PR #3868, opened 2024-09-17
Do not set BRIDGE_K8S_AUTH_BEARER_TOKEN for OCP 4.16+

KevinFCormier:
This environment variable only works with 4.15 and earlier.


fxiang1 (approved):
Thanks Kevin!

PR #3870, opened 2024-09-17
[ACM-13485] Enabled VirtualMachine actions from search results

zlayne:
PR adds ability to start, stop, restart, pause, and unpause VirtualMachines. These actions are hidden behind the `VIRTUAL_MACHINE_ACTIONS` feature flag. 

Screenshot 2024-09-17 at 2 13 25 PM


zlayne:
FYI @jlpadilla

backend/src/lib/json-request.ts
```diff
@@ -50,3 +58,39 @@ export function jsonPost<T = unknown>(
...
+
+export function jsonPut(url: string, body: unknown, token?: string): Promise<PutResponse> {
+  const headers: HeadersInit = {
+    [HTTP2_HEADER_CONTENT_TYPE]: 'application/json',
+  }
+  if (token) headers[HTTP2_HEADER_AUTHORIZATION] = `Bearer ${token}`
+  return fetchRetry(url, {
+    method: 'PUT',
+    headers,
+    agent: new Agent({ rejectUnauthorized: false }),
```
KevinFCormier, line 70: We should not be ignoring SSL errors by default. If this is needed, it needs to be captured in the threat model and documented as a security risk.

zlayne: I updated the agent to `agent: new Agent({ ca: getCACertificate() })` so the rejectUnauthorized will default to true, and the ca-cert will be overridden.

PR #3892, opened 2024-09-23
VirtualMachine edit & delete auth

zlayne:
Map VirtualMachine edit & delete auth to whether or not a user can create a ManagedClusterAction

### DIFF
--- a/frontend/src/routes/Search/SearchResults/utils.tsx
+++ b/frontend/src/routes/Search/SearchResults/utils.tsx
@@ -5,8 +5,15 @@ import { useContext, useMemo } from 'react'
 import { TFunction } from 'react-i18next'
 import { generatePath, NavigateFunction, useNavigate } from 'react-router-dom-v5-compat'
 import { useTranslation } from '../../../lib/acm-i18next'
+import { canUser } from '../../../lib/rbac-util'
 import { NavigationPath } from '../../../NavigationPath'
-import { Cluster, getBackendUrl, putRequest } from '../../../resources'
+import {
+  Cluster,
+  getBackendUrl,
+  ManagedClusterActionApiVersion,
+  ManagedClusterActionKind,
+  putRequest,
+} from '../../../resources'
 import { useRecoilValue, useSharedAtoms } from '../../../shared-recoil'
 import { AcmToastContext, compareStrings, IAlertContext } from '../../../ui-components'
 import { useAllClusters } from '../../Infrastructure/Clusters/ManagedClusters/components/useAllClusters'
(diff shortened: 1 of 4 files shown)


frontend/src/routes/Search/SearchResults/utils.tsx
```diff
@@ -27,6 +34,42 @@ export interface ISearchResult {
...
+          metadata: {
+            name,
+            namespace,
+          },
+        },
+        cluster === 'local-cluster' ? namespace : cluster,
+        name
+      )
+  }
+}
```
KevinFCormier, line 71: Shouldn't this just be based on whether the resource is on the local-cluster or a managed cluster?

zlayne: @jlpadilla Correct me if i'm wrong here, but we don't want to provide explicit auth on VM/VMI resources. To prevent users from creating resources on hub or elsewhere they shouldn't be able to?

jlpadilla: I believe that @KevinFCormier point is that the ACM hub cluster could also have the Openshift Virtualization operator. So there's 2 scenarios:
- **VM/VMI object is in the ACM Hub cluster.** 
  For this scenario we don't use the ManagedServiceAccount. For RBAC we check directly with the user account. `oc auth can-i put virtualmachine/start -n vmnamespace`
- **VM/VMI object is in a Managed cluster.**
  For this scenario we use the ManagedServiceAccount and map RBAC to ManagedClusterAction.

The logic will be akin to what we do for other resources like Pod.

KevinFCormier: My point was that I thought this code was used for all resources from search results, not just VirtualMachines. So for any resource that's not on the local-cluster, wouldn't we be taking the action via ManagedClusterAction?

But I have just skimmed, so I could be wrong here.

jlpadilla: @zlayne I see where's the confusion. **Edit & Delete** should work the same as all other resources. I don't expect any changes for these 2 actions.

The changes should be for the new actions [start, stop, restart, pause, unpause].

zlayne: @jlpadilla Oh my bad on the confusion here. For the new VM actions I handle auth in the backend [here](https://github.com/stolostron/console/blob/main/backend/src/routes/virtualMachineProxy.ts#L42). I am not currently disabling the action item in the UI, as the UI only authenticates in user token.

zlayne:
Closing PR based on included discussion.

PR #3894, opened 2024-09-24
ACM-13728: Missing s390x arch option in Cpu architecture in InfraEnv

jgyselov:
https://issues.redhat.com/browse/ACM-13728


jgyselov:
Included in https://github.com/stolostron/console/pull/3963 and https://github.com/stolostron/console/pull/3964.

PR #3895, opened 2024-09-24
ACM-14420 prevent mapped addon from throwing exception

jeswanke:
used to map managed cluster addon using a Map which occasionally messes up recoil
converted to Record and using safer ?. to access it


frontend/src/components/LoadData.tsx
```diff
@@ -426,7 +435,8 @@ export function LoadData(props: { children?: ReactNode }) {
                     }, [] as string[])
                     .join('/')
                   setter((map) => {
-                    const arr = map.get(key) || []
+                    const newMap = structuredClone(map)
```
KevinFCormier, line 438: This is good because you are not supposed to mutate the atoms, but I think you can still reuse the contents, so a full clone is probably not necessary.
```suggestion
                    const newMap = {...map}
```

jeswanke: ha--I tried {...map} first -- but it does a shallow clone--each map has an immutable array that also has to be cloned :(

KevinFCormier: Ah, so recoil does lock everything inside the state as well. Very well.

PR #3896, opened 2024-09-24
[ACM-14444] Infrastructure virtual machine page

zlayne:
Screenshot 2024-09-24 at 7 42 08 AM


zlayne:
@jlpadilla FYI I know we face the scale issue with thousands of VMs. But I have the search result limit [set to -1](https://github.com/stolostron/console/pull/3896/files#diff-ca7989df25bcb2504c60648fc8c3389698176e8875ddc4fdeff23c06e6174002R66) to return all resources. I am doing this because I need to combine the VM and its instance resource (node & IP address) into a single object.

PR #3897, opened 2024-09-24
ACM-13644-Implement AcmSearchInput (advanced search input)

Randy424:
Regarding: [ACM-13644](https://issues.redhat.com/browse/ACM-13644)

Before:
Screenshot 2024-09-24 at 1 28 02 PM

After:
Screenshot 2024-09-24 at 1 24 53 PM


Randy424:
/hold
code smells

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -678,7 +705,7 @@ export function useClusterNamespaceColumn(): IAcmTableColumn<Cluster> {
     ),
     sort: 'namespace',
     cell: (cluster, search) => {
-      return <HighlightSearchText text={cluster.namespace || '-'} searchText={search} />
+      return <HighlightSearchText text={cluster.namespace ?? '-'} searchText={search} />
```
Randy424, line 708: Code smell fix. Not directly related to my feature.

frontend/public/locales/en/translation.json
```diff
@@ -1460,6 +1463,7 @@
   "GPU Accelerated Compute": "GPU Accelerated Compute",
   "Guest": "Guest",
   "H1 - 16 TB of HDD-based local storage": "H1 - 16 TB of HDD-based local storage",
+  "Has the words": "Has the words",
```
KevinFCormier, line 1466: I think we should call this "Fuzzy search" or something similar, as that will be more accurate. @berenss and @imjoyjean any thoughts on terminology there?

berenss: I'm good with whatever decision Joy and Kevin make here!

frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
```diff
@@ -0,0 +1,273 @@
...
+      }
+      isAdvancedSearchOpen={isAdvancedSearchOpen}
+      onClear={onClear}
+      value={fuzzySearchValue}
+      enableUtilities={showResultCount}
+    />
+  )
+
+  const advancedSearchForm = (
+    <div ref={advancedSearchPaneRef} aria-label="Advanced search form">
```
KevinFCormier, line 156: aria-label should be translated, although, I'm not sure that it is usual to have one on a div.

frontend/src/lib/search-utils.ts
```diff
@@ -0,0 +1,5 @@
+/* Copyright Contributors to the Open Cluster Management project */
+
+export const handleOperatorComparison = (value: string, selectedValue: string) => {
```
KevinFCormier, line 3: Based on the name, I guess this will be updated later to also accept an operator.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -479,6 +482,31 @@ export function ClustersTable(props: {
...
 
+  const advancedFilters = useMemo<ITableAdvancedFilter<Cluster>[]>(() => {
+    return [
+      {
+        id: 'name',
+        columnDisplayName: 'Name',
+        label: t('table.name'),
+        availableOperators: [SearchOperator.Equals],
+        tableFilterFn: (selectedValues, cluster) => {
+          const filtervalue = selectedValues[0]
```
KevinFCormier, line 493: I don't love that we're assuming there's 1 entry in the array. I mentioned elsewhere this may be a consequence of trying to re-use the existing filter callback type, so I'm assuming you'll address this in iteration 2.

frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
```diff
@@ -0,0 +1,273 @@
...
+                    event.stopPropagation()
+                    searchConstraints &&
+                      setSearchConstraints([
+                        ...searchConstraints,
+                        { operator: SearchOperator.Equals, value: '', columnId: '' },
+                      ])
+                  }}
+                  icon={<PlusCircleIcon />}
+                >
+                  {t('Add a search constraint').toString()}
```
KevinFCormier, line 237:
```suggestion
                  {t('Add a search constraint')}
```
Good idea to always test with `npm run plugins`. It finds more problems in general and more problems that occur when running as a plugin.

You imported `t` from `i18next`; need to use the `useTranslation` hook.

image

frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
```diff
@@ -0,0 +1,273 @@
...
+              <ActionGroup>
+                <Button
+                  variant="primary"
+                  type="submit"
+                  onClick={(event) => {
+                    event.preventDefault()
+                    setIsAdvancedSearchOpen(false)
+                  }}
+                >
+                  Close
```
KevinFCormier, line 249: Translate, please

frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
```diff
@@ -0,0 +1,273 @@
...
+                </Button>
+                {!!onClear && (
+                  <Button variant="link" type="reset" onClick={onClear}>
+                    Reset
```
KevinFCormier, line 253: Translate, please

frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
```diff
@@ -0,0 +1,273 @@
...
+    style,
+  } = props
+
+  const searchInputRef = useRef<HTMLInputElement | null>(null)
+  const advancedSearchPaneRef = useRef<HTMLDivElement | null>(null)
+  const [isAdvancedSearchOpen, setIsAdvancedSearchOpen] = useState(false)
+  const [searchConstraints, setSearchConstraints] = useState<SearchConstraint[]>(pendingConstraints || [])
+
+  const onClear = () => {
+    fuzzySearchOnClear && fuzzySearchOnClear()
```
KevinFCormier, line 79: A bit more concise:
```suggestion
    fuzzySearchOnClear?.()
```
It might also be good practice to wrap these event handlers with `useCallback`... but it depends on whether anything in the components you are passing these to will get rerendered when they change.

KevinFCormier (changes requested):
Really cool overall. I left some suggestions on how to clean things up a bit here and there and make things a little more efficient. 

I found 2 bugs:
- Issue with translation that I left a comment about
- Style issue for the popover. There is a border showing on the right and bottom edges

image

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
Randy424: That's correct actually, this filterFn callback changes in the subsequent iteration such that we are not assuming here...it was also bugging me haha.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -479,6 +482,31 @@ export function ClustersTable(props: {
 
   const rowActions = useMemo(() => [], [])
 
+  const advancedFilters = useMemo<ITableAdvancedFilter<Cluster>[]>(() => {
+    return [
+      {
+        id: 'name',
+        columnDisplayName: 'Name',
```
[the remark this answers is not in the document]
Randy424: Hah! Tremendous oversight there...

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -236,6 +237,10 @@ export interface ITableFilter<T> {
   tableFilterFn: TableFilterFn<T>
   showEmptyOptions?: boolean
 }
+export interface ITableAdvancedFilter<T> extends Omit<ITableFilter<T>, 'options'> {
```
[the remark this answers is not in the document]
Randy424: That's correct, it gets a face lift in iteration two.

frontend/src/lib/search-utils.ts
Randy424: Yes, it had an operator param and switch in my first draft, but I removed it. Serves no purpose with only one supported operator. This is expanded in iteration 2.

frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
Randy424: Had to take a moment to think about this, but yes, I suppose since they both have some control over the constraint array or the fuzzy search this makes sense...

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -787,6 +794,33 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+          filterCategories.every((filter: string) => {
+            const filterItem: ITableAdvancedFilter<T> | undefined = findFilterMatch(filter)
+            /* istanbul ignore next */
+            const isMatch = filterItem?.tableFilterFn(advancedFilterSelections[filter], item) ?? true
+            if (!isMatch) {
+              isFilterMatch = false
+            }
+          })
+          return isFilterMatch
+        })
```
KevinFCormier, line 821: @Randy424 This is not quite right. In order for `Array.every()` to short-circuit, you need to return a boolean from the predicate, rather than setting `isFilterMatch` as a side-effect.

Randy424: ack, I confused the callback boundary.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -787,6 +794,34 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+            const filterItem: ITableAdvancedFilter<T> | undefined = findFilterMatch(filter)
+            /* istanbul ignore next */
+            const isMatch = filterItem?.tableFilterFn(advancedFilterSelections[filter], item) ?? true
+            if (!isMatch) {
+              isFilterMatch = false
+            }
+            return isMatch
+          })
+          return isFilterMatch
+        })
```
KevinFCormier, line 822: Better, but we can still avoid having to use `let` (by actually using the return value of `Array.everey()`) and also declaring the type for `filterItem`.

To test, I set `name` and `namespace` to the same value, and confirmed we hit the filtering function for both. Then I changed `name` to something that does not match any cluster and confirmed we never bothered to check `namespace`.

I notice this was copied from the other filter code, so perhaps we can make this efficiency adjustment there as well, but could be as a separate PR.

```suggestion
        filteredItems = filteredItems.filter((item: T) =>
          filterCategories.every((filter: string) => 
            /* istanbul ignore next */
            findFilterMatch(filter)?.tableFilterFn(advancedFilterSelections[filter], item) ?? true
          )
        )
```

Randy424: I see. I will add that note to my remaining work for iteration 2...

PR #3905, opened 2024-09-26
VirtualMachine actions rbac

zlayne:
Console backend should check if the user is authorized to create ManagedClusterActions in the managed cluster namespace to allow the user to take action on Virtual Machines.

### DIFF
--- a/frontend/src/routes/Search/SearchResults/utils.tsx
+++ b/frontend/src/routes/Search/SearchResults/utils.tsx
@@ -55,8 +55,8 @@ export function handleVMActions(
     .catch((err) => {
       console.error(`VirtualMachine: ${item.name} ${action} error. ${err}`)
 
-      let errMessage = err?.message ?? t('An unexpected error occurred.')
-      if (errMessage.includes(':')) errMessage = errMessage.split(':')[1]
+      let errMessage: string = err?.message ?? t('An unexpected error occurred.')
+      if (errMessage.includes(':')) errMessage = errMessage.split(':').slice(1).join(':')
       if (errMessage === 'Unauthorized') errMessage = t('Unauthorized to execute this action.')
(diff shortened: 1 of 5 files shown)


frontend/src/routes/Search/SearchResults/utils.tsx
```diff
@@ -267,9 +267,12 @@ export function getRowActions(
     id: 'startVM',
     title: t('Start {{resourceKind}}', { resourceKind }),
     click: (item: any) => {
+      const path = item?._hubClusterResource
+        ? `/apis/subresources.kubevirt.io/v1/namespaces/${item.namespace}/virtualmachines/${item.name}/start`
+        : `/virtualmachines/start`
```
jlpadilla, line 272: I'm thinking this could be simplified and also made more generic if we did something like this.

```
const path = item?._hubClusterResource
        ? `/apis/subresources.kubevirt.io/v1/namespaces/${item.namespace}/virtualmachines/${item.name}/start`
        : `/proxytocluster?path=/apis/subresources.kubevirt.io/v1/namespaces/${item.namespace}/virtualmachines/${item.name}/start`
```

Then on the backend you don't need to rebuild the path.
It also makes the backend closer to being generic.

**Maybe this change should be separate from this PR.**

zlayne: Yes, this would be nice, would require some updates to how we get the cluster proxy route and token as well. Certainly work for another PR in my opinion.

jlpadilla: Thanks Zack! We are at the finish line, so I'm good if we push this refactor to 2.13. I just want to make sure we establish the correct pattern.

robdolares (approved):
/lgtm

PR #3906, opened 2024-09-26
ACM-14531 RHOCM Import credential in Discovered Clusters fix 

vishsanghishetty:
Updates the updateROSAImportSecret to copy the correct OCM credential either `api-token` or `service-account`

### DIFF
--- a/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.test.tsx
+++ b/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.test.tsx
@@ -33,7 +33,7 @@ import {
   SubscriptionOperatorApiVersion,
   SubscriptionOperatorKind,
 } from '../../../../../resources'
-import { render, waitFor } from '@testing-library/react'
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react'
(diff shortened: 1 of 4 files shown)


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.test.tsx
```diff
@@ -919,29 +1069,73 @@ describe('Import Discovered Cluster', () => {
...
+        </MemoryRouter>
+      </RecoilRoot>
+    )
+
+    const { getAllByText, getAllByLabelText, getByDisplayValue } = render(<Component />) // Render the custom component
 
+    await waitFor(() => {
+      expect(getAllByText(mockDiscoveredClusters[2].metadata.name!)[0]!).toBeInTheDocument()
+    })
+    userEvent.click(getAllByLabelText('Actions')[1]) // Click on Kebab menu
```
KevinFCormier, line 1108: I inserted a `screen.logTestingPlaygroundURL()` just before this line to debug. From the link, I could see the list of discovered clusters. The cluster you added is the 2nd one called "rosa-discovery-cluster", so you want index 2 instead of 1 out of all the kebab menus labeled "Actions".

image

That will get you to the next issue, which is with the nocking.

PR #3913, opened 2024-09-28
ACM-14557-Implement inferred ranges for cluster list advanced search

Randy424:
Regarding: [ACM-14557](https://issues.redhat.com/browse/ACM-14557)

When input in the Advanced Search Field is incomplete the intuitive behavior (what is being implemented here) is that an assumed range of values could appear in the search results.

For example: "Distribution version = 4.", Here the table should display all values **greater than or equal** to 4.0.0 **but less than** 5.0.0

Before: 
Screenshot 2024-09-27 at 10 19 28 PM

After: 
Screenshot 2024-09-27 at 10 25 19 PM

### DIFF
--- a/frontend/src/lib/search-utils.test.ts
+++ b/frontend/src/lib/search-utils.test.ts
@@ -17,6 +17,14 @@ const versionData = {
   versionIncomplete: '1.',
 }
 
+const incompleteVersionData = {
+  versionOneIncomplete: '1.',
+  versionTwoIncomplete: '1.5',
+  versionThreeIncomplete: '2',
+  versionFourIncomplete: '2.5',
+  versionFiveIncomplete: '2.5.',
+}
+
 describe('search-utils basic string operator', () => {
   const { stringOne, stringTwo, stringThree, stringFour, stringFive } = stringData
   it('can determine equals', () => {
@@ -43,10 +51,18 @@ describe('search-utils basic string operator', () => {
 
 describe('search-utils sermver operator', () => {
   const { versionOne, versionTwo, versionThree, versionFour, versionIncomplete } = versionData
-  it('can determine greater semver', () => {
+  const {
+    versionOneIncomplete,
+    versionTwoIncomplete,
+    versionThreeIncomplete,
+    versionFourIncomplete,
+    versionFiveIncomplete,
+  } = incompleteVersionData
+
+  it('can determine greater than semver', () => {
     expect(handleSemverOperatorComparison(versionFour, versionThree, SearchOperator.GreaterThan)).toBeTruthy()
   })
-  it('can determine less semver', () => {
+  it('can determine less than semver', () => {
     expect(handleSemverOperatorComparison(versionOne, versionTwo, SearchOperator.LessThan)).toBeTruthy()
   })
   it('can determine equals semver', () => {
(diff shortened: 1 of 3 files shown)


Randy424:
/hold 

Seeking feedback

frontend/src/lib/search-utils.ts
```diff
@@ -22,32 +22,61 @@ export const handleStandardComparison = (valueOne: string, valueTwo: string, ope
...
+  // If the inputVersion is not a valid semver, but the coerced version is, we can use the incomplete version to create a range
+  if (!semver.valid(inputVersion) && semver.valid(semver.coerce(inputVersion))) {
+    const isMajorVersionAlone = !inputVersion.includes(
+      `${semver.coerce(inputVersion)?.major}.${semver.coerce(inputVersion)?.minor}`
+    )
+    const topRange = isMajorVersionAlone
+      ? semver.inc(coercedInputVersion, 'major')
+      : semver.inc(coercedInputVersion, 'minor')
+
+    const equalsValueRange = `${coercedInputVersion} - ${topRange}`
```
KevinFCormier, line 43: I think this would include `topRange`, but it should be excluded.

I think we can just use "X" or "X.Y" as a range. See [X-Ranges 1.2.x 1.X 1.2.* *](https://github.com/npm/node-semver?tab=readme-ov-file#x-ranges-12x-1x-12-): "A partial version range is treated as an X-Range, so the special character is in fact optional."

```
semver.satisfies("5.0.0", "4.0.0 - 5.0.0")
true
semver.satisfies("5.0.0", "4")
false
semver.satisfies("4.14.50", "4.14")
true
semver.satisfies("4.17.3", "4")
true
```

frontend/src/lib/search-utils.ts
```diff
@@ -22,32 +22,61 @@ export const handleStandardComparison = (valueOne: string, valueTwo: string, ope
...
+      case SearchOperator.Equals:
+        return semver.satisfies(coercedVersionToCompare, equalsValueRange)
+      case SearchOperator.GreaterThan:
+        return semver.satisfies(coercedVersionToCompare, `>${coercedInputVersion}`)
+      case SearchOperator.GreaterThanOrEqualTo:
+        return semver.satisfies(coercedVersionToCompare, `>=${coercedInputVersion}`)
+      case SearchOperator.LessThan:
+        return semver.satisfies(coercedVersionToCompare, `<${coercedInputVersion}`)
+      case SearchOperator.LessThanOrEqualTo:
+        return semver.satisfies(coercedVersionToCompare, `<=${coercedInputVersion}`)
```
KevinFCormier, line 55: Is there a reason we are changing these ones to use `satisfies` if they are not using a range?

Also, I think we should refactor this code to avoid repeating the switch statement.

frontend/src/lib/search-utils.ts
```diff
@@ -22,32 +22,61 @@ export const handleStandardComparison = (valueOne: string, valueTwo: string, ope
...
-export const handleSemverOperatorComparison = (versionOne: string, versionTwo: string, operator: SearchOperator) => {
-  // Semver coerces the version to a valid semver version if possible, otherwise it returns the original value
-  const coercedVersionOne = semver.valid(semver.coerce(versionOne)) ?? versionOne
-  const coercedVersionTwo = semver.valid(semver.coerce(versionTwo)) ?? versionTwo
+export const handleSemverOperatorComparison = (
+  versionToCompare: string,
+  inputVersion: string,
+  operator: SearchOperator
+) => {
+  const coerceVersion = (version: string) => semver.valid(semver.coerce(version)) ?? version
```
KevinFCormier, line 30: Does using `semver.valid` accomplish anything here? I think `semver.coerce` returns either a valid SemVer or null. Perhaps we could just do `const coercedInputVersion = semver.coerce(inputVersion)` and then just check for a value in the later conditions, rather than checking the result of `semver.coerce` again.

KevinFCormier (commented):
Some initial feedback

frontend/src/lib/search-utils.ts
Randy424: If I'm understanding you correctly, I believe using `satisfies` is still necessary in this situation, as this language (e.g. `>${coercedInputVersion}`)  is actually part of the [range grammar](https://github.com/npm/node-semver?tab=readme-ov-file#range-grammar), and would not work with something like `semver.gt`.

Otherwise, big copy-that on the refactor.

KevinFCormier: But `coercedInputVersion` itself is not a range. It's only a range because we added the operator before it. I think it's a valid question of whether we want to use a range in this case or not though.

To test this, in your file I added `window.semver = semver` so I could easily call semver functions in the browser console.

Suppose a user has entered "version>4.12"

semver.gt("4.12.1", "4.12.0") is true
semver.satisfies("4.12.1" ">4.12") is false

To me the latter makes sense because the user did not specify the patch version, so they are looking for 4.13 or higher.

I opened a PR against your PR with simplified implementation.
https://github.com/Randy424/console/pull/20

frontend/src/lib/search-utils.test.ts
```diff
@@ -66,4 +82,34 @@ describe('search-utils sermver operator', () => {
   it('can coerce incomplete semver strings', () => {
     expect(handleSemverOperatorComparison(versionIncomplete, versionOne, SearchOperator.Equals)).toBeTruthy()
   })
+  it('can infer on incomplete semver equals', () => {
+    expect(handleSemverOperatorComparison(versionOne, versionOneIncomplete, SearchOperator.Equals)).toBeTruthy()
+  })
+  it('can infer on incomplete semver greater than', () => {
```
KevinFCormier, line 88: This test fails with my proposed PR, which makes sense if we consider >1 to mean 2 or higher. (1.5.0 is the first arg here)

By the way, using constants for the test values here just makes it hard to understand and debug the tests. I would just repeat the numbers in the tests.

frontend/src/lib/search-utils.ts
KevinFCormier: Also, this matches how npm does version ranges. See https://semver.npmjs.com/

Enter "lodash" and ">1" and you'll see the lowest version is 2.0.0.

PR #3916, opened 2024-09-30
Fix engine svg align issue

yiraeChristineKim:
Ref: https://issues.redhat.com/browse/ACM-14573

### DIFF
--- a/frontend/src/routes/Governance/common/util.tsx
+++ b/frontend/src/routes/Governance/common/util.tsx
@@ -732,19 +732,22 @@ export function getEngineWithSvg(apiGroup: string): JSX.Element {
   }
 
   return (
-    <div style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
+    <>
       <div
         style={{
+          display: 'inline-block',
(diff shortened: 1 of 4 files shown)


mprahl (approved):

frontend/src/routes/Governance/common/util.tsx
```diff
@@ -732,19 +732,22 @@ export function getEngineWithSvg(apiGroup: string): JSX.Element {
   }
 
   return (
-    <div style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
+    <>
```
dhaiducek, line 735: I didn't know about this--looks like unnamed HTML tags are a thing now? What's the benefit? Is it an alias for `div` or something?

mprahl: @dhaiducek in React, it's an alias for `React.Fragment` which is a JSX.Element but does not impact the DOM as far as I know. It's just a way to return multiple HTLM elements as a single JSX.Element.

PR #3918, opened 2024-09-30
ACM-14576 Fix duplicate input ID

KevinFCormier:
OCM credential authentication method is appearing multiple times on the form due to a duplicate ID.


vishsanghishetty (approved):
thanks for fixing this @KevinFCormier 

/lgtm

PR #3919, opened 2024-10-01
ACM-13678 - add availability options step for kubevirt

robdolares:
related issue: https://issues.redhat.com/browse/ACM-13678


robdolares:
@KevinFCormier - could you take a look when you get a moment.  The issue I am having is that the yaml is not being updated initially or immediately after making a change to the options.  After making change to another form field outside of the availability options step, they will update.  So in the end at least we are seeing what we want, just have to iron out the state issues:
<kbd>
Screenshot 2024-10-24 at 11 25 46 AM
</kbd>

KevinFCormier:
> @KevinFCormier - could you take a look when you get a moment. The issue I am having is that the yaml is not being updated initially or immediately after making a change to the options. After making change to another form field outside of the availability options step, they will update. So in the end at least we are seeing what we want, just have to iron out the state issues: Screenshot 2024-10-24 at 11 25 46 AM

You were using the TimeWindow component as a reference implementation, but you missed a key part of it. See line 513.
https://github.com/stolostron/console/blob/main/frontend/src/routes/Applications/CreateApplication/Subscription/common/TimeWindow/index.js#L484-L515

You might notice in the control data that when we use a custom component, we do not provide any properties for it:
```
{
      id: 'availabilityOptions',
      type: 'custom',
      component: <AvailabilityOptionsForm />,
      active: []
}
```

So how does your component get the `control` property? Temptifly doesn't use the component you provided directly, but it clones it and provides several props. See https://github.com/stolostron/console/blob/main/frontend/src/components/TemplateEditor/controls/ControlPanel.js#L573-L606

One of these props is a `handleChange` function which you can call to tells temptifly to update the YAML editor.

I will leave additional comments inline.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/assisted-installer/AvailiabilityOptionsForm.tsx
KevinFCormier, line 1: Please correct the spelling error in the name of this file, and move it to an appropriate location. Although the Handlebars template file is under an `assisted-installer` directory, this flow has nothing to do with Assisted Installer.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/assisted-installer/AvailiabilityOptionsForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const { control } = props
+  const { t } = useTranslation()
+
+  // Set default option to "Highly Available"
+  if (_.isEmpty(control.active)) {
+    props.control.active = {
+      controller: 'HighlyAvailable',
+      infra: 'HighlyAvailable',
+    }
+  }
```
KevinFCormier, line 18: You could set this in the control data instead of setting `active: []`.

KevinFCormier (changes requested):
A couple minor inline comments.

After playing with this, I think I am more inclined to include the custom component on the first step, with *Cluster details*. Perhaps just before the *Additional labels* field. I think this is a setting most users will just leave as the default, so adding an extra step feels like overkill.

The alignment is off between the labels and the radio buttons though, and this will be more apparent if these controls are moved to be on the same step as other controls. I noticed though that we do have a control that uses radio buttons (the 'boolean' control) which has better alignment:

image
image

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/AvailabilityOptionsForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+export default AvailabilityOptionsForm
+
+export const summary = (control: any, t: TFunction) => {
+  const { controller, infra } = control.active || {}
+
+  return [
+    { term: t('Controller availability policy'), desc: controller },
+    { term: t('Infrastructure availability policy'), desc: infra },
```
KevinFCormier, line 130: I think the `desc` should be mapped to the translated display value here.

frontend/public/locales/en/translation.json
```diff
@@ -2384,6 +2386,8 @@
   "Single cluster": "Single cluster",
   "Single node OpenShift": "Single node OpenShift",
   "Single replica": "Single replica",
+  "Single Replica": "Single Replica",
```
KevinFCormier, line 2389: Duplicate casing

frontend/public/locales/en/translation.json
```diff
@@ -1455,6 +1455,8 @@
   "High Memory - Metal": "High Memory - Metal",
   "High Performance Compute": "High Performance Compute",
   "Highly available": "Highly available",
+  "Highly Available": "Highly Available",
```
KevinFCormier, line 1458: Duplicate casing

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/AvailabilityOptionsForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+    </AcmForm>
+  )
+}
+
+export default AvailabilityOptionsForm
+
+export const summary = (control: any, t: TFunction) => {
+  const { controller, infra } = control.active || {}
+  const controlDesc: string = controller === 'HighlyAvailable' ? t('Highly Available') : t('Single Replica')
+  const infraDesc: string = infra === 'HighlyAvailable' ? t('Highly Available') : t('Single Replica')
```
KevinFCormier, line 132: Exact same code repeated twice (and casing issue). Please extract to helper function.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/AvailabilityOptionsForm.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        </MemoryRouter>
+      </RecoilRoot>
+    )
+  }
+
+  test('renders', async () => {
+    const { container } = render(<Component />)
+    await waitForText('Controller availability policy')
+    expect(container).toMatchSnapshot()
+  })
```
KevinFCormier, line 44: I expect this is not needed for coverage (since the next test renders the same thing), so I don't think we need a snapshot test here. Can we remove?

KevinFCormier (changes requested):
Just a bit of cleanup needed. I will test out as well, but the code looks good otherwise.

PR #3920, opened 2024-10-01
ACM- 13528-Application-list-'System'-can-vary

jeswanke:

### DIFF
--- a/backend/src/routes/aggregators/applicationsOCP.ts
+++ b/backend/src/routes/aggregators/applicationsOCP.ts
@@ -1,7 +1,6 @@
 /* Copyright Contributors to the Open Cluster Management project */
 import { getClusterMap } from '../../lib/clusters'
 import { getMultiClusterEngine } from '../../lib/multi-cluster-engine'
-import { getMultiClusterHub } from '../../lib/multi-cluster-hub'
 import { getPagedSearchResources } from '../../lib/search'
 import { IResource } from '../../resources/resource'
 import { getKubeResources } from '../events'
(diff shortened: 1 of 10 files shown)


KevinFCormier:
@jeswanke I tested this out and it works well, but it looks like we should keep `open-cluster-management` in the list of prefixes and add the RHACM installation namespace as well, because some components have a static installation namespace.
image
image

backend/src/routes/aggregators/applicationsOCP.ts
```diff
@@ -265,12 +266,19 @@ function getValues(labels: { annotation: any; value: any }[]) {
   }
 }
 
+export const systemAppNamespacePrefixes: string[] = []
+export async function discoverSystemAppNamespacePrefixes() {
+  if (!systemAppNamespacePrefixes.length) {
+    systemAppNamespacePrefixes.push('openshift')
+    systemAppNamespacePrefixes.push('hive')
+    systemAppNamespacePrefixes.push('rhacm')
+    systemAppNamespacePrefixes.push('open-cluster-management')
```
KevinFCormier, line 275: Sorry John - by "RHACM installation namespace" I still meant you should look up what that actually is... not the hardcoded `rhacm` namespace that I used as an example.
```suggestion
    systemAppNamespacePrefixes.push('open-cluster-management')
    const mch = await getMultiClusterHub()
    if (mch?.metadata?.namespace && mch.metadata.namespace !== 'open-cluster-management') {
      systemAppNamespacePrefixes.push(mch.metadata.namespace )
    }
```

jeswanke: I got to remember to see if there's a suggested change 🥇

PR #3926, opened 2024-10-03
Make the PolicySet donut chart consistent

mprahl:
Use a number like other policy related donut charts rather than a percentage.

Relates:
https://issues.redhat.com/browse/ACM-14661

image


yiraeChristineKim:
Great!!

PR #3929, opened 2024-10-03
ACM-14332 > Clusters page CSV export fixes extra space

vishsanghishetty:
[ACM-14332](https://issues.redhat.com/browse/ACM-14332) > Clusters page CSV export fixes extra space


KevinFCormier (changes requested):
Please fix the problem at its source:
https://github.com/stolostron/console/blob/main/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx#L975

@Randy424 I see this one is not translated... we should probably set that up while we are at it as well.

PR #3957, opened 2024-10-10
Rosa-cluster-via-ocm-edge-case

vishsanghishetty:


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -132,6 +135,7 @@ type Action =
   | ({ type: 'setClusterID' } & Pick<State, 'clusterID'>)
   | ({ type: 'setNamespace' } & Pick<State, 'namespace'>)
   | ({ type: 'setCredential' } & Pick<State, 'credential'>)
+  | { type: 'updateCredentials'; payload: { credentialList: ProviderConnection[] } }
```
KevinFCormier, line 138: I had mentioned that you do not need to pass in `ocmCredentials` (named `credentialList` here) because the reducer function can already access it. See line 311 for example.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -115,7 +115,10 @@ type State = {
   api_token: string
   namespace: string
   credentials: ProviderConnection[]
+  credentialList: ProviderConnection[]
   credential: string
+  isNamespaceAvailable?: boolean
+  isCredentialAvailable?: boolean
```
KevinFCormier, line 121: We don't need to add anything new to the state.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
```diff
@@ -314,6 +321,22 @@ export default function ImportClusterPage() {
...
+          const credentialExists = updatedCredentialList.some(
+            (credential) => credential.metadata.name === state.credential
+          )
+
+          return {
+            ...state,
+            isNamespaceAvailable: namespaceExists,
+            isCredentialAvailable: credentialExists,
+            credentialList: updatedCredentialList,
+          }
```
KevinFCormier, line 338: What I wanted you to do was think about these situations:
- Is the currently selected namespace (if any) still available
- Is the currently selected credential (if any) still available
And then decide the best way to handle them. 

For example, suppose in the import form I have selected namespace `default` and credential `ocm-cred`, and that is the only OCM credential in that namespace. Someone else deletes the credential. There is one other OCM credential defined in namespace `my-project`. Do we update the state so that `namespace` and `credential` are empty string, clearing out the dropdowns in the form? Or do we automatically select the first namespace and the first credential in that namespace?

So some of the code you have here is a good start - you will want to check if we have a namespace selected if it still exists. For the credential, note that the name is only unique within a namespace, so we could have credential `ocm-creds` defined in two different namespaces. Might be a bit confusing if which namespace we were using changed on us.

KevinFCormier (changes requested):
A good start! I left some comments and questions.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ImportCluster/ImportCluster.tsx
vishsanghishetty: thank you for the feedback @KevinFCormier, I will work on fixing it

PR #3966, opened 2024-10-14
[ACM-14640] Use _hubClusterResource insterad of "local-cluster" for hub identific…

zlayne:
…ation

This changes the search pages to use _hubClusterResource field to determine if the resource is located on hub rather than using the static "local-cluster" string. 
2.13 users can custom name the hub cluster - so we can't rely on "local-cluster" string.


robdolares (approved):
/lgtm

PR #3969, opened 2024-10-15
[ACM-14872] Remove deprecated overview page layout

zlayne:
In 2.12 we deprecated the classic overview layout.
This change for 2.13 officially removes the classic overview layout - so fleet view will be the only view.


robdolares (approved):
/lgtm

PR #3971, opened 2024-10-15
ACM-14860 Edge case ROSA import bug

vishsanghishetty:
https://issues.redhat.com/browse/ACM-14860

Bug Description:
The Import Cluster wizard does not dynamically update when OCM credentials are added or deleted. If a credential is removed or added from an external source while the wizard is open, the list of available credentials and the selected namespace/credential dropdowns are not refreshed accordingly. This can lead to situations where deleted credentials remain selected or unavailable credentials are shown.

Expected Behavior:

When OCM credentials are deleted:
  - If the currently selected credential no longer exists, the wizard should either reset the credential field to an empty state or select the first available credential in the current namespace (if any).
  - If the namespace no longer contains any credentials, the namespace field should be cleared, and the credential dropdown should reset.
When new OCM credentials are added:
  - The wizard should automatically update the credential list to reflect the newly available credentials in the appropriate namespace.

### DIFF
--- a/frontend/src/lib/test-metadata.ts
+++ b/frontend/src/lib/test-metadata.ts
@@ -110,6 +110,18 @@ export const mockCRHCredential2: Secret = {
   },
 }
 
+export const mockCRHCredential3: Secret = {
+  apiVersion: SecretApiVersion,
+  kind: SecretKind,
+  metadata: {
+    name: 'test-credential',
+    namespace: 'ocm',
+    labels: {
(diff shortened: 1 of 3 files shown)


KevinFCormier (changes requested):
Looks good now @vishsanghishetty. Can you please remove the old tests that weren't adding to the coverage (the ones checking for empty string) and the `screen.logTestingPlaygroundURL()` call (as well as the import of `screen`).

PR #3976, opened 2024-10-18
ACM-14986-ArgoCD-apps-remain-on-Ul-forever-after-deletion

jeswanke:


KevinFCormier:
Overrode coverage requirements as code is specifically excluded when running in test environment

PR #3977, opened 2024-10-18
ACM-14986-ArgoCD-apps-remain-on-Ul-forever-after-deletion (#3976)

KevinFCormier:
* ACM-14986-ArgoCD-apps-remain-on-Ul-forever-after-deletion

* fix

* fix

* oops

* Make wrapPromise generic and remove "reason" arg Change reject argument to Error

* tweaks

---------


jeswanke (approved):

KevinFCormier:
Overrode coverage requirements as code is specifically excluded when running in test environment

PR #3985, opened 2024-10-22
[ACM-14990] Change VM table console column header to vm details

zlayne:
Change VM table "Console URL" column header to "VM details". This avoids confusion with the VM Console (terminal)


robdolares (approved):
/lgtm

PR #4008, opened 2024-10-28
ACM-15148-exception-in-InfraEnvironmentsPage.tsx

jeswanke:


KevinFCormier:
@jgyselov John noticed a crash in the Infra Environments page under certain conditions. Could you please review this one character fix and see if it makes sense?

jgyselov:
Sorry, this does not make much sense to me. How come an infraEnv is `null` there?

Also, with this fix, I am worried about the results of this function if the infraEnv happens to be null. We use it to get unique keys - if the infraEnv is null, these keys will hardly be unique, and even if the page doesn't crash, it might cause other issues (e.g. with selecting infraEnvs in the table).

jeswanke:
@jgyselov would a try catch work better---I found this on an environment with 3000 clusters--if this isn't guarded against, user sees an un-oh message when trying to look at the inventory list

PR #4014, opened 2024-10-31
ACM-15206 - Convert policy item to string label

Randy424:
Regarding: https://issues.redhat.com/browse/ACM-15206

Deconstructs policy item, returns string label for source.


robdolares (approved):
/lgtm

PR #4015, opened 2024-10-31
ACM-15130 Replace link to login command page with text instructions

KevinFCormier:
Direct the user to find the login command via the **Copy login command** in the masthead. This is a bit awkward for the modal case, where the masthead will not be available until they close it, but I think most users will be familiar with how to login anyway.
image
image

`window.SERVER_FLAGS` is internal and subject to change, which it did. There is an endpoint we could potentially use to find the URL to direct to, but there is complication because OCP can be configured with external auth. It didn't seem worth it to me to invest the time in trying to make that work, when ultimately these CLI instructions may someday be replaced by full UI support.


fxiang1 (approved):
Looks good Kevin! It's a nice and simple fix. Thanks!

PR #4023, opened 2024-11-01
ACM-15322 Ping-search-api-and-log-cleanup

jeswanke:
1. pings search api to make sure it's there--good when
  a.  there's never going to be a search api (mce)
  b. on startup when search api hasn't come up yet
  c. when search api is just having a bad day
2. consolidated search app log messages


backend/src/lib/search.ts
```diff
@@ -162,3 +159,34 @@ export function getSearchResults(
...
+          reject(e)
+        }
+      })
+    })
+    req.on('error', (e) => {
+      reject(e)
+    })
+    req.write(
+      '{"operationName":"searchResult","variables":{"input":[{"filters":[{"property":"kind","values":["Application"]},{"property":"apigroup","values":["argoproj.io"]},{"property":"cluster","values":["!local-cluster"]},{"property":"name","values":["a*","i*","n*"]}],"limit":1}]},"query":"query searchResult($input: [SearchInput]) {\\n  searchResult: search(input: $input) {\\n    items\\n  }\\n}"}'
+    )
```
KevinFCormier, line 189: This seems like a rather arbitrary query to execute as a test. Perhaps @jlpadilla or @zlayne might have a suggestion for a simpler check for whether the search API is ready.

jeswanke: using @zlayne suggestion to ping for existence of pod

PR #4026, opened 2024-11-01
Verify CSV output format in tests

KevinFCormier:
Current tests pass even if you comment out the click on the export option, so while they may exercise the code to some extent, they are not verifying anything.

I updated these as an example, so that they verify the output data and the filename that is used, but there are more tests to be updated in a similar fashion.

### DIFF
--- a/frontend/src/routes/Applications/AdvancedConfiguration.test.tsx
+++ b/frontend/src/routes/Applications/AdvancedConfiguration.test.tsx
@@ -36,7 +36,14 @@ import {
   SubscriptionKind,
 } from '../../resources'
 import { nockIgnoreApiPaths, nockIgnoreRBAC, nockSearch } from '../../lib/nock-util'
-import { clickByTestId, waitForText, clickByLabel, clickByText } from '../../lib/test-util'
+import {
+  clickByTestId,
+  waitForText,
+  clickByLabel,
+  clickByText,
+  getCSVExportSpies,
+  getCSVDownloadLink,
+} from '../../lib/test-util'
 import {
   mockSearchQueryArgoApps,
   mockSearchQueryOCPApplications,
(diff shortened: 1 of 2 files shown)


frontend/src/routes/Applications/AdvancedConfiguration.test.tsx
```diff
@@ -370,97 +377,86 @@ describe('Export from application tables', () => {
...
 
-    expect(createElementSpyOn).toHaveBeenCalledWith('a')
-    expect(anchorMocked.download).toContain('table-values')
-
-    document.body.appendChild = documentBody
-    document.createElement('a').dispatchEvent = documentCreate
+    expect(blobConstructorSpy).toHaveBeenCalledWith(
+      [
+        'Name,Namespace,Clusters,Created\n' +
+          '"helloworld-simple-placement-3","helloworld-simple-placement-3","1 Remote, 1 Local","4 months ago"',
```
KevinFCormier, line 437: @Randy424 Here is another case where we have a relative timestamp in the CSV export data. We'll need to fix this.

frontend/src/routes/Applications/AdvancedConfiguration.test.tsx
```diff
@@ -370,97 +377,86 @@ describe('Export from application tables', () => {
...
-    expect(anchorMocked.download).toContain('table-values')
-
-    document.body.appendChild = documentBody
-    document.createElement('a').dispatchEvent = documentCreate
+    expect(blobConstructorSpy).toHaveBeenCalledWith(
+      [
+        'Name,Namespace,Channel,Applications,Clusters,Time window,Created\n' +
+          '"helloworld-simple-subscription-1","helloworld-simple-ns","ggithubcom-app-samples",-,"None",-,"-"\n' +
+          '"helloworld-simple-subscription-2","helloworld-simple-ns","ggithubcom-app-samples",-,"None",-,"-"\n' +
+          '"helloworld-simple-subscription-3","helloworld-simple-ns","ggithubcom-app-samples","1","1 Remote, 1 Local","Active","-"',
```
KevinFCormier, line 397: @Randy424 We should ensure that we have some data for each column in the tests.

Randy424 (approved):
/lgtm

Randy424:
@KevinFCormier Tracking follow-up work here: https://issues.redhat.com/browse/ACM-15313

frontend/src/routes/Applications/AdvancedConfiguration.test.tsx
jgyselov: @KevinFCormier @Randy424  this is causing issues. I created a temporary fix in https://github.com/stolostron/console/pull/4059 but it will need a better, long term solution.

Randy424: Opened a separate ticket for this issue^ https://issues.redhat.com/browse/ACM-15591

PR #4037, opened 2024-11-04
ACM-15277 - Update SearchInput filter state location and clear filter callback on AcmTable

Randy424:
Regarding: https://issues.redhat.com/browse/ACM-15277


frontend/src/ui-components/AcmSearchInput/AcmSearchInput.tsx
```diff
@@ -53,7 +53,8 @@ export interface AcmSearchInputProps extends SearchInputProps {
   searchableColumns?: SearchableColumn[]
   canAddConstraints?: boolean
   disableAddConstraint?: boolean
-  pendingConstraints?: SearchConstraint[]
+  pendingConstraints: SearchConstraint[]
+  setPendingConstraints: (constraints: SearchConstraint[]) => void
```
Randy424, line 57: Control of this piece of state needed to be pulled out of the SearchInput, and controlled in the AcmTable component. Now the state and it's setter hook are handed back to the SearchInput as props.

PR #4049, opened 2024-11-07
ACM-15430 Fix control plane type calculation for details page

KevinFCormier:
https://issues.redhat.com/browse/ACM-15430

Remove code duplication between cluster list and details, and use most up-to-date version (cluster list).


vishsanghishetty (approved):
/lgtm

PR #4070, opened 2024-11-14
ACM-13002-Policy-overview-with-lots-of-clusters

jeswanke:
@mprahl @showeimer 

This is meant to address the case where the user has lots of violating clusters in their environment:

**BEFORE:** all clusters in environment were listed alphabetically till the end--if there are 3000 clusters, it will be a very long list:
CleanShot 2024-11-14 at 12 20 07

**AFTER:** the clusters with the most violations will be sorted to the top which will naturally put the compliant clusters at the bottom which will be hidden in a collapsed section:
CleanShot 2024-11-14 at 12 23 16

If you look at Scott's dev environment,  the cluster that seems to require the most attention is towards the bottom of the list:
CleanShot 2024-11-14 at 12 28 42@2x


showeimer:
Hey @jeswanke, this looks good for now; and thanks for taking the time to try to improve this.  I was aware that this card was definitely not going to scale well with high cluster counts; so I think this workaround is probably okay for now.  I am planning to have this card moved into a proper table to make this experience more first-class and simultaenously address this scale problem; so there will probably come a time where this card will be removed.  This is a decent solution for the time being, thanks again!

backend/src/app.ts
```diff
@@ -34,6 +34,8 @@ const isProduction = process.env.NODE_ENV === 'production'
 const isDevelopment = process.env.NODE_ENV === 'development'
 const eventsEnabled = process.env.DISABLE_EVENTS !== 'true'
 
+// process.env.MOCK_CLUSTERS = '100'
```
yiraeChristineKim, line 37: Is this comment required?

backend/src/lib/server-side-events.ts
```diff
@@ -277,12 +278,24 @@ export class ServerSideEvents {
...
+    let allEvents: ServerSideEvent[] = []
     for (const eventID in this.events) {
       if (Number(eventID) <= lastEventID) continue
-      this.sendEvent(clientID, this.events[eventID])
-      sentCount++
+      allEvents.push(this.events[eventID])
+    }
+
+    // mock a large environment
+    if (process.env.MOCK_CLUSTERS) {
```
yiraeChristineKim, line 288: This is for test right?

yiraeChristineKim: Retaining test code is good!

jeswanke: I removed that--- I had added functionality to mock lots of clusters but I moved it into it's own pr

PR #4071, opened 2024-11-14
Make all ports configurable with environment variables

KevinFCormier:
This allows us to avoid port conflicts, including running 2 different versions of the console simultaneously.

For example, from one clone of the repository, run:
```
npm run setup
npm run plugins
```

And from the other, run:
```
export CONSOLE_PORT=9001 BACKEND_PORT=4001 MCE_PORT=3003 ACM_PORT=3004
npm run setup
npm run plugins
```

This also works for the standalone development console, for which you only need to set `BACKEND_PORT` and `FRONTEND_PORT`.


frontend/webpack.config.ts
dhaiducek, line 1: Since the defaults live in the `port-defaults.sh` file now, for consistency did you want to remove the default fallback for the backend here also?

https://github.com/stolostron/console/blame/a5501261d0abcf420c28209ced6c66fb8b504fde/frontend/webpack.config.ts#L174

KevinFCormier: Yes, thank you!

dhaiducek (approved):
LGTM!

PR #4076, opened 2024-11-15
ACM-14462-Application-page-with-lots-of-clusters:

jeswanke:
all done in AcmTable-- applies to application table or any table with lots of filter options

if one of the filters has more then 30 options, table creates a second dropdown filter that can be searched:

CleanShot 2024-11-15 at 12 57 28@2x


KevinFCormier (approved):
Looks great, John!

I only found two small issues which I don't think we would hit in practice. I played around by lowering the threshold to a very small number.

- The search is by option value, not by display label, so if we ever have a filter with translated labels that exceeds the threshold, searching for it is strange. This works fine for clusters where the value and label are the same.
- If all filters are removed from the main filter, the main filter still appears, but empty.

PR #4090, opened 2024-11-20
Reduce renders and race conditions in policy template details

mprahl:
This refactors policy template details components to reduce renders.

Additionally, fireManagedClusterView was improved to be able to explicitly name the created ManagedClusterView objects to prevent conflicts and properly ignore duplicate fireManagedClusterView calls. This mostly helps local development or when the console is run in strict mode and hooks are run twice. It also helps when two separate browser windows are looking at the same policy template, especially Kyverno ones with many audit violations.

Lastly, some loading states were being set to false after the fireManagedClusterView promise was created but before it resolved. This is now fixed to wait for the promise to be resolved.

### DIFF
--- a/frontend/src/routes/Governance/components/ViewDiffApiCall.tsx
+++ b/frontend/src/routes/Governance/components/ViewDiffApiCall.tsx
@@ -2,6 +2,7 @@
 
 import { useEffect, useState } from 'react'
 import { fireManagedClusterView } from '../../../resources'
+import { v4 as uuidv4 } from 'uuid'
 import { ResultsTableData } from '../policies/policy-details/PolicyDetailsResults'
 import { useTranslation } from '../../../lib/acm-i18next'
 import { Alert, Button, Divider, Modal, ModalVariant, Skeleton, Title } from '@patternfly/react-core'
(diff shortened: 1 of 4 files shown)


frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetails.tsx
```diff
@@ -455,57 +440,80 @@ const KyvernoMessages = ({
...
+              return
+            }
+
             if (viewResponse?.message) {
               setErrString(viewResponse.message)
             } else {
               // policy field: namespace/policyName
               const results: { message: string; rule: string; policy: string }[] = viewResponse?.result?.results.filter(
-                ({ policy: policyNsName, rule, message }: { policy: string; rule: string; message: string }) => {
+                ({ policy: policyNsName }: { policy: string; rule: string; message: string }) => {
```
yiraeChristineKim, line 475: You can delete "rule: string; message: string" type

yiraeChristineKim:
@mprahl Amazing solution for dev mode

frontend/src/routes/Governance/components/ViewDiffApiCall.tsx
```diff
@@ -21,26 +22,47 @@ export function ViewDiffApiCall({ item }: Readonly<{ item: ResultsTableData }>)
...
+
     if (isModalOpen) {
       setIsFetching(true)
       setRelatedObjs(undefined)
       setTemplateErr(undefined)
 
-      fireManagedClusterView(cluster, kind, apiVersion, templateName, cluster)
+      const viewName = process.env.NODE_ENV === 'test' ? undefined : uuidv4()
+
+      fireManagedClusterView(cluster, kind, apiVersion, templateName, cluster, viewName, viewName !== undefined)
```
yiraeChristineKim, line 34: Our goal was to reduce the frequency of hitting fireManagedClusterView. But this still hitting fireManagedclusteview when ignore is true. ?  should more if ignore return to before firemanaged?

mprahl: The goal to reduce the `fireManagedClusterView` was in reducing the rerenders. The `ignore` variable is so that if the effect is reexecuted, the old promise results are ignored and don't affect the UI results.

KevinFCormier (commented):
@mprahl Thanks for looking into this. It's been awhile since I looked at this code, so I'm trying to make sense of it.

- When creating a `ManagedClusterView`, it takes some time to return the result, so we generate a stable name for it based on the cluster name, resource name, and kind
  - Note: This is not actually unique enough; it would need to also include account namespace and API version; but good enough if we'll only use this in testing going forward
- The original code does not take into account the fact that multiple users might be looking at the same page, and needing the results from the same `ManagedClusterView`, because as soon as we get results, we delete the resource - so it cannot really be shared by different users.

To address this last point, you started using `uuidv4()` to generate the name of the `ManagedClusterView`, but only for your particular use-case.

I think we should probably move this into the implementation, so that all users benefit. But this would probably require a larger refactor to transform this into a proper hook on its own, so if we want to start with only having the fix for this particular use-case, that is OK with me.

I am still unsure about whether it is correct to skip the check for the existing `ManagedClusterView` though. This code pre-dates the upgrade to React 18 and the strict mode double mounting, so I'm wondering what case it was handling. If we generate a new UUID for the name, then I see that this request would be useless, but what case was it optimizing before? Wondering if @zlayne has any recollection?

zlayne:
May not quite be following the updates here. But if I remember correctly the intent of looking for the existing MCV was just to avoid duplicates. Since the naming was just cluster-name-kind and not unique to each newly created MCV creating duplicates was possible. But if we are using a unique MCV name each time then that shouldn't be of concern i guess.

PR #4110, opened 2024-11-25
ACM-13003-Fast-cluster-page-load

jeswanke:
1. environment label MOCK_CLUSTERS="number" in order to mock that many clusters and
      3 policies and
      44 system apps for each cluster
3. that context error you found was caused by trying to use PluginContext in acm (mce only it looks like)
4. there's a new LoadStatusProvider that any plugin can use that tells at what stage LoadData is:
  loadStarted --> the first packet has been received, if the data for that page is in the first packet it can start displaying
  loadCompleted --> the load of all data is done (the old loaded context value)

the first packets are these:
        case 'ManagedCluster':
        case 'HostedCluster':
        case 'ClusterDeployment':
        case 'Policy':
        case 'PolicySet':
        case 'AgentClusterInstall':
        case 'ManagedClusterInfo':
        case 'ManagedClusterAddOn':
        case 'Search':
        case 'Secret':


KevinFCormier (commented):
Noted a bad import here. I will ping you separately about another change, and then re-test. Aiming to deliver this on Thursday (start of Train-22).

frontend/src/ui-components/AcmEmptyState/AcmEmptyState.tsx
```diff
@@ -8,16 +8,21 @@ import {
   EmptyStateIcon,
   Title,
 } from '@patternfly/react-core'
-import { ReactNode } from 'react'
+import { ReactNode, useContext } from 'react'
 import { CubesIcon, SearchIcon } from '@patternfly/react-icons'
+import { LoadingPage } from '../AcmLoadingPage/AcmLoadingPage.stories'
```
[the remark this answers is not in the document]
jeswanke: fixed

KevinFCormier (changes requested):
Hey John,

I'm seeing some loading issues in places.

Applications, Advanced configuration tab
- All 4 tables show a skeleton indefinitely

Governance, Overview tab
- Shows spinner indefinitely
Governance, Policies / Discovered policies tabs
- Shows table skeleton indefinitely

jeswanke:
hey Kevin,

oops I used:
  const { loadCompleted } = useContext(PluginDataContext)
instead of:
  const { dataContext } = useContext(PluginContext)
  const { loadCompleted } = useContext(dataContext)

all better now

backend/src/lib/server-side-events.ts
```diff
@@ -276,12 +289,90 @@ export class ServerSideEvents {
...
+
+    // mock a large environment
+    if (process.env.MOCK_CLUSTERS) {
+      const loaded = parts.pop()
+      parts = [...parts, ...getGiganticEvents()]
+      parts.push(loaded)
+    }
+
+    const start = parts.splice(0, 2)
+    const end = parts.pop()
```
KevinFCormier, line 305: The hard-coded `2` appears to be here so that we send the START event and the SETTINGS event. The settings event is not always the 2nd event - if settings change, it will appear later in the event stream.

Some comments here would be good to indicate that we depend on the first and last events being certain things.

frontend/src/components/LoadData.tsx
```diff
@@ -515,12 +534,14 @@ export function LoadData(props: { children?: ReactNode }) {
...
 
-    const timeout = setInterval(processEventQueue, THROTTLE_EVENTS_DELAY)
+    const timeout = setInterval(processEventQueue, 500)
     return () => {
       clearInterval(timeout)
       if (evtSource) evtSource.close()
     }
-  }, [caches, mappers, setters, setSettings])
+    // this effect must only run once--it sets up the call to /events on the backend
+    // this should be [], but linter would complain
```
KevinFCormier, line 543: I think it would be better to use the directive to silence the linter here, and to keep the comment explaining why you've done so. It looks to me like the list of dependencies is all static, so you could also point out in the comment that they shouldn't change.

frontend/src/components/PluginDataContextProvider.test.tsx
```diff
@@ -18,22 +18,11 @@ const TestPluginDataContextProvider = (props: ProviderProps<PluginData>) => {
...
-  it('does not render LoadData until requested', async () => {
-    const pluginData: PluginData = {
-      ...defaultContext,
-      loaded: false,
-      load: jest.fn(),
-    }
-    render(<TestPluginDataContextProvider value={pluginData} />)
-    await waitForText('Loading')
-    expect(pluginData.load).toHaveBeenCalled()
-  })
```
KevinFCormier, line 30: This test verified 2 things:
- the `load()` function is called if we're not already loaded
- 'Loading' displays until then

We no longer have the 'Loading' text to display, but we could keep the other part of this test.

frontend/src/routes/Governance/components/GovernanceEmptyState.tsx
```diff
@@ -77,6 +77,7 @@ export function GovernanceCreatePolicysetEmptyState(props: { rbac: boolean }) {
           components={{ bold: <strong /> }}
         />
       }
+      ignoreLoading={true}
```
KevinFCormier, line 80: Why do we ignore loading in this case? (Probably needs a comment in the code.)

frontend/src/routes/Governance/overview/Overview.tsx
```diff
@@ -31,6 +31,9 @@ import { SecurityGroupPolicySummarySidebar } from './SecurityGroupPolicySummaryS
 import keyBy from 'lodash/keyBy'
 import { TFunction } from 'react-i18next'
 import { JSX } from 'react/jsx-runtime'
+import { LoadingPage } from '../../../components/LoadingPage'
+// import { PluginDataContext } from '../../../lib/PluginDataContext'
```
KevinFCormier, line 35: Remove commented out code

frontend/src/routes/Governance/overview/Overview.tsx
```diff
@@ -385,12 +397,12 @@ function renderClusterList(
   return (
     <div style={{ paddingBottom: '10px', display: 'grid', gridTemplateColumns: '1fr auto auto auto auto', gap: 16 }}>
       {clusterList.map(({ cluster, violations }) => {
-        const key = `${cluster.metadata.name}-card`
+        const key = `${cluster?.metadata?.name}-card`
```
KevinFCormier, line 400: Seems strange that optional chaining would be needed here. And that `cluster` is of type any. Looks like we have the types at all call sites of `renderClusterList`, so why not use them? Could we update the function like this?

```
function renderClusterList(
  clusterList: { cluster: ManagedCluster; violations: ViolationSummary }[],
  onClick: { (cluster: ManagedCluster, compliance: string): void },
  t: TFunction
) {
```

I don't follow why the `onClick` has two different signatures currently.

frontend/src/routes/Search/Details/DetailsPage.test.tsx
```diff
@@ -133,19 +133,4 @@ describe('DetailsPage', () => {
...
-        search: '?',
-        state: {
-          from: '/multicloud/search',
-          fromSearch: '?filters={%22textsearch%22:%22kind%3APod%2',
-        },
-      },
-    })
-    const res = getResourceParams()
-    expect(res).toMatchSnapshot()
-  })
```
KevinFCormier, line 150: Why are you removing this test?

KevinFCormier (commented):
Please see inline comments. A few updates are needed.

frontend/src/components/LoadData.tsx
jeswanke: true that-- this confused me at first--I added some states to list and suddenly /events was being called multiple times, which is ng

frontend/src/routes/Governance/components/GovernanceEmptyState.tsx
jeswanke: doesn't seem to have any effect, removing

frontend/src/routes/Search/Details/DetailsPage.test.tsx
jeswanke: keeps messing up the snapshot--I'll put it back

frontend/src/components/PluginDataContextProvider.test.tsx
jeswanke: restored and testing for load()

PR #4117, opened 2024-11-29
Use OpenShift console timestamp component

KevinFCormier:
Makes the `<Timestamp />` component from OpenShift console available and simplifies the current mechanism for using console dynamic plugin SDK hooks and components while maintaining the ability to run in standalone mode.

The `<Timestamp />` component gives us an easy way to have consistent display of timestamps that are localized.

**Placeholder for standalone dev mode:**
image

**Demonstrating translated date/time string (French) and tooltip, with icon removed:**
image

**Relative time display (once the current bug is implemented):**
image

### TODOs:

- [x] Implement a simplified version of the component for standalone development use
- [x] Implement a wrapper component to handle removing the icon (which we will not want to use in a table context)
- [ ] Contribute a fix for the broken relative string support (see draft PR https://github.com/openshift/console/pull/14555)

### DIFF
--- a/frontend/src/components/ACMNotReadyWarning.test.tsx
+++ b/frontend/src/components/ACMNotReadyWarning.test.tsx
@@ -7,8 +7,7 @@ import { clickByRole, waitForText } from '../lib/test-util'
 import { SubscriptionOperator, SubscriptionOperatorApiVersion, SubscriptionOperatorKind } from '../resources'
 import { nockIgnoreOperatorCheck } from '../lib/nock-util'
 import { ACMNotReadyWarning } from './ACMNotReadyWarning'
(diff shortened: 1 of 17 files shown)


vishsanghishetty:
@KevinFCormier I just updated with a date formatting function.

KevinFCormier:
Thanks @vishsanghishetty 
I made a couple small changes like using en-US instead of en-GB and making "09:30 AM" display as "9:30 AM" without the leading zero.

Randy424 (approved):
/lgtm

vishsanghishetty:
@KevinFCormier and @Randy424 please review, reverted the changes in `ManagedClusters` component

Randy424:
Will re-review once this goes green.

PR #4123, opened 2024-12-03
ACM-15840 Validate hosted cluster namespace for HCP on KubeVirt wizard

vishsanghishetty:
## Description
This PR is for [ticket](https://issues.redhat.com/browse/ACM-15840) 

## Acceptance Criteria
Adds validation for 
- [x] Hosted cluster namespace dropdown does not show namespaces of managed clusters
- [x] Validation of the Hosted cluster namespace field is enhanced to check that the name is a valid namespace name AND that it does not match an existing managed cluster or the name of the cluster currently being created


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/CreateCluster.tsx
```diff
@@ -325,6 +325,36 @@ export default function CreateCluster(props: { infrastructureType: ClusterInfras
       case 'connection':
         setConnectionControl(control)
         break
+      case 'namespace':
+        control.validation.contextTester = (
+          active: string | undefined,
+          _controlData: object,
+          templateObjectMap: { [x: string]: { HostedCluster: { $raw: { metadata: { namespace: any; name: any } } }[] } }
+        ) => {
+          if (infrastructureType === Provider.kubevirt && templateObjectMap['<<main>>'].HostedCluster[0]) {
```
KevinFCormier, line 334: This is common code, so you are correct to check that we are only making this change for `Provider.kubevirt`... but is the check in the correct place? Currently, I don't think any other provider has a `namespace` control, but what would happen if one did?

KevinFCormier (changes requested):
Hi Vishali,

Validation seems to be working well. I am still trying to figure out why the test does not work.

In the meantime, I noticed a small issue in the code (see inline comment). And I noticed that this PR does not meet this acceptance criteria from the story:

- **Hosted cluster namespace** dropdown does not show namespaces of managed clusters

It's good to have the validation in case the user types in the name of a managed cluster namespace, but we should not include them in the dropdown from the beginning.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/CreateCluster.tsx
```diff
@@ -325,6 +325,36 @@ export default function CreateCluster(props: { infrastructureType: ClusterInfras
...
+            if (managedClusters.length) {
```
KevinFCormier, line 335: This check is also not quite in the right place. Even if there are zero managed clusters, we should still not allow the HostedCluster name and namespace to match, because we will be creating a ManagedCluster of the same name.

vishsanghishetty:
@oafischer can you please review the translation strings and let us know the changes needed?

frontend/public/locales/en/translation.json
```diff
@@ -1514,6 +1514,7 @@
   "Hosted control plane": "Hosted control plane",
   "Hosted control plane is already upgrade to the latest version available. Cluster node pools can be upgraded to match the control plane.": "Hosted control plane is already upgrade to the latest version available. Cluster node pools can be upgraded to match the control plane.",
```
oafischer, line 1515:
```suggestion
  "Hosted control plane is already upgraded to the latest version available. Cluster node pools can be upgraded to match the control plane.": "Hosted control plane is already upgraded to the latest version available. Cluster node pools can be upgraded to match the control plane.",
```

frontend/public/locales/en/translation.json
```diff
@@ -1871,6 +1872,7 @@
   "Namespace": "Namespace",
   "Namespace where the selected Argo application resources are deployed.": "Namespace where the selected Argo application resources are deployed.",
   "Namespace: {{namespace}}": "Namespace: {{namespace}}",
+  "namespace.exists.error": "The selected namespace is already used by an existing managed cluster.",
```
oafischer, line 1875:
```suggestion
  "namespace.exists.error": "The namespace you selected is already used by an existing managed cluster.",
```

vishsanghishetty:
@olivelakra thanks for reviewing. can you also review the one on line 1517 which is being added as part of this PR

oafischer (approved):
Line 1517 looks good to me. thanks for pulling me in on this.

frontend/src/resources/managed-cluster.ts
```diff
@@ -57,3 +57,10 @@ export const createManagedCluster = (data: {
     spec: { hubAcceptsClient: true },
   })
 }
+
+export function listManagedClusters() {
+  return listResources<ManagedCluster>({
+    apiVersion: ManagedClusterApiVersion,
+    kind: ManagedClusterKind,
+  })
+}
```
KevinFCormier, line 66: We may want to remove this since it is unused.

frontend/public/locales/en/translation.json
```diff
@@ -1514,6 +1514,7 @@
   "Hosted control plane": "Hosted control plane",
   "Hosted control plane is already upgrade to the latest version available. Cluster node pools can be upgraded to match the control plane.": "Hosted control plane is already upgrade to the latest version available. Cluster node pools can be upgraded to match the control plane.",
   "Hosted control plane operator must be enabled in order to continue": "Hosted control plane operator must be enabled in order to continue",
+  "hosted.cluster.namespace.error": "Namespace name cannot be the same as the cluster name",
```
KevinFCormier, line 1517:
```suggestion
  "hosted.cluster.namespace.error": "The namespace cannot be the same as the cluster name.",
```

frontend/public/locales/en/translation.json
KevinFCormier: I know this is not part of your work @vishsanghishetty, but it's a small thing we might as well fix now.

PR #4129, opened 2024-12-05
ACM-15850-Polling-from-console-pods-adding-too-much-load-on-search

jeswanke:
good news-- no more fixed a*, b* table-- now the queries are created automagically:

https://docs.google.com/document/d/1EB4WxK-z1570Of2ETksKr7T4AHWQDXxFrcFS02_HWcQ/edit?tab=t.vgwxdgm2vbv2


backend/src/lib/config.ts
```diff
@@ -43,7 +43,7 @@ export async function loadConfigSettings(): Promise<void> {
       }
     }
     for (const key in settings) {
-      if (key.startsWith('LOG_')) {
+      if (key.startsWith('LOG_') || key.startsWith('APP_SEARCH_')) {
```
KevinFCormier, line 46: Nit: I guess these are settings that are used in the backend only. A comment here would be good.

backend/src/routes/aggregators/utils.ts
```diff
@@ -0,0 +1,516 @@
...
+    let applications: ITransformedResource[] = []
+    if (applicationCache[remoteCacheKey].resources) {
+      applications = applicationCache[remoteCacheKey].resources
+    } else if (applicationCache[remoteCacheKey].resourceMap) {
+      applications = Object.values(applicationCache[remoteCacheKey].resourceMap).flat()
+    }
+    if (applications.length) {
+      // create array of frequency of name prefixes
+      const a = 'a'.charCodeAt(0)
+      const z = '0'.charCodeAt(0)
```
KevinFCormier, line 295: Nit: If `a` is the letter 'a'... then I expect `z` to be the letter 'z', not zero. Can you choose better names, please?

backend/src/routes/aggregators/utils.ts
```diff
@@ -0,0 +1,516 @@
...
+      const sz = 26 + 10
+      const prefixFrequency = new Array(sz).fill(0) as number[]
+      applications.forEach((app) => {
+        const name = app.transform[AppColumns.name][0]
+        const ltr = name.charCodeAt(0)
+        const index = ltr < a ? ltr - z + 26 : ltr - a
```
KevinFCormier, line 301: Nit: Is this index mapping really necessary? Wouldn't it be a whole lot easier to store the values in a map by character?

KevinFCormier (approved):

PR #4140, opened 2024-12-09
Fix critical console bugs related to policies with exceptional templates

yiraeChristineKim:
Ref: https://issues.redhat.com/browse/ACM-16238

### DIFF
--- a/frontend/src/routes/Governance/common/util.tsx
+++ b/frontend/src/routes/Governance/common/util.tsx
@@ -484,7 +484,7 @@ export function getPolicyRemediation(policy: Policy | undefined, propagatedPolic
   const remediationSet = new Set()
 
   templates?.forEach((template: PolicyTemplate) => {
-    const templateRemediation = template.objectDefinition.spec.remediationAction
+    const templateRemediation = template.objectDefinition.spec?.remediationAction
     remediationSet.add(templateRemediation)
     if (remediationSet.size === 3) {
       return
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Governance/common/util.tsx
```diff
@@ -484,7 +484,7 @@ export function getPolicyRemediation(policy: Policy | undefined, propagatedPolic
   const remediationSet = new Set()
 
   templates?.forEach((template: PolicyTemplate) => {
-    const templateRemediation = template.objectDefinition.spec.remediationAction
+    const templateRemediation = template.objectDefinition.spec?.remediationAction ?? 'inform'
```
yiraeChristineKim, line 487: Default is inform

dhaiducek: Should the template be ignored instead if `remediationAction` isn't set? Otherwise, it won't hit this code block further down:
```
    // empty root remediation and template remedation all empty
    if (!rootRA) {
```

dhaiducek (changes requested):

frontend/src/routes/Governance/common/util.tsx
yiraeChristineKim: because of informOnly we should consider template remediation when root is set enforce or inform

dhaiducek: I guess what I'm poking at is that in the issue, a policy type isn't even provided and it's missing the `spec` also. I feel like if the `spec` is missing AND it's not even a policy type, the remediationAction is ignored. Otherwise, with the policy in the issue, when the root remediation isn't defined the UI would still show `inform`, which isn't accurate.

yiraeChristineKim: OH now I get it  Thanks!

dhaiducek (approved):
LGTM!

dhaiducek (approved):

yiraeChristineKim:
/override ci/prow/unit-tests-sonarcloud

no issue in local test

dhaiducek:
@yiraeChristineKim Does this need to be backported to ACM 2.12 and 2.11?

yiraeChristineKim:
@dhaiducek Thanks Dale!!

PR #4142, opened 2024-12-10
ACM-15271-Prevent unintended react form wizard submission

Randy424:
Regarding: [ACM-15271](https://issues.redhat.com/browse/ACM-15271)

Before: 
Pressing enter while focused on a textInput will result in form submission.

https://github.com/user-attachments/assets/2d306cf4-48b8-4f43-bba9-69970cc4c575

After: 
Form submission via "Enter" key is suppressed within the ArgoWizard and PolicyWizard components.

https://github.com/user-attachments/assets/1a18d3cc-4c5c-457f-b78b-0c8eacbe6cc3

### DIFF
--- a/frontend/src/resources/utils/utils.ts
+++ b/frontend/src/resources/utils/utils.ts
@@ -17,6 +17,20 @@ export function getLatest<T>(items: T[], key: string) {
   })
 }
 
+export const preventDefaultSubmissionWithListener = () => {
+  const handleKeyDown = (event: KeyboardEvent) => {
+    if (event.key === 'Enter') {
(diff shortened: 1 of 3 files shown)


KevinFCormier (changes requested):
I don't think this is the right fix. 
- Since the issue appears in any form wizard of this type, the fix should be made in the wizard itself so it doesn't have to repeated everywhere 
- The fix blocks legitimate presses of the Enter key, like on the buttons to go to the next step or cancel, or in the YAML editor, etc.

Randy424:
@KevinFCormier Right, bad move.

PR #4179, opened 2025-01-06
ACM-3448 Credentials created column relative time

vishsanghishetty:
## Description
This PR is for [ticket](https://issues.redhat.com/browse/ACM-3448) 
This PR is for [ticket](https://issues.redhat.com/browse/ACM-3456) 

## Acceptance Criteria
- [x] Uses [PR](https://github.com/stolostron/console/pull/4117) by @KevinFCormier 
- [x] Changes Creation date on the Clusters page
- [x] Changes Created date in the Credentials page
- [x] Changes Last report in Governance > Policies > Policy >Results page
- [x] Unit tests


frontend/src/lib/AcmTimestamp.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { css } from '@emotion/css'
+import { SimpleTimestamp } from './SimpleTimestamp'
+import { PluginContext } from './PluginContext'
+import { useContext } from 'react'
+import { TimestampProps } from '@openshift-console/dynamic-plugin-sdk'
+
+type AcmTimestampProps = TimestampProps & {
+  showIcon?: boolean
+  relative?: boolean
```
KevinFCormier, line 11: The `relative` prop is only used by the fallback implementation that we use for local development. Since it doesn't affect the plugin version, I don't think we should offer it. It makes standalone behave differently from plugin version:
image

In the plugin, the relative time would only be displayed for times within about 10 minutes of the current time, if the bug I found gets fixed. But currently, a relative time is never shown.

So I don't think it's useful to support relative times in the fallback implementation. If we did, it should behave in the same way as in the plugin, only being used for dates that at 10 minutes old.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/ManagedClusters.tsx
```diff
@@ -991,7 +992,7 @@ export function useClusterCreatedDateColumn(): IAcmTableColumn<Cluster> {
     search: 'creationDate',
     cellTransforms: [nowrap],
     cell: (cluster) => {
-      return getCreationTimestampString(cluster)
+      return <AcmTimestamp timestamp={cluster.creationTimestamp ?? ''} />
```
KevinFCormier, line 995: This change is not mentioned in the PR description. Can you update please?

frontend/src/routes/Credentials/CredentialsPage.tsx
```diff
@@ -231,13 +231,18 @@ export function CredentialsTable(props: {
...
+                  <AcmTimestamp timestamp={resource.metadata.creationTimestamp} relative={true} />
+                ) : (
+                  '—'
+                )}
               </span>
             ),
             exportContent: (item: Secret) => {
               if (item.metadata.creationTimestamp) {
-                return moment(new Date(item.metadata.creationTimestamp)).fromNow()
+                return new Date(item.metadata.creationTimestamp).toISOString()
```
KevinFCormier, line 243: I am not sure if we should make this change. There definitely is a problem here - we should not be exporting relative date-time.

When reviewing the fix for [ACM-16979](https://issues.redhat.com/browse/ACM-16979) by @Randy424, I was thinking about suggesting we use ISO 8601, which is the true standard for dates. However, I found that spreadsheet programs don't seem to like this format... Numbers at least will only recognize it as text. This is also true of the type of date-times that we are exporting where we have fixed it.

I suggest we leave it to @Randy424 to address this via some open defects he currently has to improve testing around exported values. We should export date-times in a consistent way, and it would be great if we can remove the use of `moment` altogether.

frontend/src/routes/Governance/policies/policy-details/PolicyDetailsResults.tsx
```diff
@@ -277,9 +277,8 @@ export default function PolicyDetailsResults() {
         header: t('Last report'),
         sort: 'timestamp',
         cell: (item: ResultsTableData) =>
-          item.timestamp ? moment(item.timestamp, 'YYYY-MM-DDTHH:mm:ssZ').fromNow() : '-',
-        exportContent: (item: ResultsTableData) =>
-          item.timestamp ? moment(item.timestamp, 'YYYY-MM-DDTHH:mm:ssZ').toString() : '-',
+          item.timestamp ? <AcmTimestamp timestamp={item.timestamp} relative={true} /> : '—',
+        exportContent: (item: ResultsTableData) => (item.timestamp ? new Date(item.timestamp).toLocaleString() : '—'),
```
KevinFCormier, line 281: I suggest leaving this one to @Randy424 as well.

KevinFCormier (changes requested):
Please see the inline comments for a few things to address. I think we should leave any changes to the exported values out of this PR. Also, please update the PR description for accuracy - it says the PR removes the `moment` library, but it does not.

KevinFCormier:

KevinFCormier:

PR #4183, opened 2025-01-08
ACM-15128 networking options support for node pools HCP cluster

vishsanghishetty:
### OpenShift Virtualization wizard supports networking options for node pools
This PR is for [ticket](https://issues.redhat.com/browse/ACM-15128) 

## Acceptance Criteria
- [x] Feature Addition: Introduced the ability to define additional networks for each node pool in the format `<namespace>/<name>`
- [x] Each node pool maintains its list of additional networks, enabling independent and specific configurations.
- [x] Attach default pod network checkbox - checked and disabled when there are no additional networks
- [x] Unit tests


KevinFCormier (changes requested):
Hey Vishali,

I haven't looked at the code yet, but I've played around with the behaviour. There are a few things to fix up.

1. On the Review step, additional networks is listed as a required field, but it does not block submission of the form. (See screenshot.) It should not be required, and should not block. Also, we should only include `additionalNetworks` in the YAML if the user has entered some.
2. I should only be able to disable the attachment of the default network if I have entered at least 1 additional network.
3. The default value for `defaultPodNetwork` is `true`, but if I uncheck the box, this property is removed from the YAML, meaning the default of `true` will still apply. We need to explicitly set it to `false` when the box is unchecked.
4. The validation mentions you can use the format `<namespace>/<name>` or just `<name>`. I have not seen any reference to the format with just the name. Are you sure about that part?
image

vishsanghishetty:
@KevinFCormier thanks for taking a look. I will work on fixing the above.

frontend/public/locales/en/translation.json
```diff
@@ -1877,6 +1883,7 @@
   "name": "name",
   "Name": "Name",
   "Name needs to be unique to the namespace on each of the managed clusters.": "Name needs to be unique to the namespace on each of the managed clusters.",
+  "Name specify the network attached to the nodes in the format \"[namespace]/[name]\" to reference the multus network attachment definition": "Name specify the network attached to the nodes in the format \"[namespace]/[name]\" to reference the multus network attachment definition",
```
oafischer, line 1886: Typo here? I think it's supposed to be "The name specifies the network..."

frontend/public/locales/en/translation.json
```diff
@@ -1288,6 +1292,7 @@
   "Ensure these settings are correct. The saved cluster draft will be used to determine the available network resources. Therefore after you press Save you will not be able to change these cluster settings.": "Ensure these settings are correct. The saved cluster draft will be used to determine the available network resources. Therefore after you press Save you will not be able to change these cluster settings.",
   "Enter a name for this search query": "Enter a name for this search query",
   "Enter a value": "Enter a value",
+  "Enter additional networks in the format <namespace>/<name>": "Enter additional networks in the format <namespace>/<name>",
```
KevinFCormier, line 1295: Can we make this singular? Within the text field, you can only enter 1 network name.

frontend/public/locales/en/translation.json
KevinFCormier: This is marked as resolved but not fixed?

frontend/public/locales/en/translation.json
```diff
@@ -3014,6 +3022,7 @@
   "tooltip.creation.app.reconcileRate": "The frequency of resource reconciliation that is used as a global repository setting. The medium default setting checks for changes to apply every three minutes and re-applies all resources every 15 minutes, even without a change. Select low to reconcile every hour. Select high to reconcile every two minutes. If you select off, the deployed resources are not automatically reconciled.",
   "tooltip.creation.app.settings.existingRule": "If available in the application namespace, you can select a predefined placement configuration",
   "tooltip.creation.app.subReconcileRate": "Turn the auto-reconciliation off for this specific application regardless of the reconcile rate setting in the repository.",
+  "tooltip.creation.nodepool.additional.network": "Specify additional networks to be attached to the nodes in the format '[namespace]/[name]'.",
```
KevinFCormier, line 3025: This message uses different syntax from the others (`'[namespace]/[name]'` vs `<namespace>/<name>` vs `"[namespace]/[name'"`). Please check if we've used syntax like this before and then be consistent with whatever you find.

frontend/public/locales/en/translation.json
```diff
@@ -3047,6 +3056,7 @@
   "tooltip.creation.ocp.name": "The unique name of your cluster. The value must be a string that contains lowercase alphanumeric values, such as dev. Cannot be changed after creation.",
   "tooltip.creation.ocp.node.pool.autorepair": "Enables machine auto-repair with machine health checks",
   "tooltip.creation.ocp.node.pool.core.count": "Number of cores assigned to each worker node VM",
+  "tooltip.creation.ocp.node.pool.default.pod.network": "Check this box to attach the default pod network to your cluster nodes.",
```
KevinFCormier, line 3059:
```suggestion
  "tooltip.creation.ocp.node.pool.default.pod.network": "Check this box to attach the default pod network to your cluster nodes.",
```
```suggestion
  "tooltip.creation.ocp.node.pool.default.pod.network": "Check this box to attach the default pod network to nodes from this pool.",
```

frontend/src/components/TemplateEditor/utils/validation-types.ts
```diff
@@ -194,3 +194,17 @@ export const VALIDATE_BASE_DNS_NAME_REQUIRED: Validator = {
...
 export const VALID_DNS_LABEL = '^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$'
+
+export const getK8sNameValidator = (t: TFunction): Validator => ({
+  tester: {
+    test: (value) => {
+      const regex = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/
+      const parts = value.split('/')
+      if (parts.length !== 2) return false
+      const [namespace, name] = parts
+      return regex.test(namespace) && namespace.length <= 253 && regex.test(name) && name.length <= 253
```
KevinFCormier, line 205: In your previous story that involved validating the namespace name, we used the `VALID_DNS_LABEL` regex, which is correct. I tested this as follows:
```
% oc create namespace 123456789-123456789-123456789-123456789-123456789-123456789-123456789-foo
The Namespace "123456789-123456789-123456789-123456789-123456789-123456789-123456789-foo" is invalid: 
* metadata.name: Invalid value: "123456789-123456789-123456789-123456789-123456789-123456789-123456789-foo": must be no more than 63 characters
* metadata.labels: Invalid value: "123456789-123456789-123456789-123456789-123456789-123456789-123456789-foo": must be no more than 63 characters
```

Other points for improvement:

1. The name of this validator is too generic. I would call it something like `getNamespaceSlashNameValidator`.
2. If I enter text like `testFoo/Name` I see the error message `Invalid format. Use <namespace>/<name>`. I would prefer for you to use a contextTester like you did in the previous story so that you can have different validation messages for different cases. I would use the new message only if you could not split the string by '/' into 2 parts. (For example, there is no slash, or there are 3 slashes, etc.) If you are able to separate into namespace and name, but one of them is invalid, it would be good to specify which, and then use the standard message for this validator. For example `The namespace is invalid. The value must be a valid DNS label that consists of up to 63 lowercase alphanumeric characters. The character '-' is also permitted, as long as it does not appear in the first or last position.` Or similar if the name does not follow the correct syntax.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/templates/kubevirt-template.hbs
```diff
@@ -181,5 +194,4 @@ spec:
   searchCollector:
     enabled: true
   certPolicyController:
-    enabled: true
-{{/if}}
\ No newline at end of file
+    enabled: true{{/if}}
```
KevinFCormier, line 197: Is there a reason for this change? If not, please put the closing `{{/if}}` on a newline, and end the file with another newline.

KevinFCormier (changes requested):
Functionality is much improved! A few more things to look at around the messages and validation...

frontend/public/locales/en/translation.json
KevinFCormier: These are more consistent now, using angle brackets everywhere, but sometimes it is surrounded by quotes, and other times not.

KevinFCormier:

KevinFCormier:

PR #4198, opened 2025-01-16
VM actions should use cluster-proxy Service instead of Route

zlayne:


backend/src/routes/virtualMachineProxy.ts
```diff
@@ -63,32 +62,17 @@ export async function virtualMachineProxy(req: Http2ServerRequest, res: Http2Ser
...
-              logger.error({ msg: 'Error getting cluster proxy Route', error: err.message })
-              return undefined
-            })
-
           // req.url is one of: /virtualmachines/<action> OR /virtualmachineinstances/<action>
           // the VM name is needed between the kind and action for the correct api url.
           const splitURL = req.url.split('/')
           const joinedURL = `${splitURL[1]}/${body.vmName}/${splitURL[2]}`
-          const path = `${proxyServer}/${body.managedCluster}/apis/subresources.kubevirt.io/v1/namespaces/${body.vmNamespace}/${joinedURL}`
+          const proxyService = 'https://cluster-proxy-addon-user.multicluster-engine.svc.cluster.local:9092'
```
KevinFCormier, line 69: Namespace can be customized by customers, and should not be hardcoded. See https://github.com/stolostron/console/blob/main/backend/src/routes/aggregators/utils.ts#L100 for example.

The other parts of the service addressed being hardcoded is OK, as they won't change in a given released version.

KevinFCormier:

KevinFCormier:

PR #4204, opened 2025-01-20
ACM-17121 Fix global hub cluster detection

KevinFCormier:
https://issues.redhat.com/browse/ACM-17121

Code was originally only checking the presence of the `feature.open-cluster-management.io/addon-multicluster-global-hub-controller` label, but it needs to confirm the value is `available` otherwise non-regional hubs can display as control plane type **Hub**.


KevinFCormier:

KevinFCormier:

PR #4207, opened 2025-01-20
ACM-16958 Fuzzy search highlighting issues on namespace field for cluster list

nitin-dhevar:


KevinFCormier (commented):
Thanks, Nitin! This does indeed fix the problem.

Can you consider 2 things please:

1. Should we include `cluster.hive.clusterPoolNamespace` in the search if it is not shown anywhere in the table?
2. Please take a look at the Sonar results. There is 1 issue listed which is valid.

KevinFCormier:

PR #4237, opened 2025-01-29
ACM-17293: Implement validation for vCenter username

nitin-dhevar:


frontend/src/routes/Credentials/CredentialsForm.tsx
```diff
@@ -338,34 +339,7 @@ export function CredentialsForm(
...
-    let annotations = providerConnection ? providerConnection?.metadata.annotations : undefined
-    if (annotations) {
-      delete annotations['kubectl.kubernetes.io/last-applied-configuration']
-      if (Object.keys(annotations).length === 0) annotations = undefined
-    }
-    if (annotations) {
-      secret.metadata.annotations = annotations
-    }
-
+  function getStringCredentialsData(credentialsType: string, stringData: any) {
```
KevinFCormier, line 342: Was the change to factor out this function made in response to a Sonar complaint about code complexity?

We often ignore those warnings because they are overly aggressive.

In general, although side-effects are permitted in JavaScript, they make code hard to understand, and I avoid them. If you want to do this refactoring, I would recommend moving the creation of the `stringData` object to this function, and returning it from the function. Otherwise the name `getStringCredentialsData` is misleading.

nitin-dhevar: Yes, I made this refactoring due to a sonarqube issue regarding function complexity, shall I revert it then?

KevinFCormier: Yes, I would suggest either tweaking it slightly so that there are no side-effects (the modification of the `stringData` argument object), or just revert.

KevinFCormier:

KevinFCormier:

PR #4241, opened 2025-01-30
ACM-17541: Fix for legend label getting clipped in Clusters overview card.

nitin-dhevar:


frontend/src/routes/Home/Overview/components/SummaryClustersCard.tsx
```diff
@@ -90,7 +90,7 @@ export function SummaryClustersCard(props: {
   }, [chartData, chartLabel?.subTitle, chartLabel?.title, colorScale, isPieChart, title])
 
   const legendWidth = useMemo(() => {
-    const columns = Math.trunc(legendData.length / 6)
+    const columns = Math.ceil(legendData.length / 6)
```
KevinFCormier, line 93: I don't think this is a correct solution. The number 6 here is used to calculate how many columns of legend items we will have, because `itemsPerRow` along with `orientation` is used to specify there will be 6 legend items per column. In the next line, it already accounts for the remainder items. So your change is just always adding 1 more column into the calculation.
```
<ChartLegend
          labelComponent={<LegendLabel />}
          data={legendData}
          itemsPerRow={6}
          rowGutter={-5}
          gutter={10}
          symbolSpacer={8}
          height={150}
          width={legendWidth}
          orientation={'vertical'}
          y={offset}
        />
```

The `legendWidth` is not used as a literal number of pixels, but it is used relative to the `height` property to compute the aspect ratio. When we have fewer than 6 items, we are saying the legend should be square, which unfortunately does not use the entire space on the card.

With your changes, I tried adding some things to the legend so that we would have 2 columns:
image

So now we calculate that there should be 3 columns, and since these labels are not long, we have wasted space.
image

And if we have long text on the labels, it just pushes the 2nd column totally off the card:
image

So this is a tricky problem. We may need to see if it is possible to limit the width of labels and provide a tooltip if there is overflow, or to somehow calculate the actual width of the labels and take this into account for the aspect ration of the legend.

FYI @zlayne ... not sure if you have any ideas here.

It might be useful to play with some of the examples here:
https://v5-archive.patternfly.org/charts/legends

zlayne: As you mentioned, there are many variables that make displaying the legend very tricky. Also, the legend component is limiting as (if I remember correctly) it only accepts strings, so I'm not sure how viable the tooltip would be.

I think I chose 150 px as the default width of a column as it seemed to fit most legend item widths. Certainly not exact and is likely causing the string cutoff you are trying to fix. 

Due to the legend using SVG as the underlying element, it gets tricky with sizing and responsive design. It could be worth looking into creating a custom legend component

KevinFCormier: I noticed we have our own `LegendLabel` wrapper... so might be able to do something with tooltips/truncation there.

KevinFCormier:

KevinFCormier:

PR #4270, opened 2025-02-13
Standardize timestamp formatting using AcmTimestamp component across console UI

vishsanghishetty:
## Description
This PR is for [ticket](https://issues.redhat.com/browse/ACM-17795) 

## Acceptance Criteria
- [x] Replace existing `moment` JS timestamp implementations with ACM's standard timestamp `AcmTimestamp` component
- [x] Uninstallation of the `moment` and `moment-timezone`
- [x] Tests are configured to run in the UTC timezone via the `package.json` script
- [x] Standardize timestamp formatting using the AcmTimestamp component across the console UI


frontend/src/routes/Applications/ApplicationDetails/ApplicationOverview/ApplicationOverview.test.tsx
```diff
@@ -230,7 +231,7 @@ const mockApplicationDataSubscription: ApplicationDataType = {
               'magchen-sibling-ns/magchen-sibling-subscription-1-local',
             'apps.open-cluster-management.io/reconcile-option': 'merge',
           },
-          creationTimestamp: '2022-03-01T21:30:02Z',
+          creationTimestamp: getShortDateTime('2022-03-01T21:30:02Z'),
```
KevinFCormier, line 234: This change doesn't seem to be needed, and doesn't make sense to me. The code should call this function, not the test data.

frontend/src/routes/Applications/ApplicationDetails/ApplicationOverview/ApplicationOverview.test.tsx
```diff
@@ -570,7 +571,7 @@ describe('Overview Tab', () => {
     await waitForText('Clusters')
     await waitForText('None')
     // created
-    await waitForText('Mar 1 2022, 9:30 pm')
+    await screen.findByText(/Mar 1, 2022, \d{1,2}:\d{2} (AM|PM)/)
```
KevinFCormier, line 574: You needed to accept any time here because the way you set the timezone to UTC is not working.

I found that if I changed the npm test scripts to set the timezone there, I could restore this file to original and it worked:
image

frontend/src/routes/Applications/helpers/resource-helper.tsx
```diff
@@ -351,13 +351,10 @@ export const getResourceLabel = (type: string, count: number, t: TFunction) => {
   return label + optionalCount
 }
 
-export const getAge = (item: IResource, locale: string, timestampKey: string) => {
+export const getAge = (item: IResource, _locale: string, timestampKey: string) => {
```
KevinFCormier, line 354: The name of this function no longer describes what it returns.

frontend/src/routes/Applications/helpers/resource-helper.tsx
```diff
@@ -403,24 +400,8 @@ export const getEditLink = (params: {
...
-  }
-  const date = getMoment(timestamp)
-  if (date.isSame(now, 'day')) {
-    return date.format(timeFormat)
-  } else if (date.isSame(now, 'year')) {
-    return date.format(`${monthDayFormat}, ${timeFormat}`)
-  } else {
-    return date.format(`${monthDayFormat} ${yearFormat}, ${timeFormat}`)
-  }
+export const getShortDateTime = (timestamp: string) => {
```
KevinFCormier, line 403: The name of this function also no longer describes what it returns.

KevinFCormier (changes requested):
I left a few comments on the code. I also noticed that when you look at the applications list, it is very crowded, and the Created column dates all wrap to take up two lines. I think we might need to adjust column widths and/or delay this to next release and introduce column management for this table.

frontend/src/routes/Applications/ApplicationDetails/ApplicationOverview/ApplicationOverview.test.tsx
vishsanghishetty: @KevinFCormier, I updated the npm scripts in frontend `package.json` with the above and restored the test file to original, this test doesn't pass on my local. Did I miss anything from what you have said above?

vishsanghishetty: alt text

vishsanghishetty:
> I left a few comments on the code. I also noticed that when you look at the applications list, it is very crowded, and the Created column dates all wrap to take up two lines. I think we might need to adjust column widths and/or delay this to next release and introduce column management for this table.

@KevinFCormier I adjusted the column width and added `nowrap` to prevent the date from wrapping. Please let me know what you think.

frontend/src/routes/Applications/ApplicationDetails/ApplicationOverview/ApplicationOverview.tsx
```diff
@@ -289,15 +293,19 @@ export function ApplicationOverviewPageContent() {
...
         },
         {
           key: t('Last sync requested'),
           value: (
             <Fragment>
-              {renderData(getShortDateTime(lastSynced), getShortDateTime(lastSynced), '30%')}
+              {renderData(lastSynced, lastSynced ? <AcmTimestamp timestamp={lastSynced} /> : '-', '30%')}
               {renderData(
-                getShortDateTime(lastSynced),
+                lastSynced,
```
KevinFCormier, line 308: I was a bit concerned here that we're passing in something different to what was being passed in before. But if you look at the `renderData` implementation, it is checking that the first argument is not `-1`, and displaying the second argument, otherwise showing a skeleton with the width prescribed by the third argument.

FYI @fxiang1  I don't think this is currently working correctly, and I'm not sure why we would hide the Sync button if there has not been a previous sync, so I'm not clear on the intent. It looks like it has been this way for a long time.

I think there might also be a problem with showing the application repos for application sets, where it looks like we are just rendering `true`?
https://github.com/stolostron/console/blame/main/frontend/src/routes/Applications/ApplicationDetails/ApplicationOverview/ApplicationOverview.tsx#L351

frontend/src/routes/Applications/Overview.test.tsx
```diff
@@ -138,9 +138,8 @@ describe('Applications Page', () => {
     await waitForText('feng-remote-argo8')
     expect(screen.getByText(SubscriptionKind)).toBeTruthy()
     expect(screen.getByText(mockApplication0.metadata.namespace!)).toBeTruthy()
-    expect(screen.getAllByText('Local')).toBeTruthy()
```
KevinFCormier, line 141: No need to remove this.

frontend/src/routes/Applications/Overview.test.tsx
```diff
@@ -138,9 +138,8 @@ describe('Applications Page', () => {
...
     expect(screen.getAllByText('Git')).toBeTruthy()
-    expect(screen.getByText('a few seconds ago')).toBeTruthy()
+    expect(screen.getByText(/^[A-Z][a-z]{2} \d{1,2}, \d{4}, \d{1,2}:\d{2} (AM|PM)$/)).toBeTruthy()
```
KevinFCormier, line 142: Since we're no longer showing strings like "a few seconds ago", there is no need to create a recent timestamp in the test data. It would be better to just hardcode a date and then confirm it shows up here formatted correctly.

frontend/src/routes/Applications/Overview.tsx
```diff
@@ -629,8 +629,13 @@ export default function ApplicationsOverview() {
       {
         header: t('Created'),
         cell: (resource) => {
-          return <span>{getAge(resource, '', 'metadata.creationTimestamp')}</span>
+          return (
+            <span style={{ whiteSpace: 'nowrap' }}>
+              {getResourceTimestamp(resource, '', 'metadata.creationTimestamp')}
+            </span>
+          )
```
KevinFCormier, line 636: Good. I'm happy with how this looks.

KevinFCormier (changes requested):
I like the adjustments to the Applications table. A few more items to consider.

package.json
KevinFCormier, line 1: Please revert this change

package-lock.json
KevinFCormier, line 1: Please revert this file

frontend/src/routes/Applications/ApplicationDetails/ApplicationTopology/helpers/diagram-helpers.js
```diff
@@ -11,26 +11,22 @@
...
-    if (value.includes('T')) {
-      return moment(value, 'YYYY-MM-DDTHH:mm:ssZ').fromNow()
-    } else {
-      return moment(value, 'YYYY-MM-DD HH:mm:ss').fromNow()
-    }
+  if (!value) {
+    return '-'
   }
-  return '-'
+  return <AcmTimestamp timestamp={value} />
```
KevinFCormier, line 29: Seems unneeded since the component now renders '-' itself if the timestamp is not truthy.

frontend/src/routes/Credentials/CredentialsPage.tsx
```diff
@@ -231,11 +231,7 @@ export function CredentialsTable(props: {
             sort: 'metadata.creationTimestamp',
             cell: (resource) => (
               <span style={{ whiteSpace: 'nowrap' }}>
-                {resource.metadata?.creationTimestamp ? (
-                  <AcmTimestamp timestamp={resource.metadata.creationTimestamp} />
-                ) : (
-                  '—'
-                )}
+                <AcmTimestamp timestamp={resource.metadata?.creationTimestamp ?? ''} />
```
KevinFCormier, line 234:
```suggestion
                <AcmTimestamp timestamp={resource.metadata?.creationTimestamp} />
```
No need for the nullish coalescing operator here... timestamp is permitted to be undefined now.

frontend/src/routes/Governance/components/AutomationDetailsSidebar.tsx
```diff
@@ -98,12 +98,16 @@ export function AutomationDetailsSidebar(props: {
...
+            started: jobResult?.started ? (
+              <AcmTimestamp timestamp={jobResult.started} />
+            ) : (
+              <AcmTimestamp timestamp={ansibleResultCondition?.lastTransitionTime ?? ''} />
+            ),
+            finished: jobResult?.finished ? (
+              <AcmTimestamp timestamp={jobResult.finished} />
+            ) : (
+              <AcmTimestamp timestamp={ansibleResultCondition?.lastTransitionTime ?? ''} />
+            ),
```
KevinFCormier, line 110:
```suggestion
            started: <AcmTimestamp timestamp={jobResult?.started ?? ansibleResultCondition?.lastTransitionTime} />,
            finished: <AcmTimestamp timestamp={jobResult?.finished ?? ansibleResultCondition?.lastTransitionTime} />,
```

KevinFCormier (changes requested):
This is looking quite good. Given the scope of the changes, I think I'd be more comfortable committing this for the next release only, and requesting deferral on the associated defect. There are a few more tables that might need adjustment, for example.

It's great you were able to fully remove `moment`. Please add to this PR the uninstallation of the `moment` and `moment-timezone` packages from the `frontend/package.json`.

Thanks!

PR #4289, opened 2025-02-25
ACM-17820: Policy API Group string value optional

Randy424:
Regarding: https://issues.redhat.com/browse/ACM-17810

Draft PR until react-form-wizard change is merged and published.

### DIFF
--- a/frontend/src/wizards/Governance/Policy/PolicyWizard.tsx
+++ b/frontend/src/wizards/Governance/Policy/PolicyWizard.tsx
@@ -455,12 +455,14 @@ export function PolicyWizardTemplates(props: { policies: IResource[] }) {
             path={`${selectorPath}.include`}
             label={t('Include namespaces')}
             placeholder={t('Add namespace')}
+            required
           />
           <WizStringsInput
             id="exclude-namespaces"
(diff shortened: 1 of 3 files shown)


frontend/src/wizards/Governance/Policy/PolicyWizard.tsx
```diff
@@ -853,7 +853,7 @@ function ObjectTemplate() {
           placeholder={t('Add rule')}
           collapsedPlaceholder={t('Expand to edit')}
         >
-          <WizStringsInput label={t('API Groups')} path="apiGroups" />
+          <WizStringsInput required={false} label={t('API Groups')} path="apiGroups" />
           <WizStringsInput label={t('Resources')} path="resources" />
           <WizStringsInput label={t('Verbs')} path="verbs" />
```
KevinFCormier, line 858: To maintain existing behaviour for every other `WizStringsInput` instance, I think you'll need to add the `required` prop? I think there should be no need to add it on line 856 now, as it will not be required by default.

Randy424: Ack. Brain glitch.

PR #4290, opened 2025-02-26
Update base image to match downstream

KevinFCormier:
Update to match downstream version:
https://gitlab.cee.redhat.com/acm-cpaas-midstream/console/-/blob/rhacm-2.13-rhel-9/distgit/containers/console/Dockerfile.in?ref_type=heads#L43


Randy424 (approved):
/lgtm

PR #4293, opened 2025-02-27
Bump AI Lib to 2.14.3-cim

jgyselov:


rawagner (approved):

KevinFCormier:
@jgyselov I'm not sure what the problem is, but the tests also hang when you run them locally (`npm test` from `console/frontend` directory) with the changes in this PR. And there are some failed tests as well.

PR #4296, opened 2025-02-27
Catch request aborted errors from effect cleanup

KevinFCormier:
For the development console, attempting to edit an Automation Template results in an error being shown:
image

This has to do with React 18 safe mode mounting cycles, and effect clean-ups not being handled properly.


vishsanghishetty:
Verified, works as expected. thanks for the fix, Kevin!

/lgtm

PR #4297, opened 2025-02-27
ACM-18149 Preserve most user changes to auto-import-secret

KevinFCormier:
https://issues.redhat.com/browse/ACM-18149


vishsanghishetty:
Verified: The annotation entered in the YAML editor is preserved in auto-import-secret

/lgtm

PR #4307, opened 2025-03-04
ACM-20968: Decouple InfraEnv Create Page

rawagner:
The code that remains in the ACM Console is just to expose CIM components. Once CIM runs as dynamic plugin, the remaining code will be migrated to CIM app. (which esentially will be just definitions in `console-extensions.json` )


KevinFCormier:
@rawagner Can this be closed?

PR #4308, opened 2025-03-04
[ACM-17420] Add ACM Searchbar to VM page for filtering

zlayne:
This PR adds the ACM searchbar to the Virtual Machines page. Allowing users to filter VM's against any VM property saved in search.

Screenshot 2025-03-12 at 10 19 29 AM

Search-API PR to include query params on search schema query. This limits the search suggestions to only options relevant to the current search. https://github.com/stolostron/search-v2-api/pull/334


KevinFCormier:
Looks great, @zlayne! Since the search is live, I wonder if we should hide the submit button on this page?

zlayne:
@KevinFCormier Great point! I plan to meet with Joy when she is back to iron out some of the inconsistencies between searchbar usage on this page and search.

PR #4309, opened 2025-03-04
Acm 17869 poll argo resources instead of watch in console backend

jeswanke:
1. in events.ts, appset has a new isPolled attrbute:
 { kind: 'ApplicationSet', apiVersion: 'argoproj.io/v1alpha1', **isPolled: true** },
2. if listAndWatch finds isPolled, rather then putting a kube watch on the appset list, it will poll the list every 30 seconds
3. since appset is removed from recoil states, the backend:
 a. **appSetRelatedResources** in the frontend requires all appsets--so rather then send all appsets to fontend, the backend will calculate appSetRelatedResources and append to each app and send it to frontend as a '**uidata**' attribute on the appset
 4. in the frontend all appset recoil states are replace with a fetch


frontend/src/routes/Applications/CreateApplication/CreateApplicationArgo.tsx
```diff
@@ -92,10 +92,30 @@ export function CreateApplicationArgo() {
 
   const { cancelForm, submitForm } = useContext(LostChangesContext)
   const [createdResource, setCreatedResource] = useState<any>()
+  const [applicationSets, setApplicationSets] = useState<ApplicationSet[]>()
+  const [loadingAppSets, setLoadingAppSets] = useState(true)
+
+  // instead of burdoning recoil with appsets, use old fashioned fetch
+  // opening wizard may take longer, but argo wizards are probably seldom used
```
KevinFCormier, line 99: Looking at how the application sets are used by the wizard, it seems the only purpose is to pre-populate the Git and Helm URLs in the dropdowns. In an environment with thousands of application sets, I think it is likely that there will be far fewer unique sources, so it would be more efficient to calculate the unique repoURLs from the cached data on the backend, and then just send that information to the UI.

This is the sort of thing I had hoped would be investigated as part of the spike [ACM-17870](https://issues.redhat.com/browse/ACM-17870), which was closed without any findings.

KevinFCormier (changes requested):
I have not tested, but from reviewing the code, I think there could be an issue after you create an application set and get redirected to the overview page. Any ideas how to address this?

frontend/src/routes/Applications/CreateApplication/CreateApplicationArgo.tsx
jeswanke: Yes--it's likely there are fewer unique sources--but I think it's even likelier that if there are thousands of appsets that they're being created by the cli and not this wizard, and that my awesome work adding a new backend api would go unappreciated :(

backend/src/routes/events.ts
```diff
@@ -235,14 +235,20 @@ interface IWatchOptions {
   kind: string
   labelSelector?: Record<string, string>
   fieldSelector?: Record<string, string>
+  isPolled?: boolean
```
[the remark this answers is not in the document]
jeswanke: added a 'notStreamed' along with isPolled
commented both

frontend/src/routes/Applications/CreateApplication/CreateApplicationArgo.tsx
```diff
@@ -92,10 +92,30 @@ export function CreateApplicationArgo() {
...
+        setLoadingAppSets(false)
+      }
+    }
+    fetchAppSets()
+  }, [])
 
   // don't navigate to details page until application exists in recoil
   useEffect(() => {
-    if (createdResource) {
+    if (createdResource && applicationSets) {
```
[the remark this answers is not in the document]
jeswanke: absolutely right:
1. now the details are openned immediately
2. the details use a fetch to get the new appset
3. and if they're looking for this resource in the app list, events.ts polls faster when there are less appsets

backend/src/routes/aggregators/utils.ts
```diff
@@ -491,18 +538,27 @@ interface ResultType {
...
+  // convert path into key array
+  const keys: string[] = []
+  path.replace(keyEx, function (match, number, quote, subString) {
+    keys.push(
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      quote && typeof subString === 'string' ? subString.replace(/\\(\\)?/g, '$1') : number || match
+    )
+    return ''
+  })
+
```
KevinFCormier, line 553: I only noticed the `get` function being called with known paths. Could we simply change it to allow either a dot-separated string path, or an array? Also, I didn't notice any use of the `dflt` param... did you add it but forget to use it?

frontend/src/routes/Applications/CreateApplication/CreateApplicationArgo.tsx
KevinFCormier: OK, so we'll consider this a separate problem.

backend/src/routes/aggregators/utils.ts
KevinFCormier: @jeswanke I see where you were using the `dflt` param now. Also, I appreciate that you added an explanation for what this code does.

However, I still feel like the use of a very complicated regex that updates a string in place and uses side effects to capture each path segment seems like a bit of a "weird flex bro" moment. Correct me if I'm wrong, but the only paths I see used with this are hard-coded. Either simple ones with only dots, or we have `clusterDecisionResource.labelSelector.matchLabels["cluster.open-cluster-management.io/placement"]` and `metadata.ownerReferences[0].name`. It seems like a waste to run this regex on them every time, when we could simply express them as an array in the first place:

```
const appSetPlacementStr = [
  "clusterDecisionResource",
  "labelSelector",
  "matchLabels",
  "cluster.open-cluster-management.io/placement"
]
```

We already have the [get-value](https://www.npmjs.com/package/get-value) package installed in the backend, so you could use that. It accepts an array format, or you could express as a string (but note that the format is a little different... I don't think it accepts square brackets).

What do you think?

jeswanke: ah yeah you're right! get-value is suppose to be fatser then lodash and I guess the reason is it's not running regex-- but you do have to meet it halfway by changing your selector into an array

KevinFCormier:
@jeswanke Thanks for fixing the topology view. I tested both push and pull model. Seems to be working well. Can you please take a look at the Sonar findings? There are several little things we can clean up.

jeswanke:
@KevinFCormier -- I made the changes

PR #4319, opened 2025-03-07
Make "Allowed Cluster Service Versions" optional

JustinKuli:
It was mistakenly marked as required, which will prevent users from being able to create what would be a perfectly valid OperatorPolicy.

Refs:
 - https://issues.redhat.com/browse/ACM-18541


dhaiducek:
@JustinKuli Would you revert the other updates from the https://github.com/stolostron/console/pull/4289 PR? It seems the `namespaceSelector` is affected also:
image

dhaiducek:
Looks to me like there's some nuance to the fix in place where "required" means that the input must be populated with non-empty strings while `required=false` means that the input can be nothing or anything (including empty strings)...

JustinKuli:
Good idea @dhaiducek

dhaiducek (approved):
LGTM! Thanks!

PR #4321, opened 2025-03-07
ACM-18453-ClusterSets-not-included-in-First-Events-Packet

jeswanke:
oh man--I got pr4321!!!!

I used to just check in a couple spots whether or not recoil states were fully loaded like this:
```
  const { dataContext } = useContext(PluginContext)
  const { loadCompleted } = useContext(dataContext)
  return loadCompleted ? (
    <PageSection>
      <Card>
        <CardBody>
          <ErrorState {...props} />
        </CardBody>
      </Card>
    </PageSection>
  ) : (
    <LoadingPage />
  )

```
but there'll always be something that gets through--instead I created a list of pages approved for fast load, everything else gets a loading page:

```
      // Credentials
      NavigationPath.credentials,
    ] as string[]
  ).includes(location.pathname)
  useEffect(() => {
    if (!loadStarted) {
      load()
    }
  }, [load, loadStarted])

  return (fastLoadPage && loadStarted) || loadCompleted ? (

```

### DIFF
--- a/frontend/src/components/LoadPluginData.tsx
+++ b/frontend/src/components/LoadPluginData.tsx
@@ -40,12 +40,7 @@ export const LoadPluginData = (props: { children?: ReactNode }) => {
       NavigationPath.advancedConfiguration,
       NavigationPath.createApplicationArgo,
       NavigationPath.createApplicationArgoPullModel,
-      NavigationPath.editApplicationArgo,
       NavigationPath.createApplicationSubscription,
-      NavigationPath.applicationDetails,
-      NavigationPath.applicationOverview,
-      NavigationPath.applicationTopology,
-      NavigationPath.editApplicationSubscription,
 
       // Governance
       NavigationPath.governance,
@@ -55,7 +50,7 @@ export const LoadPluginData = (props: { children?: ReactNode }) => {
       // Credentials
       NavigationPath.credentials,
     ] as string[]
-  ).includes(location.pathname)
+  ).includes(location.pathname.replace(/\/$/, ''))
   useEffect(() => {
     if (!loadStarted) {
       load()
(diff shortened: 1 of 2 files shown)


frontend/src/components/LoadPluginData.tsx
```diff
@@ -3,15 +3,68 @@ import { ReactNode, useContext, useEffect } from 'react'
...
+      NavigationPath.resourceYAML,
+      NavigationPath.resourceRelated,
+      NavigationPath.resourceLogs,
+
+      // Infrastructure
+      NavigationPath.infrastructure,
+
+      // Infrastructure - Clusters - Managed Clusters
+      NavigationPath.clusters,
+      NavigationPath.managedClusters,
```
KevinFCormier, line 35: I found `fastLoadPage` was true for `/multicloud/infrastructure/clusters/managed` but not `/multicloud/infrastructure/clusters/managed/`.

frontend/src/components/LoadPluginData.tsx
```diff
@@ -3,15 +3,68 @@ import { ReactNode, useContext, useEffect } from 'react'
...
+
+      // Infrastructure - Clusters - Cluster Sets
+      NavigationPath.clusterSets,
+      NavigationPath.applications,
+      NavigationPath.advancedConfiguration,
+      NavigationPath.createApplicationArgo,
+      NavigationPath.createApplicationArgoPullModel,
+      NavigationPath.editApplicationArgo,
```
KevinFCormier, line 43: Path contains params... doesn't work for `/multicloud/applications/edit/argo/openshift-gitops/appset-perf-4998` for example

KevinFCormier (commented):
I'm not sure about this mechanism. See inline comments. Did you really go through and verify that all those routes work properly when loaded as the initial page?

PR #4322, opened 2025-03-07
ACM-18631 Remove namespace from channel tooltip

davidtluong:
Resolves https://issues.redhat.com/browse/ACM-18631, but looks like the translations will still need to be fixed


KevinFCormier:
@davidtluong Thank you for your PR. Can you please remove the changes to the non-English translation files? We have translation services that will provide the translations for us at the end of the next release.

PR #4328, opened 2025-03-12
ACM-18820 Use spec.name instead of metadata.name for operator check

KevinFCormier:
Name of the subscription could vary depending on how subscription is installed on clusters (i.e. using custom automation rather than OperatorHub UI install).

https://issues.redhat.com/browse/ACM-18820


nitin-dhevar:
lgtm

PR #4346, opened 2025-03-17
In late binding flow, all options for master node number are choosable, but only 3 master are allowed

jgyselov:
https://issues.redhat.com/browse/MGMT-20078


jgyselov:
Duplicate

PR #4408, opened 2025-04-15
ACM-19511 Ansible Automation template next button disabling

vishsanghishetty:
## Description
This PR is for bug [ticket](https://issues.redhat.com/browse/ACM-19511) 

## Acceptance Criteria
- [x] Automation Template Wizard Next Button Disabling
- [x] Unit test - should disable the Next button when an invalid credential is selected

### DIFF
--- a/frontend/src/components/AcmDataForm.tsx
+++ b/frontend/src/components/AcmDataForm.tsx
@@ -124,7 +124,8 @@ export function AcmDataFormPage(props: AcmDataFormProps): JSX.Element {
 
   const { editorTitle, schema, secrets, immutables, formData, globalWizardAlert, hideYaml, isModalWizard } = props
   const [showFormErrors, setShowFormErrors] = useState(false)
-  const showErrors = showFormErrors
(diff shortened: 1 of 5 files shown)


frontend/src/components/AcmDataForm.tsx
```diff
@@ -124,6 +124,7 @@ export function AcmDataFormPage(props: AcmDataFormProps): JSX.Element {
 
   const { editorTitle, schema, secrets, immutables, formData, globalWizardAlert, hideYaml, isModalWizard } = props
   const [showFormErrors, setShowFormErrors] = useState(false)
+  const showErrors = props.mode === 'wizard' ? showFormErrors : formData.showErrors ?? false
```
KevinFCormier, line 127: I don't think this is correct. `formData.showErrors` will only ever be true for your one use-case with Ansible credential validation. The idea is to force the showing of form errors when this happens, but the existing mechanism that uses `showFormErrors` still needs to work.

PR #4475, opened 2025-05-07
ACM-18521: Remove deprecated PF Table, TableHeader, TableBody

Randy424:
# 📝 Summary

**Ticket Summary (Title):**  
"Remove deprecated PF5 components in AcmTable.tsx"

**Ticket Link:**  
[ACM-18521](https://issues.redhat.com/browse/ACM-18521)

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [x] 🔧 Refactor
- [x] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

### DIFF
--- a/frontend/src/ui-components/AcmTable/AcmTable.test.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmTable.test.tsx
@@ -788,7 +788,7 @@ describe('AcmTable', () => {
     expect(getByLabelText('Current page')).toHaveValue(6)
     expect(setSearch).not.toHaveBeenCalled()
     expect(setSort).not.toHaveBeenCalled()
-    expect(container.querySelector('tbody tr:last-of-type [data-label="First Name"]')).toHaveTextContent('Alyce')
+    expect(container.querySelector('tbody:last-of-type [data-label="First Name"]')).toHaveTextContent('Alyce')
 
     expect(getByLabelText('Reset')).toBeVisible()
     userEvent.click(getByLabelText('Reset'))
@@ -825,7 +825,7 @@ describe('AcmTable', () => {
   })
   test('renders a table with expandable rows', async () => {
     const expandedDeleteAction = jest.fn()
-    const { getAllByLabelText, getByRole, getByTestId, getByText } = render(
+    const { getByTestId } = render(
       <MemoryRouter>
         <AcmTable<IExampleData>
           emptyState={<AcmEmptyState title="No addresses found" message="You do not have any addresses yet" />}
@@ -901,20 +901,6 @@ describe('AcmTable', () => {
     )
     userEvent.click(getByTestId('expandable-toggle0'))
     expect(getByTestId('expanded')).toBeInTheDocument()
-
-    // Run delete action for code coverage (no delete support on expanded content)
-    userEvent.click(getAllByLabelText('Actions')[1])
-    await waitFor(() => expect(getByRole('menu')).toBeVisible())
-    expect(getByText('Delete item')).toBeVisible()
-    userEvent.click(getByText('Delete item'))
-    expect(expandedDeleteAction).not.toHaveBeenCalled()
-
-    // Run tooltipped delete action for code coverage (no delete support on expanded content)
-    userEvent.click(getAllByLabelText('Actions')[1])
-    await waitFor(() => expect(getByRole('menu')).toBeVisible())
-    expect(getByText('Delete item tooltip')).toBeVisible()
-    userEvent.click(getByText('Delete item tooltip'))
-    expect(expandedDeleteAction).not.toHaveBeenCalled()
   })
 
   test('renders with filtering and filters options work correctly', async () => {
--- a/frontend/src/ui-components/AcmTable/AcmTable.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmTable.tsx
@@ -1730,7 +1730,7 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
                           {/* include spacing for expandable and selection columns in subrow */}
                           {onCollapse && <Td />}
                           {hasSelectionColumn && <Td />}
-                          <Td key={addedSubRows[rowIndex]?.props?.key} colSpan={columns.length}>
+                          <Td key={addedSubRows[rowIndex]?.props?.key} colSpan={selectedSortedCols.length}>
                             <ExpandableRowContent>
                               {addedSubRows[rowIndex]?.cells?.map((cell) => renderCellContent(cell))}
                             </ExpandableRowContent>
--- a/frontend/src/ui-components/AcmTable/AcmTable.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmTable.tsx
@@ -1,6 +1,7 @@
 /* Copyright Contributors to the Open Cluster Management project */
 
 import { css } from '@emotion/css'
+import { css as cssPF } from '@patternfly/react-styles'
 import {
   Badge,
   ButtonVariant,
@@ -18,6 +19,7 @@ import {
   ToolbarGroup,
   ToolbarItem,
   Tooltip,
+  TooltipPosition,
   TooltipProps,
 } from '@patternfly/react-core'
 import {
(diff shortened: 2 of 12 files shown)


frontend/src/ui-components/AcmTable/AcmTable.test.tsx
```diff
@@ -899,22 +899,8 @@ describe('AcmTable', () => {
...
-    expect(getByText('Delete item')).toBeVisible()
-    userEvent.click(getByText('Delete item'))
-    expect(expandedDeleteAction).not.toHaveBeenCalled()
-
-    // Run tooltipped delete action for code coverage (no delete support on expanded content)
-    userEvent.click(getAllByLabelText('Actions')[1])
-    await waitFor(() => expect(getByRole('menu')).toBeVisible())
-    expect(getByText('Delete item tooltip')).toBeVisible()
-    userEvent.click(getByText('Delete item tooltip'))
-    expect(expandedDeleteAction).not.toHaveBeenCalled()
```
Randy424, line 917: Unnecessary tests. They don't cover real scenarios and were being used for increase coverage.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -557,6 +565,32 @@ const applyFilters = <T, S>(
   )
 }
 
+function mergeProps(...props: any) {
+  const firstProps = props[0]
+  const restProps = props.slice(1)
+
+  if (!restProps.length) {
+    return mergeWith({}, firstProps)
+  }
```
KevinFCormier, line 574: Seems strange to call `mergeWith` with no `customizer` function. From the original code, there is a comment mentioning not to mutate the original... so probably for this case and the one further down, we can just do `{ ...firstProps }`.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -557,6 +565,32 @@ const applyFilters = <T, S>(
...
+      if (a && b) {
+        return cloneElement(a, {
+          children: b,
+        })
+      }
+      return { ...b, ...a }
+    }
+
+    if (key === 'className') {
+      return css(a, b)
```
KevinFCormier, line 587: In your code, this is `css` from `@emotion/css` and in the original code it is from `@patternfly/react-styles` which is a simpler function that just appends the class names. That would be sufficient for this case, rather than generating a new class name with emotion.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1543,47 +1622,128 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+                              isSelected: selected[row.props.key],
+                              isDisabled: row.disableSelection,
+                            }}
+                          />
+                        )}
+                        {row?.cells?.map((cell, cellIndex) => {
+                          // setup row props, including celltransforms
+                          const transforms: ITransform[] | undefined = selectedSortedCols[cellIndex]?.cellTransforms
+                          const iTransformCellProps = transforms
+                            ? mergeProps(
```
KevinFCormier, line 1698: I was surprised that the transforms do not seem to be deprecated, but I did not find examples from PF of using them. I wonder if we should open a follow up task to look at all the transforms we are using and seeing if we can move to the new way of doing them?

KevinFCormier (changes requested):
Good stuff! Everything seems to work as expected from the bit of ad hoc testing I did.

Please see inline comments for a few suggestions and questions.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1543,47 +1622,128 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+                            )}
+                          </Td>
+                        )}
+                      </Tr>
+                      {addedSubRowCount > 0 && (
+                        <Tr isExpanded={row.isOpen} key={addedSubRows[rowIndex]?.props?.key}>
+                          {/* include spacing for expandable and selection columns in subrow */}
+                          {onCollapse && <Td />}
+                          {hasSelectionColumn && <Td />}
+                          <Td key={addedSubRows[rowIndex]?.props?.key} colSpan={9}>
```
KevinFCormier, line 1737: Please do not hard-code `colSpan` here.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -503,10 +509,12 @@ const tableClass = css({
     },
   },
 })
+const rowActionClass = css({
+  'grid-column-start': '2 !important',
+  display: 'block !important',
+})
```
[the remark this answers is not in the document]
Randy424: The existing code uses a grid-column-start and block display to position the kebab menu at the bottom right of the enlarged row item (when the page width crosses the table width breakpoint). This particular styling, FWICT, isn't built into the PF5 <Td> component, and so I installed a custom class and used `!important` to override the default behavior. Alas, it would come as no surprise to find another (probably better) way to handle the styling here. I'll test out your proposed changes.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1543,47 +1622,128 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
+                          const isMenuKebab = isKebabMenu(cell)
+                          const rowProps = {
+                            dataLabel: isMenuKebab ? undefined : columns[cellIndex].header,
+                            ...iTransformCellProps,
+                            className: isMenuKebab ? rowActionClass + ' ' + iTransformCellProps?.className : '',
+                          }
+                          return (
+                            <Td key={`row-${rowIndex}-cell-${cellIndex}`} {...rowProps}>
+                              {renderCellContent(cell)}
+                            </Td>
```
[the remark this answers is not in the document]
Randy424: I came across this setup (with isActionCell) but wanted to try to preserve the original appearance of the table if possible. That said, I think this looks great and I'll take it if we don't mind the change.

frontend/src/ui-components/AcmTable/AcmTable.tsx
Randy424: [I looked into that as well](https://patternfly.slack.com/archives/C4FM977N0/p1744808941706229?thread_ts=1744783266.728319&cid=C4FM977N0). I'm not so sure there is a "new way" as much as there is no specified way of doing this from PatternFly any longer. Even [the example given from this PF engineer](https://patternfly.slack.com/archives/C4FM977N0/p1744808941706229?thread_ts=1744783266.728319&cid=C4FM977N0) does not actually use an itransform type.

To your other point though, I agree that it might be worth iterating on a unified pattern for ACM console's transformation needs. Opened a task here: https://issues.redhat.com/browse/ACM-20735

Randy424:
Apart from the `ActionCol` thread, this is ready for another review.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -1543,47 +1622,128 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
                 borders={!props.noBorders}
                 variant={TableVariant.compact}
                 gridBreakPoint={gridBreakPoint ?? breakpoint}
               >
-                <TableHeader />
-                <TableBody />
+                <Thead>
+                  <Tr>
+                    {onCollapse && <Th screenReaderText="Row collapse" />}
+                    {hasSelectionColumn && <Th screenReaderText="Row select" />}
```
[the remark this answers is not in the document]
KevinFCormier: Remember to run `npm run i18n:fix`

frontend/src/ui-components/AcmTable/AcmTable.tsx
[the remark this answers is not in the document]
KevinFCormier: Yeah, I think the top-right placement is correct for card view. Not sure why we had it at the bottom-right.

frontend/src/ui-components/AcmTable/AcmTable.tsx
KevinFCormier: You need to calculate the correct `colSpan` here based on the number of columns currently displayed. Test with the Policies table, where you can vary the number of columns that are currently used.

image

Randy424: Using the `addedSubRowCount` for the number of sub-rows (sub-columns) in each row, which aligns with this PF guidance as well: https://www.patternfly.org/components/table/#nested-column-headers

Thanks!

Randy424:
@KevinFCormier ready to go.

frontend/src/ui-components/AcmTable/AcmTable.tsx
KevinFCormier: \*\*BUZZER NOISE\*\* Try again 😕

https://github.com/user-attachments/assets/979da3a3-fecc-4d97-b0e4-07d17fb90274

Randy424: Dumb mistake!
Though I'm noticing now that the # of subRows used as a colSpan doesn't quite get us to the behavior of the current table. I'm thinking the # of items in column array might be a better approximation. We also have no great way calculating the number of subRows right now (except maybe by taking advantage of the exportSubrow array, which also seems hacky). We could pass them in, but that also feels a bit too involved.

PR #4477, opened 2025-05-07
Set serviceAccountName for mce builds

KevinFCormier:
Syncing up MCE Tekton config with the ACM version (was missing serviceAccountName) to try to fix pipeline failures.


Randy424 (approved):
/lgtm

PR #4494, opened 2025-05-14
MGMT-19879: Binding-day-2-host-to-cluster-defaults-host-to-worker-role

ElayAharoni:
https://issues.redhat.com/browse/MGMT-19879
removed the default behavior that was setting the role for any added host to be worker.
added the option to edit the host role in the add hosts modal.


KevinFCormier:
@ElayAharoni can this be closed?

PR #4523, opened 2025-05-21
ACM-20296: Column management can be broken by outdated local storage

nitin-dhevar:
# 📝 Summary

**Ticket Summary (Title):**  

ACM-20296: Column management can be broken by outdated local storage

**Ticket Link:**  

[Ticket](https://issues.redhat.com/browse/ACM-20296)

**Type of Change:**  

- [x] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

### DIFF
--- a/frontend/src/routes/Governance/policies/Policies.tsx
+++ b/frontend/src/routes/Governance/policies/Policies.tsx
@@ -195,7 +195,7 @@ export default function PoliciesPage() {
         isFirstVisitChecked: true,
         exportContent: (item) => handlePolicySetCell(item, policySets, true),
       },
-      { ...policyClusterViolationsColumn, id: 'cv', isDefault: true, order: 6 },
+      { ...policyClusterViolationsColumn, id: 'clusterviolations', isDefault: true, order: 6 },
       {
         header: t('Source'),
         sort: (itemA: PolicyTableItem, itemB: PolicyTableItem) => {
--- a/frontend/src/routes/Governance/policies/Policies.tsx
+++ b/frontend/src/routes/Governance/policies/Policies.tsx
@@ -195,7 +195,7 @@ export default function PoliciesPage() {
         isFirstVisitChecked: true,
         exportContent: (item) => handlePolicySetCell(item, policySets, true),
       },
-      { ...policyClusterViolationsColumn, id: 'clusterviolations', isDefault: true, order: 6 },
+      { ...policyClusterViolationsColumn, id: 'cv', isDefault: true, order: 6 },
       {
         header: t('Source'),
         sort: (itemA: PolicyTableItem, itemB: PolicyTableItem) => {
--- a/frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
+++ b/frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
@@ -1,5 +1,5 @@
 /* Copyright Contributors to the Open Cluster Management project */
-import { FormEvent, useEffect, useState } from 'react'
+import { FormEvent, useState } from 'react'
 import {
   Button,
   DataList,
@@ -21,6 +21,7 @@ import {
 import { IAcmTableColumn } from './AcmTable'
 import { useTranslation } from '../../lib/acm-i18next'
 import ColumnsIcon from '@patternfly/react-icons/dist/js/icons/columns-icon'
+import { setColumnValues } from './localColumnStorage'
 
 interface AcmManageColumnProps<T> {
   allCols: IAcmTableColumn<T>[]
@@ -30,7 +31,6 @@ interface AcmManageColumnProps<T> {
   defaultColIds?: string[]
   setColOrderIds: (colOrderIds: string[]) => void
   colOrderIds: string[]
-  tableId?: string
 }
 
 export function AcmManageColumn<T>({
@@ -41,7 +41,6 @@ export function AcmManageColumn<T>({
   setSelectedColIds,
   requiredColIds,
   defaultColIds,
-  tableId,
 }: AcmManageColumnProps<T>) {
   const [isModalOpen, setIsModalOpen] = useState<boolean>(false)
   const { t } = useTranslation()
@@ -54,7 +53,6 @@ export function AcmManageColumn<T>({
     <>
       <ManageColumnModal<T>
         {...{
-          tableId,
           isModalOpen,
           selectedColIds,
           allCols,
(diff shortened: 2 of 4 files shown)


frontend/src/routes/Governance/policies/Policies.tsx
```diff
@@ -195,7 +195,7 @@ export default function PoliciesPage() {
         isFirstVisitChecked: true,
         exportContent: (item) => handlePolicySetCell(item, policySets, true),
       },
-      { ...policyClusterViolationsColumn, id: 'cv', isDefault: true, order: 6 },
+      { ...policyClusterViolationsColumn, id: 'clusterviolations', isDefault: true, order: 6 },
```
KevinFCormier, line 198: Is this change necessary? Won't it break selected columns and ordering that users have saved in their local storage?

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
```diff
@@ -113,10 +123,19 @@ function ManageColumnModal<T>(props: ManageColumnModalProps<T>) {
...
   const [localSelectedIds, setlocalSelectedIds] = useState<string[]>(selectedColIds)
 
+  useEffect(() => {
+    localStorage.setItem(tableId + 'SavedColOrder', JSON.stringify(colOrderIds))
+  }, [tableId, colOrderIds])
+
+  useEffect(() => {
+    localStorage.setItem(tableId + 'SavedCols', JSON.stringify(localSelectedIds))
+  }, [localSelectedIds, tableId])
+
```
KevinFCormier, line 138: There is code in AcmTable that does this which you didn't remove. Do we really need it in both places? If not, which is correct?

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -745,12 +745,15 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
     [columns]
   )
-  const localSavedCols = JSON.parse(localStorage.getItem(id + 'SavedCols')!)
-  const localSavedColOrder = JSON.parse(localStorage.getItem(id + 'SavedColOrder')!)
-  const [colOrderIds, setColOrderIds] = useState<string[]>(localSavedColOrder || defaultOrderIds)
+  const localSavedCols = JSON.parse(localStorage.getItem(id + 'SavedCols') || '[]')
+  const localSavedColOrder = JSON.parse(localStorage.getItem(id + 'SavedColOrder') || '[]')
+  const [colOrderIds, setColOrderIds] = useState<string[]>(
+    localSavedColOrder?.length > 0 ? localSavedColOrder : defaultOrderIds
+  )
```
KevinFCormier, line 752: There are utility methods in this file that handle exceptions in JSON parsing (`setLocalStorage` and `getLocalStorage`)... I'm not sure why those were used initially, but could we make use of them here instead?

When we work with localStorage, we should always treat it as untrusted data and code defensively. I am able to crash the page by putting something that is not JSON into the local storage values.

KevinFCormier (changes requested):
Hey Nitin,

Good start - I see that the Manage Columns dialog now contains all the columns again.

Please see my inline comments for other changes.

I would also like you to consider more significant changes. There are a couple of additional problems.

1. I can remove a column that has the `isDefault` property set to `true` by modifying local storage. This should not be possible.

2. Suppose in a new release of ACM we are introducing new columns to a table, and some of these columns have `isFirstVisitChecked` set to `true`. If this was a fresh install, these columns would be included, but for an upgrading user, they already have a saved column selection, so this column will not be added. But since we wanted users to see this column by default, we should probably have it show up so they'll know about it. 

Implementing (2) will require a change in the JSON structure that we store, because we need to know if the user has explicitly deselected the column, or if this column just didn't exist the last time they viewed this table, and we can't represent that in an array.

If we implement (2), then ideally we should check if the current values are in the old format (a simple array) and respect the selected columns and column order to the extent we can. 

I would like you to address the inline comments and item (1) in this PR. Think on item (2) and if you want to pursue that we can do that in a second PR.

Thanks!

frontend/src/routes/Governance/policies/Policies.tsx
nitin-dhevar: I was using the header values earlier instead of the id's, so needed this change for that, but I have fixed it now.

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
nitin-dhevar: I have removed the one from AcmTable, since we were not setting the values anywhere in it, we send the setter method to AcmManageColumn, and the update takes place in it, so I think that is the correct place for this useEffect.

frontend/src/ui-components/AcmTable/AcmTable.tsx
```diff
@@ -745,12 +745,18 @@ export function AcmTable<T>(props: AcmTableProps<T>) {
...
-  const [colOrderIds, setColOrderIds] = useState<string[]>(localSavedColOrder || defaultOrderIds)
+  const localSavedCols = getLocalStorage(id + 'SavedCols', [])
+  const localSavedColOrder = getLocalStorage(id + 'SavedColOrder', [])
+  const [colOrderIds, setColOrderIds] = useState<string[]>(
+    localSavedColOrder?.length > 0 ? localSavedColOrder : defaultOrderIds
+  )
   const [selectedColIds, setSelectedColIds] = useState<string[]>(
-    localSavedCols || [...requiredColIds, ...defaultColIds]
+    localSavedCols?.length > 0
+      ? [...requiredColIds, ...localSavedCols.filter((val: string) => !requiredColIds.includes(val))]
```
nitin-dhevar, line 755: Adding the [...requiredColIds] here, to cover the first point you mentioned about preventing deletion of values whose isDefault property is set to true.

frontend/src/ui-components/AcmTable/AcmManageColumn.tsx
KevinFCormier: I disagree, because it was working before, and because the table needs to read the local storage every time it loads, whether the manage column dialog is being used. So if we centralize local storage access in AcmTable.tsx, that file can be the owner of that particular local storage item (both read and write are centralized).  The keys "SavedCols" and "SavedColOrder" should probably be defined once as contants.

frontend/src/ui-components/AcmTable/AcmTable.tsx
KevinFCormier: There is still a case where I can remove a required column by deleting it from the saved column order.

This is another argument for centralizing the local storage handling in one place. I think the issue is that AcmManageColumn fixes the column order and writes it to local storage, but AcmTable won't see that until the next time it loads. Instead, AcmTable should fix up saved columns and orders when it reads from local storage, so that it can display correctly on the first try and pass correct information to AcmManageColumn.

https://github.com/user-attachments/assets/6d9f4801-99f7-49d3-8dae-c5d9ecf19898

KevinFCormier (commented):
Some more thoughts on read/write from local storage... see inline comments. Thanks.

PR #4578, opened 2025-06-06
ACM-21258-Edit-namespaces-bindings-not-aligned

jeswanke:


frontend/src/components/AcmSelectBase.tsx
```diff
@@ -182,11 +182,14 @@ export function AcmSelectBase(props: AcmSelectBaseProps) {
   let placeholder = props.placeholderText
   if (!placeholder) {
     if (isMulti && selections.length > 0) {
-      placeholder = selections.join(', ')
+      placeholder =
+        variant !== SelectVariant.typeaheadMulti ? selections.join(', ') : props.placeholder ?? t('Select choices')
```
KevinFCormier, line 186: 2 questions:
- why do we have both placeholderText and placeholder props?
- why do we use `t('Select choices')` here and `''` on line 197 when `props.placeholder` is nullish? (I would prefer not to introduce a default string here and just supply the relevant placeholder if you noticed we were missing one somewhere... we could make the prop required)

frontend/src/components/AcmSelectBase.tsx
```diff
@@ -564,69 +570,78 @@ export function AcmSelectBase(props: AcmSelectBaseProps) {
...
+                  }}
+                >
+                  {getMultiTypeaheadChildren(selection)}
+                </Chip>
+              ))}
+            </ChipGroup>
+          )}
+        </TextInputGroupMain>
+        <TextInputGroupUtilities {...getClearBtnDisplayStyle()}>
+          <Button variant="plain" onClick={() => onClearSelection()} aria-label="Clear input value">
```
KevinFCormier, line 638: Untranslated string

frontend/src/components/AcmSelectBase.tsx
```diff
@@ -564,69 +570,78 @@ export function AcmSelectBase(props: AcmSelectBaseProps) {
...
+          placeholder={placeholder}
+          {...(activeItemId && { 'aria-activedescendant': activeItemId })}
+          role="combobox"
+          aria-label={ariaLabel}
+          aria-labelledby={ariaLabelledBy}
+          isExpanded={isOpen}
+          aria-controls="select-multi-typeahead-listbox"
+        >
+          {variant === SelectVariant.typeaheadMulti && (
+            <ChipGroup aria-label="Current selections">
```
KevinFCormier, line 622: Untranslated string

frontend/src/components/AcmSelectBase.tsx
```diff
@@ -564,69 +570,78 @@ export function AcmSelectBase(props: AcmSelectBaseProps) {
...
+          {variant === SelectVariant.single ? renderSinglePlaceholder() : <span>{placeholder}</span>}
+        </div>
+        <TextInputGroupUtilities {...getClearBtnDisplayStyle()}>
+          <Button
+            variant="plain"
+            onClick={(e) => {
+              onClearSelection()
+              e.stopPropagation()
+            }}
+            aria-label="Clear input value"
```
KevinFCormier, line 592: Untranslated string

KevinFCormier (changes requested):
I reviewed the Sonar findings and accepted the ones about complexity and roles, but the remaining two about order of select cases and not using an index as a React key should be easy to fix.

Please see also some inline comments.

Thanks

PR #4645, opened 2025-06-19
Standalone multicluster SDK

KevinFCormier:
Contributions from @upalatucci to move the implementations into the SDK itself.


frontend/packages/multicluster-sdk/src/components/FleetResourceEventStream/EventStream.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const messageCount =
+    count < MAX_MESSAGES
+      ? t('public~Showing {{count}} event', { count })
+      : t('public~Showing most recent {{count}} event', { count })
+
+  const toggleStream = () => {
+    setActive((prev) => !prev)
+  }
+
+  if (!resource.cluster || resource.cluster === hubCluster) return <ResourceEventStream resource={resource} />
```
edewit, line 178: Am I wrong to think that this will give a stack overflow? When cluster equals the `hubCluster` it calls itself with the same resource so that doesn't change the logic and it will call itself again

PR #4715, opened 2025-07-02
ACM-20341: useFleetPrometheusPoll but we want to query across all clusters

upalatucci:
When the cluster  is `undefined` but we want to query across all clusters we need a specific param for that


KevinFCormier:
FYI @jeswanke

PR #4733, opened 2025-07-07
ACM-20341: fix nullResource on useFleetK8sWatch

upalatucci:
`nullResource` is not working propertly as `resource` is always `{}` when `initResource` is `undefined` or `null`

Doing so means that the condition `if(nullResource)` will be `false` even tho should be `true`


frontend/packages/multicluster-sdk/src/api/use-fleet-k8s-watch-resource/use-fleet-k8s-watch-resource.ts
```diff
@@ -36,7 +36,9 @@ export const useFleetK8sWatchResource = <R extends FleetK8sResourceCommon | Flee
 ): WatchK8sResult<R> | [undefined, boolean, any] => {
   const { cluster, ...resource } = initResource ?? {}
   const useFleet = cluster && cluster !== hubClusterName
-  const nullResource = !(resource?.groupVersionKind ?? resource)
+  const nullResource =
+    initResource === null || initResource === undefined || !initResource || !resource?.groupVersionKind
+
```
KevinFCormier, line 41:
```suggestion
  const nullResource = !initResource || !resource?.groupVersionKind
```
@upalatucci could we shorten to this?

upalatucci: yes

PR #4760, opened 2025-07-14
ACM-20341: change advancedSearch option type

upalatucci:
We should use an array instead of object because we can use multiple times the same key to have a strinct query like

`created:>DATE_FROM created:<DATE_TO` This type of query cannot be achieved with an object argument as the key `created` cannot be populated multiple times

### DIFF
--- a/frontend/packages/multicluster-sdk/src/api/search/types.ts
+++ b/frontend/packages/multicluster-sdk/src/api/search/types.ts
@@ -6,7 +6,9 @@ export type SearchResult<R extends K8sResourceCommon | K8sResourceCommon[]> = R 
   ? MulticlusterResource<T>[]
   : MulticlusterResource<R>
 
+export type AdvacedSearchFilter = { property: string; values: string[] }[]
+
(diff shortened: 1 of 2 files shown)


frontend/packages/multicluster-sdk/src/api/search/types.ts
```diff
@@ -8,5 +8,5 @@ export type SearchResult<R extends K8sResourceCommon | K8sResourceCommon[]> = R
 
 export type UseMulticlusterSearchWatch = <T extends K8sResourceCommon | K8sResourceCommon[]>(
   watchOptions: WatchK8sResource,
-  advancedSearch?: { [key: string]: string }
+  advancedSearch?: { key: string; value: string }[]
```
KevinFCormier, line 11: @Randy424 Maybe use `property` instead of `key` to better match search API terminology. And as discussed, allow `value` to optionally be an array.

upalatucci:
@KevinFCormier fixed thanks. I used the same filter type so we don't have to convert multiple times the param from array to string and then back to array

PR #4802, opened 2025-07-23
ACM-21336-Multicluster-SDK—observability hook

jeswanke:
1. refactored **useFleetPrometheusPoll** folder to align with other api hooks:
   index.ts --> useFleetPrometheusPoll.ts
   index.test.ts --> useFleetPrometheusPoll.test.ts
   added its constants to /internal/constants.ts
   added all other files to /internal

2. created a new **useFleetConfiguration.ts** hook that returns:
export type FleetConfiguration = {
  isGlobalHub: boolean
  localHubName: string
  isHubSelfManaged: boolean
  isObservabilityInstalled: boolean
}

now using the /hub path to determine if observability is installed and
1. rather then create a new hook for everything returned by hub I created one hook that returns them all
2. rather then get the configuration once, it polls every 15 seconds in case user changed hub name, etc

# 📝 Summary

**Ticket Summary (Title):**  
Multicluster SDK - implement hooks for observability

**Ticket Link:**  
https://issues.redhat.com/browse/ACM-21336

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [x] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---


package.json
```diff
@@ -15,6 +15,7 @@
     "start:frontend:hot": "cd frontend && npm run start:hot",
     "start:frontend:launch": "cd frontend && npm run launch",
     "plugins": "concurrently npm:start:backend npm:serve:plugins npm:ocp-console npm:watch:multicluster-sdk -n backend,frontend,console,watch -c green,blue,magenta,bgCyan",
+    "plugins-ocp": "concurrently npm:start:backend npm:serve:plugins npm:ocp-console -n backend,frontend,consoleh -c green,blue,magenta",
```
KevinFCormier, line 18: @jeswanke Was there a problem with running the watch?

jeswanke: @KevinFCormier -- watch works great--I tried doing alias too but it was no faster then watch w/o the advantage of updating the types--the reason for 'plugins-ocp' is if I was doing a lot of editing and saving it would trigger the build before the code was all set--I just like saying "all done, create a build"

frontend/packages/multicluster-sdk/src/api/useFleetConfiguration.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getBackendUrl } from './apiRequests'
+import { useEffect, useMemo, useState } from 'react'
+
+/**
+ * Hook that returns the fleet configuration. Checks periodically for changes in the fleet configuration.
+ *
+ * @returns {FleetConfiguration}
+ */
+
+export const useFleetConfiguration: UseFleetConfiguration = () => {
```
KevinFCormier, line 14: Hey @jeswanke. I think combining everything in 1 hook makes it more difficult to use. Because we have a nullable object returned as part of a tuple, I have to do something like this:
```
const [fleetConfiguration, configLoaded, configError] = useFleetConfiguration();
const { isObservabilityInstalled } = fleetConfiguration ?? {};
```

This API also overlaps with `useHubClusterName` now, so there are two slightly different ways to get that information.

Currently we don't have use cases for the other bits of information returned by the backend API, but if/when we do, we can always add new hooks. I don't think the polling is strictly necessary either, as changing the hub cluster name or installing observability are rare events.

PR #4856, opened 2025-08-06
ACM-20341:  make fieldselector work in the watch hook

upalatucci:
Something that was in TODO but forgot. 

We need to make sure that the fieldSelector and the labelSelector works in order to fetch filtered resources based on labels


frontend/packages/multicluster-sdk/src/internal/hooks/useDeepCompareMemoize.ts
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+
+export const useDeepCompareMemoize = <T = any>(value: T, strinfigy?: boolean): T | undefined => {
```
Randy424, line 4: Typo I think.
Should be "stringify"?

PR #4944, opened 2025-08-27
ACM-23618 Adds resource route extensions ACM first class resource pages

vishsanghishetty:
# 📝 Summary

**Ticket Summary (Title):**  

Summary
Register `acm.resource/route` extensions for ACM first-class resources to enable proper routing when using `@stolostron/multicluster-sdk` `FleetResourceLink` components.
Details

1. Add resource route extensions for all resources currently defined in searchDefinitions.tsx that link to ACM pages instead of generic search details
2. Use kubevirt-plugin implementation as reference pattern
3. Create separate extension for each resource kind with shared handler implementation
4. No changes to existing search implementation

**Ticket Link:**  

https://issues.redhat.com/browse/ACM-23618

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [X ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

### DIFF
--- a/frontend/plugins/acm/console-extensions.ts
+++ b/frontend/plugins/acm/console-extensions.ts
@@ -218,6 +218,62 @@ const identitiesRoute: EncodedExtension<RoutePage> = {
   },
 }
 
+// ACM resource route extensions - for fleetresourcelink integration
+
+// Managed cluster resource route
+const managedClusterResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { group: 'cluster.open-cluster-management.io', kind: 'ManagedCluster', version: 'v1' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
+
+// Cluster resource route
+const clusterResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { kind: 'Cluster' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
+
+// Application K8s resource route
+const applicationK8sResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { group: 'app.k8s.io', kind: 'Application', version: 'v1beta1' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
+
+// Application Argo resource route
+const applicationArgoResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { group: 'argoproj.io', kind: 'Application', version: 'v1alpha1' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
+
+// Policy resource route
+const policyResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { group: 'policy.open-cluster-management.io', kind: 'Policy', version: 'v1' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
+
+// Policy report resource route
+const policyReportResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { group: 'wgpolicyk8s.io', kind: 'PolicyReport', version: 'v1alpha2' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
+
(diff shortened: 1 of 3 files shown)


frontend/plugins/acm/console-extensions.ts
```diff
@@ -218,6 +218,62 @@ const identitiesRoute: EncodedExtension<RoutePage> = {
...
+}
+
+// Cluster resource route
+const clusterResourceRoute: EncodedExtension = {
+  type: 'acm.resource/route',
+  properties: {
+    model: { kind: 'Cluster' },
+    handler: { $codeRef: 'resource-routes.acmResourceRouteHandler' },
+  },
+}
```
KevinFCormier, line 239: Cluster is not a real resource - it's more like an alias of ManagedCluster that search provides. Please remove this case here and in the handler, and we'll need to handle it as a special case when we convert the search results code to use these extensions.

frontend/plugins/acm/resource-routes.ts
```diff
@@ -0,0 +1,63 @@
...
+ * ACM resource route handler for first-class resource pages.
+ * Routes resources to their dedicated ACM pages instead of generic search results.
+ */
+export const acmResourceRouteHandler: ResourceRouteHandler = ({ cluster, namespace, name, resource, model }) => {
+  const { kind, group } = model
+
+  switch (kind.toLowerCase()) {
+    case 'cluster':
+    case 'managedcluster':
+      return generatePath('/multicloud/infrastructure/clusters/details/:namespace/:name/overview', {
```
KevinFCormier, line 17: Let's move this file somewhere under frontend/src and use the NavigationPath constants, as the original code does.

frontend/plugins/acm/resource-routes.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * ACM resource route handler for first-class resource pages.
+ * Routes resources to their dedicated ACM pages instead of generic search results.
+ */
+export const acmResourceRouteHandler: ResourceRouteHandler = ({ cluster, namespace, name, resource, model }) => {
+  const { kind, group } = model
+
+  switch (kind.toLowerCase()) {
+    case 'cluster':
+    case 'managedcluster':
```
KevinFCormier, line 16: I think we should match on API group as well. You can see how Nitin did this here:
https://github.com/stolostron/console/blob/main/frontend/packages/multicluster-sdk/src/api/useFleetSearchPoll.ts#L196-L199

frontend/plugins/acm/resource-routes.ts
```diff
@@ -0,0 +1,63 @@
...
+      return generatePath('/multicloud/infrastructure/clusters/details/:namespace/:name/overview', {
+        namespace: name,
+        name,
+      })
+
+    case 'application': {
+      if (group === 'app.k8s.io' || group === 'argoproj.io') {
+        const params = queryString.stringify({
+          apiVersion: `${kind}.${group}`.toLowerCase(),
+          cluster: resource._hubClusterResource ? undefined : cluster,
```
KevinFCormier, line 26: _hubClusterResource is not an actual property that appears on k8s resources - it is something search adds. This TS error is being caught in VSCode and it should fail the builds, but I think we are not checking source files under frontend/plugin. So when you move it under frontend/src, this should show up as an error.

frontend/plugins/acm/resource-routes.ts
```diff
@@ -0,0 +1,63 @@
...
+          applicationset: resource.applicationSet ?? undefined,
```
KevinFCormier, line 27: Similarly, this is a property that search adds to its results. We'll have to handle this differently.

frontend/plugins/acm/resource-routes.ts
```diff
@@ -0,0 +1,63 @@
...
+      const path = generatePath('/multicloud/infrastructure/clusters/details/:namespace/:name/overview', {
+        namespace: namespace!,
+        name: namespace!,
+      })
+      return `${path}?${encodeURIComponent('showClusterIssues=true')}`
+    }
+  }
+
+  // fall back to null if no custom route found
+  return null as any
```
KevinFCormier, line 62: This cast to any struck me as odd. Type `ResourceRouteHandler` requires a return type of `string`. I think we should probably update the return type to be `string | undefined`.

KevinFCormier (changes requested):
Please see inline comments. Can you also remove the special handling for ManagedCluster in the FleetResourceLink component?

KevinFCormier (approved):

PR #4946, opened 2025-08-27
[ACM-23592] search details yaml page should watch resource for updates

zlayne:
# 📝 Summary

**Ticket Summary (Title):**  
Adds use of the `useFleetK8sWatchResource` function in the search details YAML tab.
This allows the page to watch the resource and recognize when there are updates to show that the YAML is now stale and needs to be refreshed.

**Ticket Link:**  
https://issues.redhat.com/browse/ACM-23592

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [x] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---


jlpadilla:
@zlayne @KevinFCormier 
- When `fine-grained-rbac-preview=false`
  - The console will need to have a ManagedServiceAccount and ClusterPermission. This is for the console workload, not tied to any individual user.
  - Console validates the individual user's RBAC (oc can-i create managedclusterview) and then uses the console token for cluster-proxy.
  - We used this logic in the earlier implementation of VM Actions.
- When `fine-grained-rbac-preview=true`
  - We just pass the user's token to the cluster-proxy.

zlayne:
@KevinFCormier, thanks for the review.

1. As is, the watch error (whether auth issues or otherwise) is essentially ignored. I don't display that an error was encountered, the effect hook that parses the response and updates the resource stale state will just not run, as an update event will never occur. In short, the page will still function, but the user will receive no indication that they are missing the watch functionality. If you think that should be changed, I can update accordingly.

2. I am working on adding the hook to the PluginContext. Is there currently a best practice function I should use to recognize console is in standalone mode? And subsequently "stub out" the hook?

KevinFCormier:
> @KevinFCormier, thanks for the review.
> 
> 1. As is, the watch error (whether auth issues or otherwise) is essentially ignored. I don't display that an error was encountered, the effect hook that parses the response and updates the resource stale state will just not run, as an update event will never occur. In short, the page will still function, but the user will receive no indication that they are missing the watch functionality. If you think that should be changed, I can update accordingly.
> 2. I am working on adding the hook to the PluginContext. Is there currently a best practice function I should use to recognize console is in standalone mode? And subsequently "stub out" the hook?

That's good that it will degrade gracefully. I don't think a notice is needed.

For (2), when you run the console in standalone mode (`npm start`) the code is not wrapped by a `PluginContextProvider`, so you will get the default value from the PluginContext (the stub). You can connect to the real hook in the context provider.

> @zlayne @KevinFCormier
> 
> * When `fine-grained-rbac-preview=false`
>   
>   * The console will need to have a ManagedServiceAccount and ClusterPermission. This is for the console workload, not tied to any individual user.
>   * Console validates the individual user's RBAC (oc can-i create managedclusterview) and then uses the console token for cluster-proxy.
>   * We used this logic in the earlier implementation of VM Actions.
> * When `fine-grained-rbac-preview=true`
>   
>   * We just pass the user's token to the cluster-proxy.

@jlpadilla I'm not keen on implementing that logic, if fine-grained RBAC is the strategic direction and will eventually become default. These service accounts would need permission to read everything. Currently the console has limited permissions for specific hub resources, and I don't want to expand that in scope and across the fleet. Also, we don't have an operator for console, and I'm not sure how achievable this would be with static configuration, as ClusterPermission is per-cluster, and even MulticlusterRoleAssignment might require you to list out individual clusters, as per current Slack discussions.

zlayne:
@KevinFCormier I have included the fleet watch function in the PluginContext [here](https://github.com/stolostron/console/pull/4946/files#diff-d4a1336c1b5e48af8949a3cbc7e553987f5f9a11666ddf0a0632e6f33d6ab2aaR21). If there is a better naming I could use, please let me know.

PR #5058, opened 2025-10-02
ACM-24895 source type detection

vishsanghishetty:
# 📝 Summary

**Ticket Summary (Title):**  

Discovered Argo CD Applications always show as Helm -> correct Argo CD source type detection logic

**Ticket Link:**  

https://issues.redhat.com/browse/ACM-24895

**Type of Change:**  

- [X] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

### DIFF
--- a/frontend/src/routes/Applications/Overview.getApplicationRepos.test.ts
+++ b/frontend/src/routes/Applications/Overview.getApplicationRepos.test.ts
@@ -1,0 +1,280 @@
+/* Copyright Contributors to the Open Cluster Management project */
+
+import {
+  ArgoApplication,
+  ArgoApplicationApiVersion,
+  ArgoApplicationKind,
+  ApplicationSet,
+  ApplicationSetApiVersion,
+  ApplicationSetKind,
+} from '../../resources'
+import { getApplicationRepos } from './Overview'
+
+describe('getApplicationRepos', () => {
+  const mockSubscriptions: any[] = []
+  const mockChannels: any[] = []
+
+  describe('Argo Applications', () => {
+    it('should return git type for application with path property', () => {
+      const gitApplication: ArgoApplication = {
+        apiVersion: ArgoApplicationApiVersion,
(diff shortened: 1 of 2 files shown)


KevinFCormier (commented):
Hey @vishsanghishetty,

I see now that the problem is that we do not have this data available from search. Your changes have simply switched the default from Helm to Git. The pairs in red boxes come from the same app and should have the same type, but you see here they are mixed now.

image

I checked this by searching for Argo applications and checking what we get back from search. See that the information is missing:
image


The problem seems to be that Argo allows either a single source or an array of sources, and search is only configured to look at the single source. See collector code here:
https://github.com/stolostron/search-collector/blob/main/pkg/transforms/argoapplication.go#L114-L118

image

I'm not sure if it is worth asking search to index either `source` or `source[0]`... unless it's common that the array format is used with a single source. Perhaps better to give no information than incomplete information.

For the details page we could certainly retrieve the full resource so we have all the source information.

@zlayne @fxiang1 @jeswanke any thoughts on this?

fxiang1:
I wonder if users would see this as a regression if we remove it? 

@ch-stark Do you have more insight if customers are actually using the `Repository` column? and if we can remove?

zlayne:
Search could add support for the sources array - likely it would store the entire array as a string that the UI would then have to parse. I'm not aware of how many sources we should expect in the array, but we may need to consider how much data this could require for search to store given an env with a large number of Apps. 

If it is not seen as a regression to remove the source column from the table. As Kevin mentioned, it would be trivial to display the full source info in the details page - as the page has the full resource YAML.

KevinFCormier:
Thanks Vishali!

PR #5110, opened 2025-10-21
MGMT-21727: change credentials type field

ElayAharoni:
https://issues.redhat.com/browse/MGMT-21727

🐞 Bug Fix  

changed the credentials type field in the console to not look like a disabled select component.

image


KevinFCormier:
@ElayAharoni can this be closed?

PR #5928, opened 2026-03-31
ACM-32046: Fix search prefill on policy details results tab

Randy424:
# 📝 Summary

**Ticket Summary (Title):**  
Policy details results tab search box may not prefilled from ?search= query parameter

**Ticket Link:**  
https://redhat.atlassian.net/browse/ACM-32046

**Type of Change:**  

- [x] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

## 🗒️ Notes for Reviewers

**Details:**
AcmTable has three layers of search state, resolved in priority order:

  props.search → storedSearch (from localStorage via context) → stateSearch (from useState)

  The old code used initialSearch, which only feeds into stateSearch (lowest priority). But AcmTableStateProvider always provides storedSearch (even
   as ''), and ?? treats '' as a valid value — so stateSearch was never reached.

  The fix:

  Instead of relying on initialSearch (uncontrolled, lowest priority), we pass search + setSearch as controlled props (highest priority). The search
   value is initialized from the URL param via useState, and setSearch lets the clear button and typing work normally.

  Also swapped window.location.search for React Router's useLocation().search so it reacts to client-side navigation.

## Summary by CodeRabbit

* **Refactor**
  * Policy details page now reads filter, search, and sort from the router URL and keeps the visible search field synced; table sort falls back to default only when no sort preference exists.
  * Table state now preserves an “unset” search (no forced empty string) until a value is provided or persisted, and initial search is stored when supplied.

* **Tests**
  * Added tests for search prefill from URL, and for various search-initialization and persistence behaviors.

### DIFF
--- a/frontend/src/routes/Governance/policies/policy-details/PolicyDetailsResults.test.tsx
+++ b/frontend/src/routes/Governance/policies/policy-details/PolicyDetailsResults.test.tsx
@@ -251,6 +251,7 @@ describe('Export from policy details results table', () => {
 
 describe('Search prefill from URL query string', () => {
   beforeEach(async () => {
+    localStorage.clear()
     nockIgnoreRBAC()
     nockIgnoreApiPaths()
   })
--- a/frontend/src/routes/Governance/policies/policy-details/PolicyDetailsResults.test.tsx
+++ b/frontend/src/routes/Governance/policies/policy-details/PolicyDetailsResults.test.tsx
@@ -248,3 +248,58 @@ describe('Export from policy details results table', () => {
     )
   })
 })
+
+describe('Search prefill from URL query string', () => {
+  beforeEach(async () => {
+    nockIgnoreRBAC()
+    nockIgnoreApiPaths()
+  })
+
+  test('Should prefill search bar when location.search contains a search param', async () => {
+    const context: PolicyDetailsContext = { policy: mockPolicy[0] }
+    render(
+      <RecoilRoot
+        initializeState={(snapshot) => {
+          snapshot.set(policiesState, mockPolicy)
+        }}
+      >
+        <MemoryRouter initialEntries={['?search=local-cluster']}>
+          <Routes>
+            <Route element={<Outlet context={context} />}>
+              <Route path="*" element={<PolicyDetailsResults />} />
+            </Route>
+          </Routes>
+        </MemoryRouter>
+      </RecoilRoot>
+    )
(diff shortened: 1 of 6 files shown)


KevinFCormier (changes requested):
Hey Randy,

I don't think this is the correct fix. When @jeswanke added caching of table state in local storage, it seems to have broken the `initialSearch` property. I think the `initialSearch` property should take precedence over any cached value in local storage. Currently it seems to be useless because of this effect that immediately overwrites the internal search value with the stored one, even if the stored search value is undefined:

https://github.com/stolostron/console/blob/206e88470044693ade8c3f12f0e049c0241b03ab/frontend/src/ui-components/AcmTable/AcmTable.tsx#L195-L199

There is also an effect in AcmTableToolbar with a debounce, and this is the one that is fixing your problem:
https://github.com/stolostron/console/blob/206e88470044693ade8c3f12f0e049c0241b03ab/frontend/src/ui-components/AcmTable/AcmTableToolbar.tsx#L362-L369

Because `search` is defined to be taken from the property, or the stored value, or the value stored in state in the toolbar (in that order of precedence), and now you are controlling the search property.
https://github.com/stolostron/console/blob/206e88470044693ade8c3f12f0e049c0241b03ab/frontend/src/ui-components/AcmTable/AcmTableToolbar.tsx#L339

But also, because you are now controlling the `search` property, if we have a cached search value, it does not get applied. 

I think this needs some more very careful thought about all of the interactions - which components are holding which pieces of state, and all of the state transitions.

KevinFCormier (commented):
Hey @Randy424. This works much better. But I still found the organization of everything a bit messy, with too much interaction between the AcmTableStateProvider, AcmTable, and AcmTableToolbar as well as effects to synchronize things. I also think when we follow a link with the search query param, that should not be stored in local storage, unless the user explicitly types it in.

So I've opened a new PR built on top of your changes:
https://github.com/stolostron/console/pull/5961

I think it's cleaner. It's not quite as reactive - we only read from local storage once upon loading the table, but I think that's a fair trade-off as we don't really have situations where we are changing the local storage key on the fly or navigating to a different search param from the same page.

PR #5957, opened 2026-04-08
ACM-30606 Wizard review collapse, link, highlight

jeswanke:
# 📝 Summary
Here is a description of the changes:

https://docs.google.com/document/d/1U2a_kY7IfbKwtBkybbJOp4qIO6bzzM0VZTOJzQY3dB8/edit?tab=t.0

Here is a video of the changes
https://redhat-internal.slack.com/archives/C02684L8JMV/p1775683819289319

**Ticket Summary (Title):**  
Make wizard review step expandable/support back link/support yaml highlight

**Ticket Link:**  
https://redhat.atlassian.net/browse/ACM-30806

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [x] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

## Summary by CodeRabbit

* **New Features**
  * Dedicated Review step: expandable sections with summary badges, expand/collapse-all toolbar, and edit controls that highlight or navigate to form fields
  * YAML editor highlighting tied to review navigation and persisted expansion state

* **UI/UX Improvements**
  * Inputs always show editable UI (removed read-only/details-only summaries)
  * Added stable wizard page IDs and refined review styling/layout; improved scroll-to-edit highlighting behavior

* **Tests**
  * Cypress tests reformatted; removed obsolete review test case


frontend/packages/react-form-wizard/src/inputs/WizArrayInput.tsx
```diff
@@ -441,3 +455,60 @@ export function ArrayInputItem(props: {
...
+      value,
+      label: label ?? '',
+      type: InputReviewMeta.ARRAY_INSTANCE,
+    })
+    bumpReviewDomTree?.()
+    return () => {
+      stepInputsRegistry.unregister(id)
+      bumpReviewDomTree?.()
+    }
+  }, [stepInputsRegistry, index, collapsedContent, measureRef, value, id, item, bumpReviewDomTree])
```
fxiang1: I think `index` can be removed from here since it's not used in the `useLayoutEffect`.

fxiang1 (changes requested):

fxiang1:
@jeswanke Overall I think this PR looks great! I just commented on a minor issue.

One question, I noticed you added this feature to the cluster import wizard but not the cluster create wizard, will you create another work item to add this to other wizards?

fxiang1:
@jeswanke I just realized on the Credential wizard we already have a design to not hide the action button on the review step. Maybe we can change the back link and highlight button to display like this? ie. not require the user to hover over to show the buttons and align to the right.

image

fxiang1:
The latest changes look great John! I'll put this on hold in case others have anything.

/lgtm
/hold

KevinFCormier (commented):
Previously, we could have custom components and they could decide how to provide content for the review step. For example in the import cluster wizard:

Image

Are we able to handle that with this approach? I'm a bit concerned that everything is centralized and that there is a large amount of custom CSS. So when you add a new component, you have to figure out how to add it in the Review step as well. I'm wondering if it would be feasible to still have the inputs provide their own content for the Review step, but give them a set of component building blocks to use. Like an expandable, a row of data, the edit button. But I recognize that that could be difficult and has the potential to lead to inconsistencies. We need to solve this issue with the cluster import wizard though.

And do we have a way to omit something from the Review step if we think it's not important? I'm not sure if Review should be comprehensive or not. But I do I think it should be easy to read through quickly as a final confirmation before you submit. So one thing I preferred before the changes is for the Sync policy, only checked boxes would show up. Here we have to scan through a column of true/false and <not set> to see what was selected. In general, are we omitting things that are not set from the Review step? That's a good way to keep it brief, I think.

frontend/packages/react-form-wizard/src/review/ReviewStep.tsx
```diff
@@ -0,0 +1,1004 @@
...
+}
+
+function renderReviewInputDescriptionContent(node: WizardInputDomNode): ReactNode {
+  if (node.error) {
+    return <span style={{ color: REVIEW_ERROR_TEXT_COLOR, fontStyle: 'italic' }}>{node.error}</span>
+  }
+  if (!isReviewValueUnset(node.value)) {
+    return formatReviewValue(node.value)
+  }
+  return <span style={{ fontStyle: 'italic' }}>{'<not set>'}</span>
```
KevinFCormier: Needs translation

KevinFCormier: We might want to translate boolean true/false as well... I don't know. Booleans don't appear in the form, only the YAML. I see the Review step as part of the "form world"

Randy424:
Hi @jeswanke,
There's some really slick stuff going on in this PR. Great job.

I think I'm stuck with slim pickings at this point but I did notice a difference in how the summary text is responding to the yaml editor as the view of the page shrinks and the editor expands. It looks like we are now using flex wrap behavior, when previously the section titles were static and set behind the editor when it expanded. I'm partial to a static behavior for the section titles in this case but also I'm not certain that's the best solution either. Maybe treat this comment as an FYI :) See the recordings below.

--also expanding the editor can cover some of the edit buttons. From a UX pov, I guess that's a no-no? Would appreciate a second opinion on that though.

Before

https://github.com/user-attachments/assets/095418d7-7686-4cfa-920b-fa607346e2e9

After

https://github.com/user-attachments/assets/38f9831c-a4b5-493b-8f29-e7bb7c0e1e5a

jeswanke:
@KevinFCormier good points:
1. the latest change omits values not set from the Review step (and therefore remove this line and its translation:{'<not set>'}
2. booleans are now displayed as checkmarks:
CleanShot 2026-04-14 at 11 00 19@2x

jeswanke:
@KevinFCormier we also have collapse to hide sections of details
I have also added a new property to Wiz components tio omit it from review. step:
CleanShot 2026-04-14 at 11 19 12@2x

KevinFCormier (changes requested):
@jeswanke Thanks for the updates. I'm happy with all the behaviour now.

Just wondering if we can fully remove DisplayMode.Details from the codebase now?

PR #6294, opened 2026-06-03
ACM-34489: Fix cluster links and empty-state copy for namespace-scoped users

Randy424:
## Summary

Fixes two UX issues on the GRC policy **Results tab** for namespace-scoped users who lack cluster-level RBAC:

1. **Cluster names rendered as clickable links** that navigated to empty/unhelpful cluster overview pages
2. **Misleading empty-state copy** ("No clusters are reporting status for this policy.") when clusters *are* reporting — the user simply lacks permission to view the propagated policy resources

### Context

During triage of [ACM-32999](https://issues.redhat.com/browse/ACM-32999), Derek Ho [reported](https://issues.redhat.com/browse/ACM-32999?focusedId=27036795&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27036795) that when navigating the console as a namespace-scoped user (`e2e-admin-ns`):

> *"for the inaccessible clusters, all their names on GRC pages are clickable. This is also misleading to the user who may think more information is there. But they should not be clickable if the user has no permission"*

He also noted the empty-state message was inaccurate:

> *"a message like 'No result is available' is better than 'No cluster results. No clusters are reporting status for this policy.', in this use case."*

This defect was distilled from ACM-32999 as the unambiguous z-stream fix for ACM 2.17.1. The broader UX question (whether to prevent namespace-scoped users from reaching the Results tab entirely vs. showing an informational empty state) is tracked separately as an RFE.

### 1. Conditional cluster link rendering (`PolicyDetailsResults.tsx`)

**What:** Cluster names now render as plain text when the cluster is not present in the `managedClustersState` Recoil atom. When the cluster IS present, the existing `<Link>` behavior is preserved.

**Why:** `ManagedCluster` is a cluster-scoped resource. The backend SSE event filter already gates `ManagedCluster` events by RBAC — if a namespace-scoped user can't `list` ManagedClusters, the atom will be empty. This is a **data-driven RBAC gate**: rather than making an additional `SelfSubjectAccessReview` API call, we check data the backend has already filtered.

**Tradeoff:** A direct RBAC check (`canUser('get', ManagedClusterDefinition)`) would be more explicit, but adds a network round-trip for a check the SSE data already encodes. The data-driven approach is simpler, has per-cluster granularity, and aligns with how the backend works.

### 2. Updated empty-state copy (`PolicyDetailsResults.tsx`)

**What:** Changed from `"No cluster results"` / `"No clusters are reporting status for this policy."` to `"No results found"` / `"No results available."`.

**Why:** The old message implied clusters weren't reporting status, which is inaccurate — clusters ARE reporting, the user just lacks permission to view the propagated policy resources. The new copy is neutral and accurate for all scenarios, directly addressing Derek's feedback.

### 3. Updated i18n keys (`translation.json`)

**What:** Added `"No results available."` key. Removed now-unused `"No cluster results"` and `"No clusters are reporting status for this policy."` keys (auto-pruned by the i18n parser).

### 4. Tests (`PolicyDetailsResults.test.tsx`)

**What:** Updated all existing tests to set `managedClustersState` (so cluster links still render for admin users). Added two new tests:
- Verifies cluster names render as **plain text** (no `<a>` tag in the Cluster column) when `managedClustersState` is empty
- Verifies the **updated empty-state message** when no results are available

## Test plan

- [x] `npx jest PolicyDetailsResults.test --no-coverage` — all 10 tests pass (8 existing + 2 new)
- [x] `npx tsc --noEmit` — no type errors
- [x] `npx eslint` on changed files — clean
- [x] `npx prettier --write` on changed files — formatted
- [x] `npm run i18n` — no warnings, no missing keys
- [x] E2E verified on weekly cluster (`cs-aws-422-qcfg4`) with both admin and namespace-scoped (`e2e-admin-ns`) users via Playwright

[ACM-32999]: https://redhat.atlassian.net/browse/ACM-32999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cluster links appearing for clusters users don't have access to; they now display as plain text instead.
  * Updated empty-state messaging to provide clearer feedback when no results are available.


fxiang1:
Thanks for fixing this Randy! Looks good to me. Will hold in case Derek has anything.

/lgtm
/hold

drkho:
LGTM

Thanks for making the changes.

PR #6303, opened 2026-06-05
ACM-30906: Fix AAP install script for OCP 4.22 and AAP 2.7

Randy424:
## Summary

- **Auto-detect operator channel** from OperatorHub catalog instead of hardcoding `stable-2.5`, which is unavailable on OCP 4.22 (only `stable-2.7` exists there)
- **Fix manifest upload endpoint** — use `/config/` with `eula_accepted` instead of `/config/subscriptions/` with `eula` (the subscriptions endpoint is for service account credentials, not manifests)
- **Add allocation reuse fallback** — reuse any existing RHSM allocation with entitlements when the date-named allocation isn't found
- **Fix RHSM export multi-hop chain** — follow the full async export pipeline (`/export` → `/exportJob/{id}` → `/export/{id}`) with ZIP verification
- **Support AAP 2.7 gateway token endpoint** — try `/api/gateway/v1/tokens/` first (AAP 2.7), fall back to `/api/controller/v2/tokens/` (AAP 2.5)
- **Fix allocation creation type** — use `SAM 1.4` instead of `Satellite 6.11` (avoids `sat-` prefix requirement)

## Test plan

- [x] Verified on OCP 4.22 / AAP 2.7 (controller v4.8.0) — full end-to-end pass
- [x] Channel auto-detected as `stable-2.7`
- [x] Allocation `AAP-Weekly-Cluster` reused with 1 AAP entitlement
- [x] Manifest exported via multi-hop (95KB ZIP with valid AAP subscription)
- [x] Manifest uploaded and subscription activated (enterprise, non-trial, expires Jan 2027)
- [x] OAuth token generated via gateway endpoint
- [x] ACM credential `aap-shared-credential` created
- [ ] Verify on OCP 4.20 / AAP 2.5 (backward compatibility)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Enhanced installation automation with automatic operator channel detection
  * Improved handling of offline subscriptions and manifests
  * Refined authentication token generation with improved reliability

PR #6307, opened 2026-06-07
ACM-30906: Auto-detect AAP operator channel when unavailable

Randy424:
## Summary

- Validate the requested operator channel against the OperatorHub catalog before installing
- If the channel doesn't exist (e.g. `stable-2.5-cluster-scoped` on OCP 4.22), fall back to auto-detecting the latest available `stable-*` channel
- This allows callers like rhacmstackem to keep their existing defaults without needing updates when OCP drops older AAP channels

## Test plan

- [x] Tested with `OPERATOR_CHANNEL=stable-2.5-cluster-scoped` on OCP 4.22 — detected unavailable, auto-fell back to `stable-2.7`
- [x] Full end-to-end pass: operator install, platform deploy, subscription activation, OAuth token, ACM credential
- [x] Cluster: `cs-aws-422-m4jn6` (OCP 4.22, AAP 2.7 / controller v4.8.0)

## Summary by CodeRabbit

* **Bug Fixes**
  * Always verify operator channels against OperatorHub before installation.
  * If no channels are returned, installation now fails fast with an error.
  * If a configured channel is invalid, a warning is shown and the installer auto-detects a channel.
  * Auto-detection prefers stable-* channels (excluding cluster-scoped) or selects the highest available version.

### DIFF
--- a/scripts/aap-automation/install-aap.sh
+++ b/scripts/aap-automation/install-aap.sh
@@ -133,7 +133,7 @@ fi
 rm -f "$OC_PKG_STDERR"
 
 if [ -n "$OPERATOR_CHANNEL" ] && ! echo "$AVAILABLE_CHANNELS" | grep -qx "$OPERATOR_CHANNEL"; then
-    log_warn "Requested channel '$OPERATOR_CHANNEL' not available — will auto-select the latest stable-* channel"
+    log_warn "Requested channel '$OPERATOR_CHANNEL' not available — will auto-select the latest stable-* channel, or the latest available channel if none match"
     OPERATOR_CHANNEL=""
 fi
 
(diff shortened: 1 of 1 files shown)


Randy424:
Hi @zlayne @KevinFCormier 
In my [last update](https://github.com/stolostron/console/pull/6303) to this script I added logic to detect a viable operator channel from OperatorHub when no version is explicitly passed to the script, but I should have also included conditions to make this happen when the version that is passed in is not available on the hub. This PR includes that feature. I think we will be ready to go following this change. The offline token in our collective cluster has already been updated and is working. I've tested the script locally myself, and will test using the weekly cron job once more but after this is merged. Thanks for bearing with me here!

KevinFCormier (approved):

PR #6339, opened 2026-06-15
ACM-35158 avoid blocking incoming requests [release-2.16]

KevinFCormier:
# 📝 Summary

**Ticket Summary (Title):**  
console-mce pod CrashLoopBackOff due to probe failure with large resource sets

**Ticket Link:**  
https://redhat.atlassian.net/browse/ACM-35158

**Type of Change:**  

- [x] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

- [x] *(Nice to have)* JSDoc comments added for new functions and interfaces

#### If Bugfix

- [x] Test(s) added to prevent regression

---

### 🗒️ Notes for Reviewers
Customer was observing CrashLoopBackoff on console-mce pods and had over 20,000 Group resources present. Initially processing these resources may have blocked us from responding to the readiness and liveness probes on time, causing Kubernetes to restart the pod.

This PR batches Promise.all calls so that we call setImmediate after each batch of 100 to yield the event queue so that new requests can be handled. This seems to smooth out memory usage of the pods. Before this change, I observed a spike at startup, then a retreat to stable size.

### Memory Usage

In the following videos, I deleted pods, then observed the memory usage of the replacements. You can see that before the patch, the pods sometimes have a memory spike during startup, and in this case we see one of the pods spike to almost 12 GB, before falling back to a more typical value. After the patch, there is no initial spike.

#### Before
https://github.com/user-attachments/assets/3501ee1e-34a5-49d2-9069-e514ac21c2a7

#### After
https://github.com/user-attachments/assets/2c64390d-1919-4158-9301-04d29d5dc72b

### Liveness Probe Response Time

I also checked the response time, turning on garbage collection tracing and using a simple script to repeatedly check the liveness probe endpoint. I tested against cluster `kevin-probe-test` which has 50,000 Group resources on it.

Here is the process I followed:
1. Run `./check-response-time.sh` or `./check-response-time.sh -t 1.0` in one terminal. (The latter command shows only when response time is greater than 1.0 seconds.)
2. In a second terminal, run `npm run plugins`.
3. Drive load to the backend by loading `https://localhost:9000/multicloud/credentials` in several tabs. Repeatedly refresh the tabs to drive new full loads of the SSE stream.

Using a [baseline test branch](https://github.com/KevinFCormier/console/tree/release-2.16-baseline) vs. a [patched test branch](https://github.com/KevinFCormier/console/tree/ACM-35158-avoid-blocking-incoming-requests-debug), I found:
- Without the patch, I can easily provoke response times of 1-2 seconds. I saw a particularly long pause of 53 s on one occasion.
- With the patch, I used `./check-response-time.sh -t 0.1` and only saw response times up to just under 0.5 s.

### DIFF
--- a/backend/src/lib/batch-promise-all.ts
+++ b/backend/src/lib/batch-promise-all.ts
@@ -1,0 +1,20 @@
+/* Copyright Contributors to the Open Cluster Management project */
+
+export const BATCH_SIZE = 100
+
+/**
+ * Process an array of items through an async mapper in batches, yielding the
+ * event loop between batches to allow HTTP requests (e.g. liveness probes) to
+ * be served.
+ */
+export async function batchPromiseAll<T, R>(items: T[], mapper: (item: T) => Promise<R>): Promise<R[]> {
+  const results: R[] = []
+  for (let i = 0; i < items.length; i += BATCH_SIZE) {
+    const batch = await Promise.all(items.slice(i, i + BATCH_SIZE).map(mapper))
+    results.push(...batch)
+    if (i + BATCH_SIZE < items.length) {
+      await new Promise((resolve) => setImmediate(resolve))
+    }
+  }
+  return results
+}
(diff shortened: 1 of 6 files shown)


fxiang1:
/lgtm

I put this on hold, not sure if the 2.16 stream is open.

PR #6490, opened 2026-07-16
FCN-247: regions api call

davidaznaur:
# 📝 Summary

**Ticket Summary (Title):**  
Regions api call for the ROSA HCP wizard

**Ticket Link:**  
[](https://redhat.atlassian.net/browse/FCN-247)

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [X] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

- [X] All acceptance criteria met
- [X] Relevant documentation or comments included

#### If Bugfix

---

## Summary by CodeRabbit

- **New Features**
  - Added a ROSA wizard API endpoint (`POST /regions`) and a matching client helper to fetch cloud providers and available regions.
  - Introduced a Regions dropdown powered by a new regions hook and query keys, showing only AWS regions that support hypershift/hosted control planes.

- **Bug Fixes**
  - Improved backend error trapping for wizard requests and standardized internal error responses.
  - Updated frontend hook error handling to surface clearer messages when requests fail.

- **Tests**
  - Added backend coverage for `POST /regions`, plus frontend tests for query keys, API wiring, region filtering/formatting, and hook behavior.


backend/src/routes/rosaWizardApi.ts
```diff
@@ -73,22 +78,62 @@ export async function getAwsBillingAccountIds(req: Http2ServerRequest, res: Http
...
+      let data: string = undefined
+      const chucks: string[] = []
+      req.on('data', (chuck: string) => {
+        chucks.push(chuck)
+      })
+
+      req.on('end', async () => {
+        try {
+          data = chucks.join()
+          const body = JSON.parse(data) as Payload
```
vishsanghishetty, line 164: Can you switch chucks.join() to chucks.join("") here? If the body comes in multiple chunks, the default join adds commas between chunks and can break JSON parsing.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
```diff
@@ -0,0 +1,41 @@
...
+import { CloudProviderResponse } from '~/resources'
+
+const hcpCloudProvidersAndRegions = (cloudProvidersResponse: CloudProviderResponse): DropdownType[] => {
+  const awsProvider = cloudProvidersResponse.items?.find((provider) => provider.id === 'aws')
+  if (!awsProvider?.regions) return []
+  return awsProvider.regions
+    .filter((region) => region.supports_hypershift === true)
+    .map((region) => ({
+      value: region.id ?? '',
+      label: `${region.id}, ${region.display_name}`,
```
vishsanghishetty, line 16: Can we handle regions without an id before mapping? Right now label uses `${region.id}` so we can end up with `undefined, ...` in the dropdown. Filtering out id-less regions here would avoid that user-facing issue.

backend/src/routes/rosaWizardApi.ts
```diff
@@ -73,22 +78,62 @@ export async function getAwsBillingAccountIds(req: Http2ServerRequest, res: Http
...
+        }
+      })
+    } catch (err) {
+      logger.error(err)
+      respondInternalServerError(req, res)
+    }
+  }
+}
+
+export async function getWizardCloudProviders(req: Http2ServerRequest, res: Http2ServerResponse) {
```
vishsanghishetty: Can you add the explicit `: Promise<void>` return type on this route handler? The other handlers in this file already use it, and it keeps the route signature consistent.

backend/src/routes/rosaWizardApi.ts
davidaznaur: @vishsanghishetty so, I know it is an llm suggested change, I got the same from coderabbit. The problem is that no other API call within ACM that uses this chucks.join implements it with chucks.join("") @KevinFCormier can you please advice on this one?

vishsanghishetty: @davidaznaur 
I checked the repo and we actually have both forms today, including chucks.join('') in virtualMachineProxy, so this wouldn’t be a new convention.

the reason I called it out is behavior, not style: default join() inserts commas, so if a request body arrives in multiple chunks the JSON can be corrupted before parse. if this endpoint can ever receive multi-chunk payloads, join('') is the safer option. happy to treat it as a follow-up cleanup across similar handlers if we don’t want to scope-creep this PR.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
davidaznaur: I don't think ID can be undefined. These are aws cloud provider regions, and I checked in AWS console, they have to have ID defined. It is OCM type schema that defines ALL the fields as optionals for some reason

backend/src/routes/rosaWizardApi.ts
davidaznaur: Yep, will add it

backend/src/routes/rosaWizardApi.ts
vishsanghishetty: chucks.join() / chunks.join() (no separator):

[backend/src/routes/rosaWizardApi.ts#L41](https://github.com/stolostron/console/blob/main/backend/src/routes/rosaWizardApi.ts#L41)
[backend/src/routes/rosaWizardApi.ts#L82](https://github.com/stolostron/console/blob/main/backend/src/routes/rosaWizardApi.ts#L82)
[backend/src/routes/rosaWizardApi.ts#L122](https://github.com/stolostron/console/blob/main/backend/src/routes/rosaWizardApi.ts#L122)
[backend/src/routes/userpreference.ts#L83](https://github.com/stolostron/console/blob/main/backend/src/routes/userpreference.ts#L83)
[backend/src/routes/operatorCheck.ts#L44](https://github.com/stolostron/console/blob/main/backend/src/routes/operatorCheck.ts#L44)

chucks.join('') (explicit empty separator):

[backend/src/routes/virtualMachineProxy.ts#L132](https://github.com/stolostron/console/blob/main/backend/src/routes/virtualMachineProxy.ts#L132)

KevinFCormier: @davidaznaur I agree we need the empty string separator. I will open a separate bug and address the pre-existing instances of this. Please use the empty string as separator for the new cases you are introducing.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const useFetchRegions = (selectedSecret: SelectedSecret) => {
+  const { useQuery } = useSharedReactQuery()
+  const { data, isLoading, isError, error, refetch } = useQuery({
+    queryKey: rosaWizardKeys.regions(selectedSecret?.client_id),
+    queryFn: async ({ signal }) => {
+      const response = await getWizardRegions(selectedSecret.client_id, selectedSecret.client_secret, signal)
+
+      return response ?? []
+    },
```
KevinFCormier, line 28: 4. frontend/src/routes/.../useFetchRegions.ts:26-28 queryFn returns response ?? [] — type mismatch with select — The queryFn returns response ?? [], where [] is never[] (not CloudProviderResponse). But the select function (hcpCloudProvidersAndRegions) expects CloudProviderResponse. If response is nullish (which shouldn't happen since getWizardData throws on errors), passing [] would cause select to call .items?.find() on an array, which has no items property, returning undefined → empty regions. The fallback masks errors rather than surfacing them.
Suggestion: Remove the ?? [] fallback — if getWizardData returns nullish, it indicates a logic error that should be surfaced:
queryFn: async ({ signal }) => {
  return getWizardRegions(selectedSecret.client_id, selectedSecret.client_secret, signal)
},

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
```diff
@@ -0,0 +1,41 @@
...
+    enabled: !!selectedSecret,
+    retry: false,
+    select: hcpCloudProvidersAndRegions,
+  })
+
+  return {
+    data,
+    isLoading,
+    isError,
+    error: isError ? JSON.stringify(error) : null,
```
KevinFCormier: 5. frontend/src/routes/.../useFetchRegions.ts:37 error: isError ? JSON.stringify(error) : null — JSON.stringify on an Error object produces {} (Error properties aren't enumerable). For a ResourceError, it might work because custom properties are set, but standard Error objects will produce "{}". Other hooks in this series use error instanceof Error ? error.message : 'Unknown error' which is more reliable.
Suggestion: Use the pattern from sibling hooks:
error: isError ? (error instanceof Error ? error.message : 'Unknown error') : null,

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
```diff
@@ -0,0 +1,41 @@
...
+      label: `${region.id}, ${region.display_name}`,
+    }))
+}
+
+export const useFetchRegions = (selectedSecret: SelectedSecret) => {
+  const { useQuery } = useSharedReactQuery()
+  const { data, isLoading, isError, error, refetch } = useQuery({
+    queryKey: rosaWizardKeys.regions(selectedSecret?.client_id),
+    queryFn: async ({ signal }) => {
+      const response = await getWizardRegions(selectedSecret.client_id, selectedSecret.client_secret, signal)
```
KevinFCormier, line 25: 6. frontend/src/routes/.../useFetchRegions.ts:22 Potential crash when selectedSecret is falsy — The hook accesses selectedSecret?.client_id for the query key (safe — optional chaining), but then selectedSecret.client_id and selectedSecret.client_secret in queryFn (unsafe — no optional chaining). If selectedSecret is falsy, enabled is false, but React Query may still construct the query object and evaluate the queryKey before checking enabled. The queryFn won't run, but the queryKey evaluates to [..., undefined, 'regions'].
Suggestion: The current code works because queryFn is guarded by enabled, but for robustness, consider adding a type guard or defaulting:
queryKey: rosaWizardKeys.regions(selectedSecret?.client_id ?? ''),

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
```diff
@@ -0,0 +1,41 @@
...
+import { getWizardRegions } from '~/lib/rosa-hcp-api'
+import { DropdownType, SelectedSecret } from '../constants/types'
+import { rosaWizardKeys } from './queryKeyFactory'
+import { CloudProviderResponse } from '~/resources'
+
+const hcpCloudProvidersAndRegions = (cloudProvidersResponse: CloudProviderResponse): DropdownType[] => {
+  const awsProvider = cloudProvidersResponse.items?.find((provider) => provider.id === 'aws')
+  if (!awsProvider?.regions) return []
+  return awsProvider.regions
+    .filter((region) => region.supports_hypershift === true)
```
KevinFCormier:
```suggestion
    .filter((region) => region.supports_hypershift)
```

KevinFCormier (changes requested):

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
davidaznaur: I guess I took hook from OCM without changing it. I dont think we will need anything other than what RQ provides as an error

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRegions.ts
davidaznaur: The mistake is not in getWizardRegions but in rosaWizardKeys.regions(selectedSecret?.client_id), there is no need for optional chaining. client_id will always be selected, otherwise user wont be able to get into the wizard

KevinFCormier (dismissed):

KevinFCormier (approved):

davidaznaur:
@KevinFCormier I had to add another commit here after rebase and merge conflict fixes, its just prettier in backend and frontend folders

KevinFCormier (approved):

PR #6498, opened 2026-07-16
FCN-254: role arns api call

davidaznaur:
# 📝 Summary

**Ticket Summary (Title):**  
Account, OCM and User role ARNs api calls

**Ticket Link:**  
[](https://redhat.atlassian.net/browse/FCN-254)

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [X] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

- [] All new display strings are externalized for localization (English only)

#### If Feature

- [X] All acceptance criteria met

#### If Bugfix

---

## Summary by CodeRabbit

* **New Features**
  * Added ROSA/HCP wizard support to fetch AWS role ARNs, OCM role details, and the STS user role.
  * Introduced new backend POST endpoints and matching frontend proxy routes for `/sts-role-arns`, `/sts-ocm-role`, and `/sts-user-role`.
* **Bug Fixes**
  * Improved request JSON reconstruction for AWS role lookups.
  * Enhanced role fetching error handling, including friendly “user role not found” messaging.
* **Tests**
  * Expanded coverage for the new routes and the roles-fetching hook, including auth failures.
* **Refactor**
  * Updated frontend wizard API helpers and role normalization/types to support the new data flow.

### DIFF
--- a/backend/src/routes/rosaWizardApi.ts
+++ b/backend/src/routes/rosaWizardApi.ts
@@ -19,6 +19,7 @@ type OrgType = {
   }
   service_account: boolean
   username: string
+  id: string
 }
 
 type Payload = {
@@ -221,6 +222,131 @@ export async function getClusterNameCheck(req: Http2ServerRequest, res: Http2Ser
             accessTokenSSO
           ).catch((err: Error) => {
             logger.error({ msg: 'Error getting account info', error: err.message })
+          })
+
+          res.setHeader('Content-Type', 'application/json')
+          res.end(JSON.stringify(accReq))
+        } catch (err) {
+          logger.error(err)
+          respondInternalServerError(req, res)
+        }
+      })
+    } catch (err) {
+      logger.error(err)
+      respondInternalServerError(req, res)
+    }
+  }
+}
+
+export async function getRoleARNs(req: Http2ServerRequest, res: Http2ServerResponse): Promise<void> {
+  const token = await getAuthenticatedToken(req, res)
+  if (token) {
+    try {
+      let data: string = undefined
+      const chucks: string[] = []
+      req.on('data', (chuck: string) => {
+        chucks.push(chuck)
+      })
+
+      req.on('end', async () => {
+        try {
+          data = chucks.join()
+          const body = JSON.parse(data) as WithAwsAccount
+
+          const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
+
+          const accountPath = `${API_URL}/api/clusters_mgmt/v1/aws_inquiries/sts_account_roles`
+
+          const requestBody = {
+            account_id: body.aws_account_id,
+          }
+          const accReq = await jsonPost(accountPath, requestBody, accessTokenSSO).catch((err: Error) => {
+            logger.error({ msg: 'Error gettting account info', error: err.message })
+          })
+
+          res.setHeader('Content-Type', 'application/json')
+          res.end(JSON.stringify(accReq))
+        } catch (err) {
+          logger.error(err)
+          respondInternalServerError(req, res)
+        }
+      })
+    } catch (err) {
+      logger.error(err)
+      respondInternalServerError(req, res)
+    }
+  }
+}
+
+export async function getOCMRoleARN(req: Http2ServerRequest, res: Http2ServerResponse): Promise<void> {
+  const token = await getAuthenticatedToken(req, res)
+  if (token) {
+    try {
+      let data: string = undefined
+      const chucks: string[] = []
+      req.on('data', (chuck: string) => {
+        chucks.push(chuck)
+      })
+
+      req.on('end', async () => {
+        try {
+          data = chucks.join()
+          const body = JSON.parse(data) as WithAwsAccount
+
+          const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
+
+          const accountPath = `${API_URL}/api/clusters_mgmt/v1/aws_inquiries/sts_ocm_role`
+
+          const requestBody = {
+            account_id: body.aws_account_id,
+          }
+          const accReq = await jsonPost(accountPath, requestBody, accessTokenSSO).catch((err: Error) => {
+            logger.error({ msg: 'Error gettting account info', error: err.message })
+          })
+
+          res.setHeader('Content-Type', 'application/json')
+          res.end(JSON.stringify(accReq))
+        } catch (err) {
+          logger.error(err)
+          respondInternalServerError(req, res)
+        }
+      })
+    } catch (err) {
+      logger.error(err)
+      respondInternalServerError(req, res)
+    }
+  }
+}
+
+export async function getUserRole(req: Http2ServerRequest, res: Http2ServerResponse): Promise<void> {
+  const token = await getAuthenticatedToken(req, res)
+  if (token) {
+    try {
+      let data: string = undefined
+      const chucks: string[] = []
+      req.on('data', (chuck: string) => {
+        chucks.push(chuck)
+      })
+
+      req.on('end', async () => {
+        try {
+          data = chucks.join()
+          const body = JSON.parse(data) as Payload
+
+          const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
+
+          // get current account and ID
+          const accountPath = `${API_URL}/api/accounts_mgmt/v1/current_account`
+          const request = (await jsonRequest(accountPath, accessTokenSSO).catch((err: Error) => {
+            logger.error({ msg: 'Failed to fetch account', error: err.message })
+            return { error: err.message }
+          })) as OrgType
+
+          const userRolesPath = `${API_URL}/api/accounts_mgmt/v1/accounts/${request.id}/labels/sts_user_role
+`
+
+          const accReq = await jsonRequest(userRolesPath, accessTokenSSO).catch((err: Error) => {
+            logger.error({ msg: 'Error gettting account info', error: err.message })
           })
 
           res.setHeader('Content-Type', 'application/json')
--- a/backend/src/routes/rosaWizardApi.ts
+++ b/backend/src/routes/rosaWizardApi.ts
@@ -250,7 +250,7 @@ export async function getRoleARNs(req: Http2ServerRequest, res: Http2ServerRespo
 
       req.on('end', async () => {
         try {
-          data = chucks.join()
+          data = chucks.join('')
           const body = JSON.parse(data) as WithAwsAccount
 
           const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
@@ -290,7 +290,7 @@ export async function getOCMRoleARN(req: Http2ServerRequest, res: Http2ServerRes
 
       req.on('end', async () => {
         try {
-          data = chucks.join()
+          data = chucks.join('')
           const body = JSON.parse(data) as WithAwsAccount
 
           const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
@@ -330,20 +330,23 @@ export async function getUserRole(req: Http2ServerRequest, res: Http2ServerRespo
 
       req.on('end', async () => {
         try {
-          data = chucks.join()
+          data = chucks.join('')
           const body = JSON.parse(data) as Payload
 
           const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
 
           // get current account and ID
           const accountPath = `${API_URL}/api/accounts_mgmt/v1/current_account`
-          const request = (await jsonRequest(accountPath, accessTokenSSO).catch((err: Error) => {
-            logger.error({ msg: 'Failed to fetch account', error: err.message })
-            return { error: err.message }
-          })) as OrgType
+          let request: OrgType
+          try {
+            request = await jsonRequest(accountPath, accessTokenSSO)
+          } catch (err) {
+            logger.error({ msg: 'Failed to fetch account', error: (err as Error).message })
+            respondInternalServerError(req, res)
+            return
+          }
 
-          const userRolesPath = `${API_URL}/api/accounts_mgmt/v1/accounts/${request.id}/labels/sts_user_role
-`
+          const userRolesPath = `${API_URL}/api/accounts_mgmt/v1/accounts/${request.id}/labels/sts_user_role`
 
           const accReq = await jsonRequest(userRolesPath, accessTokenSSO).catch((err: Error) => {
             logger.error({ msg: 'Error gettting account info', error: err.message })
--- a/frontend/src/lib/rosa-hcp-api.ts
+++ b/frontend/src/lib/rosa-hcp-api.ts
@@ -7,8 +7,8 @@ import {
   CloudProviderResponse,
   ClusterNameCheckPayload,
   ClusterNameUniquenessResponse,
-  OrganizationQuotaResponse,
   OCMRoleResponse,
+  OrganizationQuotaResponse,
   RoleARNsResponse,
   UserRoleResponse,
   WizardBasePayload,
@@ -52,30 +52,16 @@ export function getWizardData<TResponse, TPayload extends Record<string, unknown
 export const getWizardAWSAccountIds = (
   client_id: string,
   client_secret: string,
-  signal?: AbortSignal,
-  additionalData?: Record<string, unknown>
+  signal?: AbortSignal
 ): Promise<AwsAccountIdsResponse> =>
-  getWizardData<AwsAccountIdsResponse, Record<string, unknown>>(
-    client_id,
-    client_secret,
-    '/aws-account-ids',
-    signal,
-    additionalData
-  )
+  getWizardData<AwsAccountIdsResponse>(client_id, client_secret, '/aws-account-ids', signal)
 
 export const getWizardAwsBillingAccounts = (
   client_id: string,
   client_secret: string,
-  signal?: AbortSignal,
-  additionalData?: Record<string, unknown>
+  signal?: AbortSignal
 ): Promise<OrganizationQuotaResponse> =>
-  getWizardData<OrganizationQuotaResponse, Record<string, unknown>>(
-    client_id,
-    client_secret,
-    '/aws-billing-accounts',
-    signal,
-    additionalData
-  )
+  getWizardData<OrganizationQuotaResponse>(client_id, client_secret, '/aws-billing-accounts', signal)
 
 export const getWizardOIDCConfigs = (
   client_id: string,
@@ -123,40 +109,20 @@ export const getWizardRoleARNs = (
   client_id: string,
   client_secret: string,
   signal?: AbortSignal,
-  additionalData?: Record<string, unknown>
+  additionalData?: AwsAccountPayload
 ): Promise<RoleARNsResponse> =>
-  getWizardData<RoleARNsResponse, Record<string, unknown>>(
-    client_id,
-    client_secret,
-    '/sts-role-arns',
-    signal,
-    additionalData
-  )
+  getWizardData<RoleARNsResponse, AwsAccountPayload>(client_id, client_secret, '/sts-role-arns', signal, additionalData)
 
 export const getWizardOCMRoleARN = (
   client_id: string,
   client_secret: string,
   signal?: AbortSignal,
-  additionalData?: Record<string, unknown>
+  additionalData?: AwsAccountPayload
 ): Promise<OCMRoleResponse> =>
-  getWizardData<OCMRoleResponse, Record<string, unknown>>(
-    client_id,
-    client_secret,
-    '/sts-ocm-role',
-    signal,
-    additionalData
-  )
+  getWizardData<OCMRoleResponse, AwsAccountPayload>(client_id, client_secret, '/sts-ocm-role', signal, additionalData)
 
 export const getWizardUserRoleARN = (
   client_id: string,
   client_secret: string,
-  signal?: AbortSignal,
-  additionalData?: Record<string, unknown>
-): Promise<UserRoleResponse> =>
-  getWizardData<UserRoleResponse, Record<string, unknown>>(
-    client_id,
-    client_secret,
-    '/sts-user-role',
-    signal,
-    additionalData
-  )
+  signal?: AbortSignal
+): Promise<UserRoleResponse> => getWizardData<UserRoleResponse>(client_id, client_secret, '/sts-user-role', signal)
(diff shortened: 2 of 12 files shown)


backend/src/routes/rosaWizardApi.ts
```diff
@@ -96,3 +101,128 @@ export async function getAwsBillingAccountIds(req: Http2ServerRequest, res: Http
...
+
+          // get current account and ID
+          const accountPath = `${API_URL}/api/accounts_mgmt/v1/current_account`
+          const request = (await jsonRequest(accountPath, accessTokenSSO).catch((err: Error) => {
+            logger.error({ msg: 'Failed to fetch account', error: err.message })
+            return { error: err.message }
+          })) as OrgType
+
+          const userRolesPath = `${API_URL}/api/accounts_mgmt/v1/accounts/${request.id}/labels/sts_user_role
+`
```
kelvah: There's a newline in the url string.

backend/src/routes/rosaWizardApi.ts
```diff
@@ -96,3 +101,128 @@ export async function getAwsBillingAccountIds(req: Http2ServerRequest, res: Http
...
+      req.on('end', async () => {
+        try {
+          data = chucks.join()
+          const body = JSON.parse(data) as Payload
+
+          const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
+
+          // get current account and ID
+          const accountPath = `${API_URL}/api/accounts_mgmt/v1/current_account`
+          const request = (await jsonRequest(accountPath, accessTokenSSO).catch((err: Error) => {
```
kelvah: If this request fails, it makes no sense to attempt the user role call later, it will fail anyway with no id.

You can change this to another try/catch block, you log the error and early return if the request fails instead of proceeding.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,147 @@
...
+  }, [])
+
+  const refetchAll = useCallback(
+    async (accountId: string) => {
+      await Promise.all([fetch(accountId), rolesQuery.refetch(), ocmRoleQuery.refetch(), userRoleQuery.refetch()])
+    },
+    // Does not need anything else
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [rolesQuery.refetch, ocmRoleQuery.refetch, userRoleQuery.refetch]
+  )
```
kelvah, line 132: I'm not sure I get the whole picture here. Can you elaborate more?

`refetchAll` is calling `fetch(accountId)` which is a state update. But when the `accountId` changes, the queries will automatically fetch again because it's one of their keys.

If `refetchAll` is only meant to be manually triggered, it could include only `rolesQuery.refetch(), ocmRoleQuery.refetch(), userRoleQuery.refetch()`.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,147 @@
...
+      roleVersion: role.version ?? '',
+    },
+    supportRole: role.Support ? [{ label: role.Support, value: role.Support }] : [],
+    workerRole: role.Worker ? [{ label: role.Worker, value: role.Worker }] : [],
+  }))
+}
+
+const normalizedAWSAccountRole = (arrayOfRoleItems: AccountRoleARN[], prefix: any): NormalizedAccountRole =>
+  arrayOfRoleItems.reduce(
+    (roleObj: NormalizedAccountRole, { type, arn, roleVersion, ...otherRoleAttributes }: any) => ({
```
kelvah: Why are you dropping the type information here? 
`prefix` is a string. The de-structured param of the callback is an `AccountRoleARN`

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/queryKeyFactory.ts
```diff
@@ -5,4 +5,7 @@ export const rosaWizardKeys = {
   all: [ROSA_HCP_WIZARD_QUERY_KEY],
   awsInfrastructureAccounts: (id: string) => [...rosaWizardKeys.all, id, 'aws-account-ids'],
   awsBillingAccounts: (id: string) => [...rosaWizardKeys.all, id, 'aws-billing-ids'],
+  rolesArns: (id: string, awsAccountId?: string) => [...rosaWizardKeys.all, id, awsAccountId, 'roles-arns'],
+  ocmRoleArn: (id: string, awsAccountId?: string) => [...rosaWizardKeys.all, id, awsAccountId, 'ocm-role-arn'],
+  userRoleArn: (id: string) => [...rosaWizardKeys.all, id, 'user-role-arn'],
```
kelvah: Nitpick, the `id` you added previously now looks vague compared to `awsAccountId`. Can you make it more descriptive?

frontend/src/lib/rosa-hcp-api.ts
```diff
@@ -63,3 +71,45 @@ export const getWizardAwsBillingAccounts = (
...
     additionalData
   )
+
+export const getWizardRoleARNs = (
+  client_id: string,
+  client_secret: string,
+  signal?: AbortSignal,
+  additionalData?: Record<string, unknown>
+): Promise<RoleARNsResponse> =>
+  getWizardData<RoleARNsResponse, Record<string, unknown>>(
```
kelvah: Nitpick, the point of adding a generic for the payload of `getWizardData` is to be able to explain what's needed when using it, and when calling `getWizardRoleARNs` in this case.

In this example `additionalData` should be `{ aws_accounts_id: string }`.

The same consideration applies to the other functions.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,147 @@
...
+  const secretRef = useRef(selectedSecret)
+  secretRef.current = selectedSecret
+
+  const [rolesQuery, ocmRoleQuery, userRoleQuery] = useQueries({
+    queries: [
+      {
+        queryKey: rosaWizardKeys.rolesArns(selectedSecret.client_id, awsAccountId),
+        queryFn: async ({ signal }: { signal?: AbortSignal }) => {
+          const secret = secretRef.current
+          if (!secret || !awsAccountId) return { roles: [], error: null }
```
kelvah, line 75: Why do you need to do this check? Aren't the secret and the account id already included inside the `enabled` config of the query?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: Sure. We have in the wizard
`            <WizSelect<ROSAHCPCluster>
              schema={clusterValidationSchema}
              apiError={roles.error}
              isLoading={roles.isFetching}
              onRefresh={() =>
                void (awsInfrastructureAccount && roles.fetch(awsInfrastructureAccount))
              }
              labelHelp={
                <>
                  {rp.installerHelpLead}{' '}
                  <ExternalLink href={links.ROSA_ROLES_LEARN_MORE}>
                    {rp.installerLearnMoreLink}
                  </ExternalLink>
                </>
              }
              name="installer_role_arn"
              options={installerRoleOptions}
              data-testid="installer-role-select"
            />`
            
            Fetch expects awsInfra account id, if I am not mistaken it is for manual retrigerring of the field (refresh button) along with automatic refetch when user changes aws account in the dropdown. If I remember correctly Kim coded it that if user changes AWS account it will auto refetch, not just clear the field. But we also provide a button to refetch roles (in case roles were missing, user created them via CLI and has to do refetch)

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/queryKeyFactory.ts
davidaznaur: Sounds good, will update

frontend/src/lib/rosa-hcp-api.ts
davidaznaur: No problem, will update

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: Its a runtime guard. Refetch bypasses enabled while enabled just tells RQ when to auto fetch.

backend/src/routes/rosaWizardApi.ts
davidaznaur: Strange how it passed prettier...

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: Yep, copy pasting issue from OCM... Will fix it as well

backend/src/routes/rosaWizardApi.ts
davidaznaur: Interestingly, it is prettier that causes it to add a line. I assume limit of the amount of chars per line?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,147 @@
...
+        ? 'User role was not found'
+        : null
+      : null
+  const rolesError = rolesQuery.data?.error ?? (rolesQuery.error instanceof Error ? rolesQuery.error.message : null)
+
+  return {
+    data: Array.isArray(rolesQuery.data?.roles) ? rolesQuery.data.roles : [],
+    ocmRole: ocmRoleQuery.data ?? null,
+    userRole: userRoleQuery.data ?? null,
+    isLoading: rolesQuery.isLoading || ocmRoleQuery.isLoading,
```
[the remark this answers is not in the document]
davidaznaur: That is correct.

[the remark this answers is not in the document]
davidaznaur: It shouldn't because we do not care about loading of userRoles. We only care if they exist or not at the end of the day. isLoading for userRoles won't have any effect in a place where this api will be called.

backend/src/routes/rosaWizardApi.ts
kelvah: I guess so. I see the problem is gone now.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
kelvah: TIL refetch takes precedence over enabled. I wouldn't have expected it. Thanks.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,153 @@
...
+  const { refetch: refetchRoles } = rolesQuery
+  const { refetch: refetchOcmRole } = ocmRoleQuery
+  const { refetch: refetchUserRole } = userRoleQuery
+
+  const refetchAll = useCallback(
+    async (accountId: string) => {
+      if (accountId !== awsAccountId) {
+        fetch(accountId)
+      } else {
+        await Promise.all([fetch(accountId), refetchRoles(), refetchOcmRole(), refetchUserRole()])
```
kelvah: Thanks for the fix. I think you need to remove `fetch(accountId)` from `Promise.all() now`

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,153 @@
...
+    [awsAccountId, fetch, refetchRoles, refetchOcmRole, refetchUserRole]
+  )
+
+  const ocmRoleError = ocmRoleQuery.error instanceof Error ? ocmRoleQuery.error.message : null
+  const userRoleError =
+    userRoleQuery.error instanceof Error
+      ? userRoleQuery.error.message === "AccountLabel with key='sts_user_role' not found"
+        ? 'User role was not found'
+        : null
+      : null
```
KevinFCormier, line 139: SonarQube doesn't like the nested ternary:
```suggestion
  const userRoleError =
    userRoleQuery.error instanceof Error &&
    userRoleQuery.error.message === "AccountLabel with key='sts_user_role' not found"
      ? 'User role was not found'
      : null
```

KevinFCormier (changes requested):

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
KevinFCormier: Also, why are we hiding the error message if it does not match this exact case?

davidaznaur: The user roles are not used in the wizard per se. They are only being checked if they exist or not, for a specific error message.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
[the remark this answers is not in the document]
KevinFCormier: So @davidaznaur does this mean `isLoading` could switch to `false` but then `userRoleError` could change to indicate an error later on? Shouldn't the user be able to trust the results as soon as `isLoading` is `false`?

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,152 @@
...
+          if (!secret || !awsAccountId) return null
+          const response = await getWizardUserRoleARN(secret.client_id, secret.client_secret, signal)
+          return response
+        },
+        enabled: !!selectedSecret && !!awsAccountId,
+        retry: false,
+      },
+    ],
+  })
+  const fetch = useCallback(async (accountId: string) => {
```
KevinFCormier, line 115: 7. frontend/src/routes/.../useFetchRolesARNs.ts:104 fetch function is async but never awaits anything — The function is declared async but only calls setAwsAccountId(accountId), which is synchronous. The async keyword is misleading.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,152 @@
...
+  const refetchAll = useCallback(
+    async (accountId: string) => {
+      if (accountId !== awsAccountId) {
+        fetch(accountId)
+      } else {
+        await Promise.all([refetchRoles(), refetchOcmRole(), refetchUserRole()])
+      }
+    },
+    [awsAccountId, fetch, refetchRoles, refetchOcmRole, refetchUserRole]
+  )
```
KevinFCormier, line 132: 8. frontend/src/routes/.../useFetchRolesARNs.ts:108 refetchAll is async but the fetch branch doesn't await anything — When accountId !== awsAccountId, it calls fetch(accountId) which just sets state. The async keyword is misleading for that branch. More importantly, calling fetch (which changes state) won't trigger an immediate refetch in the same call — React Query will re-run queries when enabled becomes true on the next render, but the caller gets a resolved promise that doesn't wait for data.
Suggestion: Document this behavior clearly or consider if the API should be different. At minimum, remove unnecessary async.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
```diff
@@ -0,0 +1,152 @@
...
+        queryKey: rosaWizardKeys.userRoleArn(selectedSecret.client_id),
+        queryFn: async ({ signal }: { signal?: AbortSignal }) => {
+          const secret = secretRef.current
+          if (!secret || !awsAccountId) return null
+          const response = await getWizardUserRoleARN(secret.client_id, secret.client_secret, signal)
+          return response
+        },
+        enabled: !!selectedSecret && !!awsAccountId,
+        retry: false,
+      },
```
KevinFCormier, line 112: 9. frontend/src/routes/.../useFetchRolesARNs.ts:93 userRoleQuery checks !!awsAccountId in enabled and queryFn guard, but its query key doesn't include awsAccountId — The userRoleArn query key is [..., client_id, 'user-role-arn'] (no account id), but the query won't fire without awsAccountId being set. This creates an inconsistency: React Query caches this query independent of the account id, but the hook behavior depends on it. If the user switches AWS accounts, the cached user role won't be refetched because the key didn't change.
Suggestion: Either remove the awsAccountId guard from the user role query (since user roles aren't AWS-account-specific), or add awsAccountId to the query key if they are.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/constants/types.ts
```diff
@@ -3,3 +3,25 @@ export type SelectedSecret = {
...
+  prefix: string
+  version: string
+  Installer?: string
+  Support?: string
+  Worker?: string
+  'Control plane'?: string
+  hcpManagedPolicies?: boolean
+  managedPolicies?: boolean
+  [type: string]: string | boolean | undefined
+}
```
KevinFCormier, line 34: 11. frontend/src/routes/.../constants/types.ts:17-27 NormalizedAccountRole uses index signature [type: string]: string | boolean | undefined — This undermines type safety. Any property access is allowed without type checking. The known properties (Installer, Support, Worker, Control plane) are better modeled without an index signature.
Suggestion: Use a discriminated approach or Record<string, string> for the dynamic role-type mapping alongside the known typed properties:
export interface NormalizedAccountRole {
  prefix: string
  version: string
  hcpManagedPolicies?: boolean
  managedPolicies?: boolean
  roleArns: Record<string, string>  // type → arn mapping
}

backend/src/routes/rosaWizardApi.ts
```diff
@@ -96,3 +101,131 @@ export async function getAwsBillingAccountIds(req: Http2ServerRequest, res: Http
...
+
+          const accessTokenSSO = await getOcmServiceToken(body.service_account_id, body.service_account_secret)
+
+          const accountPath = `${API_URL}/api/clusters_mgmt/v1/aws_inquiries/sts_account_roles`
+
+          const requestBody = {
+            account_id: body.aws_account_id,
+          }
+          const accReq = await jsonPost(accountPath, requestBody, accessTokenSSO).catch((err: Error) => {
+            logger.error({ msg: 'Error gettting account info', error: err.message })
```
KevinFCormier, line 264:
```suggestion
            logger.error({ msg: 'Error getting account info', error: err.message })
```

KevinFCormier (changes requested):
AI-code review generated a few more things that deserve consideration. Please take a look. Thanks.

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: No need. The user roles alert is specifically to inform user that they need to create user roles using cli, it is not dependent on changing aws account. User has to create user roles and then explicitly press refetch button

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: It is a contract how it is setup in the wizard. Since we cannot export awsAccountId from the wizard but somehow have to pass it into the api call

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: Same answer as for #7 it is the contract setup in the wizard. Until I import the actual wizard and start wiring them up, lets keep it. If any troubles arise, I will change

frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/queries/useFetchRolesARNs.ts
davidaznaur: The behavior is actually correct — just poorly communicated
Here's the actual flow when fetch(accountId) is called:

setAwsAccountId(accountId) queues a state update
React re-renders the component on the next tick
useQueries re-evaluates — enabled becomes true (since awsAccountId is now set)
React Query starts the three queries → isFetching becomes true
The component shows a loading state via the reactive isFetching property
Data arrives → isFetching becomes false, data populates

So the async on fetch in useFetchRoleARNs (line 114) exists to satisfy this type contract. Without async, the return type would be void, not Promise<void>, and TypeScript would complain wherever it's assigned to a RolesResource.

kelvah: @davidaznaur The way to solve this misalignment would be to fix the types on the wizard side. IIUC we are not waiting for the promise returned by the fetch callback, so it doesn't have to return a promise to begin with.

kelvah (approved):

KevinFCormier (approved):

PR #6530, opened 2026-07-21
fix(backend): replace full CRD list fetch with targeted GET in /hub route (ACM-38058)

fxiang1:
## Root Cause Analysis

The `/hub` route handler in `backend/src/routes/hub.ts` fetched the **entire CRD list** (~51MB) from the API server every 30 seconds per connected browser tab, just to check if a single CRD (`multiclusterglobalhubs.operator.open-cluster-management.io`) exists:

```
GET /apis/apiextensions.k8s.io/v1/customresourcedefinitions
```

This caused repeated massive memory allocations/deallocations in the `console-chart-console-v2` pod, leading to memory spikes.

## Fix

Replace the full CRD list fetch with a **targeted GET** for the specific CRD by name:

```
GET /apis/apiextensions.k8s.io/v1/customresourcedefinitions/multiclusterglobalhubs.operator.open-cluster-management.io
```

This reduces the response from **~51MB to ~10KB** per call. The targeted GET returns:
- **200** with `kind: CustomResourceDefinition` when the CRD exists → `isGlobalHub: true`
- **404** with `kind: Status` when it does not → `isGlobalHub: false`

Also removed the now-unused `ResourceList` type import.

## Regression Test

Added a test case that mocks a 404 response for the targeted CRD GET and verifies the handler correctly returns `isGlobalHub: false`. Updated existing test mocks to match the new targeted endpoint.

## How to Test

1. Run backend tests: `npm run test:backend` — all 338 tests pass (including 3 hub tests)
2. Run checks: `npm run check:backend` — prettier, lint, and tsc all pass
3. Start in plugin mode (`npm run plugins`) and verify the clusters page loads correctly
4. Inspect the `/hub` network response — should contain `isGlobalHub` with correct value
If the Multiclusterglobalhub operator is installed `isGlobalHub: true` otherwise `isGlobalHub: false`

## Linked Issue

Resolves [ACM-38058](https://redhat.atlassian.net/browse/ACM-38058)

[ACM-38058]: https://redhat.atlassian.net/browse/ACM-38058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of global hub availability for more reliable hub status reporting.
  * Hub status now correctly reports a non-global hub when the required resource is unavailable.

* **Tests**
  * Added coverage for scenarios where the global hub resource does not exist.
  * Updated endpoint validation for direct resource checks.

### DIFF
--- a/backend/src/routes/hub.ts
+++ b/backend/src/routes/hub.ts
@@ -7,7 +7,6 @@ import { respondInternalServerError } from '../lib/respond'
 import { getServiceAccountToken } from '../lib/serviceAccountToken'
 import { getAuthenticatedToken } from '../lib/token'
 import type { IResource } from '../resources/resource'
-import type { ResourceList } from '../resources/resource-list'
 import { getHubClusterName, getIsHubSelfManaged, getIsObservabilityInstalled, getKubeResources } from './events'
 
 interface AuthenticationResource {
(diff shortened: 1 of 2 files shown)


KevinFCormier:
@fxiang1 IIRC, I think this was coded this way as a lazy way of avoiding adjusting the ServiceAccount permissions. I think we have permission to list/watch CRDs, but not get CRDs. Can you check the installer projects?

Were you able to observe memory growth because of this pattern?

fxiang1:
> @fxiang1 IIRC, I think this was coded this way as a lazy way of avoiding adjusting the ServiceAccount permissions. I think we have permission to list/watch CRDs, but not get CRDs. Can you check the installer projects?
> 
> Were you able to observe memory growth because of this pattern?

@KevinFCormier Thanks! you're right we don't have get permission on CRDs, only list/watch.

I don't think this is causing the memory leaks but I do see a temporary memory spike. I was testing with 10000 CRDs though but then I realized it's probably not realistic to have that many. So this change won't buy us much.

zlayne:
These changes look good to me, assuming we are okay with adding GET permissions for CRDs.

fxiang1:
@zlayne the GET permission is added with these PRs
https://github.com/stolostron/multiclusterhub-operator/pull/4454
https://github.com/stolostron/backplane-operator/pull/3696

zlayne (approved):

PR #6443, opened 2026-07-08
Ignore placement debug for this test

KevinFCormier:
Seeing this failure frequently on PRs:

```
[frontend] FAIL src/wizards/Governance/Policy/policyWizard.test.tsx (48.527 s)
[frontend]   ● Policy wizard › default tolerations persist after switching to existing and back to new placement
[frontend] 
[frontend] 
[frontend] 
[frontend]     !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
[frontend]     !!!!!!!!!!!!!!!! MISSING NOCK(S) !!!!!!!!!!!!!!!!!!!!!!!!
[frontend] 
[frontend]      there was no nocks that matched these request(s):
[frontend] 
[frontend] 
[frontend]     //---create ---
[frontend]     const createUnknown1 = { req: {} }
[frontend] 
[frontend] 
[frontend] 
[frontend]         nockCreate(createUnknown1.req, createUnknown1.res)    // create
[frontend] 
[frontend]       229 | async function setupAfterEach(): Promise<void> {
[frontend]       230 |   await new Promise((resolve) => setTimeout(resolve, 100))
[frontend]     > 231 |   expect(missingNocks).hasNoMissingNocks()
[frontend]           |                        ^
[frontend]       232 |   expect(missingNocks).hasNoPendingNocks()
[frontend]       233 | }
[frontend]       234 |
[frontend] 
[frontend]       at Object.setupAfterEach (src/setupTests.ts:231:24)
```

## Summary by CodeRabbit

* **Tests**
  * Updated a wizard test setup to include the same placement-related mock configuration used by nearby tests, improving consistency and stability of the test run.

PR #6535, opened 2026-07-24
Trim whitespace when splitting security-group annotations

sungupta-llm:
## Summary

The Governance Overview Standards/Categories/Controls cards (and the policy
drill-down sidebar behind them) group policies by splitting the
`policy.open-cluster-management.io/{standards,categories,controls}`
annotation on `,`:

```ts
const names = annotation.split(',')
for (const name of names) {
  let v = clusterViolations[name]   // raw, untrimmed string used as the aggregation key
  ...
}
```

`String.split(',')` leaves a leading space on every token after the first
(e.g. `"A, B".split(',')` → `["A", " B"]`). If the same logical value (e.g.
`"NIST SP 800-53"`) is the first item in one policy's annotation but a later
item in another policy's annotation, the two policies produce different map
keys (`"NIST SP 800-53"` vs `" NIST SP 800-53"`) and render as two separate
rows on the card instead of aggregating into one.

The same unfixed split exists in the drill-down sidebar filter
(`SecurityGroupPolicySummarySidebar.tsx`), which does a strict equality
check (`name === violation.name`) against the untrimmed token, so the
sidebar's policy list can also fail to match policies whose annotation
formatting differs from the row that was clicked.

## Fix

Trim each token after splitting, in both `useSecurityGroupViolations`
(`Overview.tsx`) and the matching filter in
`SecurityGroupPolicySummarySidebar.tsx`, so aggregation and drill-down
filtering key off the same normalized value regardless of comma-spacing.

## Reproduction (single managed cluster is sufficient)

1. Create Policy A with annotation:
   `policy.open-cluster-management.io/standards: "NIST SP 800-53, PCI-DSS 4.0"`
2. Create Policy B with annotation:
   `policy.open-cluster-management.io/standards: "CIS OpenShift Benchmark, PCI-DSS 4.0, NIST SP 800-53"`
3. Propagate both to any managed cluster, wait for Compliant status.
4. Governance Overview → Standards card shows "NIST SP 800-53" as **two**
   separate rows instead of one aggregated row.

## Test plan

- Added a regression test to `Overview.test.tsx` that renders the Standards
  card with two policies carrying the same standard in different list
  positions, and asserts a single aggregated row.
- Verified the new test **fails** without the fix (`Expected: 1, Received: 2`)
  and **passes** with it.
- `eslint --max-warnings=0` clean on all changed files.

Reported in Red Hat case 04500068.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Governance overview aggregation by trimming whitespace in comma-separated standards and security group names from annotations.
  * Fixed policy matching so clicked violation names correctly align with standards even when annotation tokens include extra spaces.
* **Tests**
  * Added regression coverage to verify standards are deduplicated after trimming and that the sidebar includes the expected policy when names match non-first tokens.

### DIFF
--- a/frontend/src/routes/Governance/overview/Overview.test.tsx
+++ b/frontend/src/routes/Governance/overview/Overview.test.tsx
@@ -16,6 +16,7 @@ import {
 import GovernanceOverview from './Overview'
 import userEvent from '@testing-library/user-event'
 import { defaultContext, PluginDataContext } from '../../../lib/PluginDataContext'
+import { Policy, PolicyApiVersion, PolicyKind } from '../../../resources'
 
 describe('Overview Page', () => {
   beforeEach(async () => nockIgnoreApiPaths())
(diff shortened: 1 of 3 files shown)


Randy424 (dismissed):
/lgtm

Randy424 (approved):
/lgtm

sungupta-llm:
@Randy424 It seems its pending on 1 more test which require your approval.

Randy424:
@sungupta-llm it looks like we are still running into some test failures ([frontend] FAIL src/wizards/Governance/Policy/policyWizard.test.tsx (51.115 s)) ...Could you review this unit test and confirm whether this failure is related to your changes?

sungupta-llm:
@Randy424 No code overlap: PolicyWizard.tsx/policyWizard.test.tsx have zero references to Overview.tsx or SecurityGroupPolicySummarySidebar.tsx (the only files this PR touches besides the test). 
I saw PR #6443 (merged 2026-07-09) fixed the identical MISSING NOCK(S) failure for a different test in this same file (default tolerations persist after switching to existing and back to new placement) — same error signature, same setupAfterEach stack trace at src/setupTests.ts:231. That fix added one missing mock line for that specific test.

cc @KevinFCormier

PR #6597, opened 2026-07-28
ACM-39212 fix(backend): use GET /api with body drain instead of HEAD [release-2.16]

Randy424:
## Summary
- Fixes 405 regression from the original HEAD /api approach — `HEAD /api` is rejected by some cluster API server / proxy configurations
- Switches `isAuthenticated()` to `GET /api` with explicit body drain (`.resume()`)
- Preserves the memory leak fix: `/api` response is ~200 bytes (vs several MB for `/apis`), drained immediately so sockets return to the keepAlive pool

## Changes
- `backend/src/lib/token.ts`: GET /api with body drain, returns `Promise<number>`
- 12 test files: nock interceptors updated from `.head('/api')` to `.get('/api')`

## Test plan
- [x] All backend unit tests pass (261 tests)

Backport of #6600 to release-2.16

### DIFF
--- a/backend/src/lib/authenticated.ts
+++ b/backend/src/lib/authenticated.ts
@@ -7,9 +7,8 @@ export function authenticated(req: Http2ServerRequest, res: Http2ServerResponse)
   const token = getToken(req)
   if (!token) return unauthorized(req, res)
   isAuthenticated(token)
-    .then((response) => {
-      res.writeHead(response.status).end()
-      void response.blob()
+    .then((status) => {
+      res.writeHead(status).end()
     })
     .catch(catchInternalServerError(res))
 }
--- a/backend/src/lib/token.ts
+++ b/backend/src/lib/token.ts
@@ -28,15 +28,18 @@ export function getToken(req: Http2ServerRequest): string | undefined {
   return token
 }
 
-// HEAD /api returns headers only — no response body — so no drain is needed and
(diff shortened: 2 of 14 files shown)


Randy424:
Rebased onto the latest `release-2.16` to resolve the merge conflict.

**Conflict:** `backend/test/routes/ansibletower.test.ts` — `release-2.16` had since picked up the glasswing RBAC-secret fix, which added `nockCredentialSecret(TOWER_HOST)` calls to several tests still mocking the auth check. This PR's first two commits (original `HEAD /api` attempt, then the `GET /api` correction) each touched the same test blocks, so both replayed with conflicts; the third (prettier cleanup) applied cleanly.

**Resolution:** Same approach used for the release-2.14 (#6671) and release-2.15 (#6598) backports, itself sourced from how `main` resolved this code: for each conflicting test, kept upstream's new `nockCredentialSecret` additions and applied only this PR's intended line-level change (mock method) on top.

**Additional fix:** A handful of `.get('/apis')` auth-check mocks elsewhere in the same file weren't touched by either commit (new tests `release-2.16` added independently) but would have failed silently once `isAuthenticated()` calls `GET /api` — `nock.disableNetConnect()` is on, so an unmatched request throws instead of returning the mocked response. Converted those too for consistency.

Verified: all 266 backend tests pass, `prettier`/`lint`/`tsc` all clean.

PR #6598, opened 2026-07-28
ACM-38826 fix(backend): use GET /api with body drain instead of HEAD [release-2.15]

Randy424:
## Summary
- Fixes 405 regression from the original HEAD /api approach — `HEAD /api` is rejected by some cluster API server / proxy configurations
- Switches `isAuthenticated()` to `GET /api` with explicit body drain (`.resume()`)
- Preserves the memory leak fix: `/api` response is ~200 bytes (vs several MB for `/apis`), drained immediately so sockets return to the keepAlive pool

## Changes
- `backend/src/lib/token.ts`: GET /api with body drain, returns `Promise<number>`
- 12 test files: nock interceptors updated from `.get('/apis')` to `.get('/api')`

## Test plan
- [x] All backend unit tests pass (214 tests)

Backport of #6600 to release-2.15

### DIFF
--- a/backend/src/lib/authenticated.ts
+++ b/backend/src/lib/authenticated.ts
@@ -7,9 +7,8 @@ export function authenticated(req: Http2ServerRequest, res: Http2ServerResponse)
   const token = getToken(req)
   if (!token) return unauthorized(req, res)
   isAuthenticated(token)
-    .then((response) => {
-      res.writeHead(response.status).end()
-      void response.blob()
+    .then((status) => {
+      res.writeHead(status).end()
     })
     .catch(catchInternalServerError(res))
 }
--- a/backend/src/lib/token.ts
+++ b/backend/src/lib/token.ts
@@ -28,15 +28,14 @@ export function getToken(req: Http2ServerRequest): string | undefined {
   return token
 }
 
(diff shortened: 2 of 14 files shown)


Randy424:
Rebased onto the latest `release-2.15` to resolve the merge conflict.

**Conflict:** `backend/test/routes/ansibletower.test.ts` — `release-2.15` had since picked up the glasswing RBAC-secret fix, which added `nockCredentialSecret(TOWER_HOST)` calls and new tests to this file. This PR's two commits (original `HEAD /api` attempt, then the `GET /api` correction) each touched the same test blocks, so both replayed with conflicts.

**Resolution:** Followed the same approach used for the release-2.14 backport (#6671), which itself mirrored how `main` resolved this code: for each conflicting test, kept upstream's new `nockCredentialSecret`/RBAC-field additions and applied only this PR's intended line-level change (rename + mock method) on top.

**Additional fix:** Several `.get('/apis')` auth-check mocks in the same file were never touched by either commit's diff (new tests `release-2.15` added independently) but would have failed silently once `isAuthenticated()` calls `GET /api` — `nock.disableNetConnect()` is on, so an unmatched request throws instead of returning the mocked response. Converted those too for consistency.

Verified: all 227 backend tests pass, lint clean.

PR #6648, opened 2026-08-06
ACM-39327: Fix non-admin SSE OOM under large inventory (release-2.13)

Ginxo:
## Summary

- Manual backport of the RBAC/SelfSubjectRulesReview fix from #6638 to `release-2.13` (fast-forwards to `backplane-2.8`).
- After cluster-scoped `list` is denied, use one `SelfSubjectRulesReview` per token/kind instead of O(namespaces) namespaced list + O(N) per-object SSARs, which OOMs/hangs restricted users under large inventory (`MOCK_CLUSTERS` / real fleets).
- Harden the access cache: SHA-256 token keys, verb in cache keys, TTL cleanup, per-token entry cap.
- SSRR HTTP failures are marked `unavailable` and fall back to SSAR (not cached as deny-all).

**Not ported** (not present on 2.13): event-body compression, `meta` on cached events, and filter-before-inflate in `server-side-events.ts`. Those half of #6638 do not apply here; the O(N) SSAR path does.

Related to https://github.com/stolostron/console/pull/6638

## Test plan

- [x] `backend` unit tests for access cache + SSRR short-circuit (`npm test -- --testPathPattern=test/routes/events.test.ts`)
- [x] `npm run check` in backend
- [x] Local: `MOCK_CLUSTERS=1000`, kubeadmin vs restricted/`none` on Inventory; confirm backend RSS/CPU stay bounded
- [ ] Confirm admin Inventory still populates clusters normally
- [ ] Confirm `none` completes SSE load without backend hang

### Test support

I normally use
```
watch -n 1 "ps -p $(fuser 4000/tcp 2>/dev/null) -o pid,%cpu,%mem,rss,cmd"
```
for checking the CPU/MEM consumption on the local environment

### Test results

I can confirm the situation is even worse for 2.13 and the memory grows up out of control for non-admin users. This PR is totally mandatory for 2.13

#### Image generation
```
git checkout bug/ACM-39327-release-2.13

USERNAME=YOUR_USER_NAME
TAG="acm39327-6648-$(date +%Y%m%d-%H%M%S)"
IMAGE="quay.io/${USERNAME}/console-mce:${TAG}"

podman login registry.redhat.io
podman login quay.io

podman build --arch amd64 -f Containerfile.mce.konflux -t "${IMAGE}" .
podman push "${IMAGE}"
```

#### Deployment

```
./scripts/patch-deployment.sh "${TAG}" quay.io/${USERNAME}/console-mce
oc -n multicluster-engine rollout status deployment/console-mce-console
# Verify                                                     
oc get deployment console-mce-console -n multicluster-engine \
  -o jsonpath='{.spec.template.spec.containers[0].image}{"\n"}'
oc get pods -n multicluster-engine -l app=console-mce

# Additional verification
POD=$(oc get pods -n multicluster-engine -l app=console-mce \
  -o jsonpath='{.items[0].metadata.name}')

oc get pod "$POD" -n multicluster-engine \
  -o jsonpath='{.spec.containers[0].image}{"\n"}'
```
and the output should be something like `quay.io/emingora/console-mce:acm39327-6648-20260818-105951`

#### Test execution

* uncompress attached file [acm-39327-repro.zip](https://github.com/user-attachments/files/31174751/acm-39327-repro.zip) to `/tmp`
```bash
cd /tmp/acm-39327-repro
source ./env.sh
./gate.sh
```

we should see something like

```
./gate.sh 
GATE_A: MC=25 MCI=25 MCA=200 POL=3 NS=25  → need 25/25/200/3/25
GATE_B: CD=25 MP=50 CP=25 SEC=150           → need 25/50/25/≥100
GATE_C: IE=25 AG=250 BMH=250 NMC=250 ACI=25 AM=0 → need 25/250/250/250/25/0
GATE_D: PR=25 CSR=25                          → need 25/25
LABELED_TOTAL=1378 (target ≥1329 or ≥1079 without AgentMachine)
GATE: PASS
```
otherwise execute
```
source ./env.sh 
./apply-layer-d.sh
./gate.sh 
```

Execute the resource watcher in a terminal
```
WORKDIR=/tmp/acm-39327-repro
START_TS=$(date +%Y%m%dT%H%M%S)
LOG="$WORKDIR/acm39327-console-mce-${START_TS}"
MCE_NS=$(oc get pods -A -l app=console-mce -o jsonpath='{.items[0].metadata.namespace}')

printf 'time\tsource\tpod\tcpu\tmemory\n' > "$LOG"

( while true; do
    TS=$(date -Is)
    while read -r name cpu mem; do
      [[ "$name" == "NAME" || -z "$name" ]] && continue
      printf '%s\ttop\t%s\t%s\t%s\n' "$TS" "$name" "$cpu" "$mem" >> "$LOG"
    done < <(oc adm top pod -n "$MCE_NS" -l app=console-mce --no-headers 2>/dev/null)
    sleep 2
  done ) & TOP_PID=$!

( while true; do
    TS=$(date -Is)
    POD=$(oc get pods -n "$MCE_NS" -l app=console-mce -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)
    if [[ -n "$POD" ]]; then
      BYTES=$(oc exec -n "$MCE_NS" "$POD" -- sh -c \
        'cat /sys/fs/cgroup/memory.current 2>/dev/null || cat /sys/fs/cgroup/memory/memory.usage_in_bytes' 2>/dev/null)
      [[ -n "$BYTES" ]] && printf '%s\tcgroup\t%s\t-\t%.1fMi\n' "$TS" "$POD" "$(awk -v b="$BYTES" 'BEGIN{print b/1024/1024}')" >> "$LOG"
    fi
    sleep 2
  done ) & CGROUP_PID=$!

echo "Watchers: top=$TOP_PID cgroup=$CGROUP_PID  log=$LOG"
tail -f "$LOG"
```

> Note: kill it once it is over by executing `kill $TOP_PID $CGROUP_PID`

and then start the test execution on another terminal
```
export CONSOLE_URL=$(oc get route -A -o json | jq -r '.items[] | select(.metadata.name|test("console")) | "https://\(.spec.host)"' | head -1)
export ACM_NONE_PASS='Acm39327!'
export ACM_IDP_NAME='acm39327-htpasswd'
export ACM_NONE_USERS="$(seq -w 1 20 | tr '\n' ' ')"

node ./acm39327-host-inventory.mjs
```

### DIFF
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -194,49 +194,59 @@ function enforceAccessCacheEntryCap(tokenCache: Record<string, { time: number; p
 
 let accessCacheCleanupTimer: NodeJS.Timeout | undefined
 
+function expireTimedEntries<T extends { time: number }>(cache: Record<string, T>, cutoffTime: number) {
+  for (const key in cache) {
+    if (cache[key].time < cutoffTime) {
+      delete cache[key]
+    }
+  }
+}
+
+/** Prune one token's SSAR entries; returns newest remaining time, or undefined if the token was removed. */
+function pruneAccessCacheToken(
+  token: string,
+  tokenCache: Record<string, { time: number; promise: Promise<boolean> }>,
+  cutoffTime: number
+): number | undefined {
+  let newestTime = 0
+
+  for (const key in tokenCache) {
+    if (tokenCache[key].time < cutoffTime) {
+      delete tokenCache[key]
+    } else if (tokenCache[key].time > newestTime) {
+      newestTime = tokenCache[key].time
+    }
+  }
+
+  if (Object.keys(tokenCache).length === 0) {
+    delete accessCache[token]
+    return undefined
+  }
+
+  enforceAccessCacheEntryCap(tokenCache)
+  return newestTime
+}
+
 export function cleanupAccessCache() {
-  const now = Date.now()
-  const cutoffTime = now - ACCESS_CACHE_TTL
+  const cutoffTime = Date.now() - ACCESS_CACHE_TTL
   const tokenStats: Array<{ token: string; newestTime: number }> = []
 
   for (const token in accessCache) {
-    const tokenCache = accessCache[token]
-    let newestTime = 0
-
-    for (const key in tokenCache) {
-      if (tokenCache[key].time < cutoffTime) {
-        delete tokenCache[key]
-      } else if (tokenCache[key].time > newestTime) {
-        newestTime = tokenCache[key].time
-      }
-    }
-
-    if (Object.keys(tokenCache).length === 0) {
-      delete accessCache[token]
-    } else {
-      enforceAccessCacheEntryCap(tokenCache)
+    const newestTime = pruneAccessCacheToken(token, accessCache[token], cutoffTime)
+    if (newestTime !== undefined) {
       tokenStats.push({ token, newestTime })
     }
   }
 
-  for (const key in subjectRulesCache) {
-    if (subjectRulesCache[key].time < cutoffTime) {
-      delete subjectRulesCache[key]
-    }
-  }
-  for (const key in kindGetAccessCache) {
-    if (kindGetAccessCache[key].time < cutoffTime) {
-      delete kindGetAccessCache[key]
-    }
-  }
+  expireTimedEntries(subjectRulesCache, cutoffTime)
+  expireTimedEntries(kindGetAccessCache, cutoffTime)
 
-  if (tokenStats.length > ACCESS_CACHE_MAX_TOKENS) {
-    tokenStats.sort((a, b) => a.newestTime - b.newestTime)
-    const tokensToRemove = tokenStats.length - ACCESS_CACHE_MAX_TOKENS
+  if (tokenStats.length <= ACCESS_CACHE_MAX_TOKENS) return
 
-    for (let i = 0; i < tokensToRemove; i++) {
-      delete accessCache[tokenStats[i].token]
-    }
+  tokenStats.sort((a, b) => a.newestTime - b.newestTime)
+  const tokensToRemove = tokenStats.length - ACCESS_CACHE_MAX_TOKENS
+  for (let i = 0; i < tokensToRemove; i++) {
+    delete accessCache[tokenStats[i].token]
   }
 }
 
@@ -887,6 +897,28 @@ function getSubjectRules(token: string): Promise<SubjectRulesStatus> {
   return promise
 }
 
+function ruleGrantsKindAccess(
+  rule: SubjectRulesStatus['resourceRules'][number],
+  group: string,
+  resourcePlural: string,
+  accessVerbs: Set<string>
+): { allowAll: true } | { names: string[] } | null {
+  const verbs = rule.verbs ?? []
+  if (!verbs.includes('*') && !verbs.some((verb) => accessVerbs.has(verb))) return null
+
+  const groups = rule.apiGroups ?? []
+  if (!groups.includes('*') && !groups.includes(group)) return null
+
+  const resources = rule.resources ?? []
+  if (!resources.includes('*') && !resources.includes(resourcePlural)) return null
+
+  const resourceNames = rule.resourceNames
+  if (!resourceNames || resourceNames.length === 0 || resourceNames.includes('*')) {
+    return { allowAll: true }
+  }
+  return { names: resourceNames }
+}
+
 function evaluateKindGetAccess(rules: SubjectRulesStatus, kind: string, apiVersion: string): KindGetAccess {
   const group = apiVersion.includes('/') ? apiVersion.split('/')[0] : ''
   const resourcePlural = pluralize(kind.toLowerCase())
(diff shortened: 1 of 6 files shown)


backend/src/routes/events.ts
```diff
@@ -659,21 +817,161 @@ function canListNamespacedScopedKind(resource: IResource, token: string): Promis
   )
 }
 
-function canGetResource(resource: IResource, token: string): Promise<boolean> {
-  return canAccess(resource, 'get', token)
+/** Used by SSE eventFilter after list checks fail; prefers SelfSubjectRulesReview over N SSARs. */
```
KevinFCormier: This comment seems inaccurate. I only see this function used in test cases. Can we clarify the comment to indicate that?

KevinFCormier (changes requested):
There are a couple issues with the logic, so fixing them an maintaining the performance gain may be tricky.

- Checking permissions in the `default` namespace is used a proxy for checking cluster-scoped resource access, but many resources are not cluster-scoped, and there is no code guaranteeing that this check is used only for cluster-scoped resources. I granted user `user1` permission to get secrets in the `default` namespace, but then the user is able to see all Credentials. This is major security hole.
- I granted `user1` cluster-admin access for the `acm39327-mc-01` cluster. With the old code, it shows up in the cluster list. With your new code, it does not.

Other areas for improvement - caching keys include the API version, but permissions checking is only done against groups. I think you could drop the version part earlier.

Ginxo:
> There are a couple issues with the logic, so fixing them an maintaining the performance gain may be tricky.
> 
> * Checking permissions in the `default` namespace is used a proxy for checking cluster-scoped resource access, but many resources are not cluster-scoped, and there is no code guaranteeing that this check is used only for cluster-scoped resources. I granted user `user1` permission to get secrets in the `default` namespace, but then the user is able to see all Credentials. This is major security hole.
> * I granted `user1` cluster-admin access for the `acm39327-mc-01` cluster. With the old code, it shows up in the cluster list. With your new code, it does not.
> 
> Other areas for improvement - caching keys include the API version, but permissions checking is only done against groups. I think you could drop the version part earlier.

**What was wrong**

`SelfSubjectRulesReview` is namespaced. The old code always queried `default` and then applied `allow-all` to every resource of that `kind`. Kubernetes does not tell you whether a rule came from a `ClusterRoleBinding` or a `RoleBinding`, so:

* `get secrets` in `default` made every Credential visible
* `cluster-admin` in `acm39327-mc-01` never showed up in that review, so namespaced cluster objects (`ManagedClusterInfo`, `ClusterDeployment`, …) were denied, and the cluster list is the union of those, not only `ManagedCluster`

**What changed**

* Namespaced resources: `SelfSubjectRulesReview` in that resource’s namespace, cached per `token + namespace`
* Cluster-scoped resources: `default` is only a probe. Empty rules still deny with no `SelfSubjectAccessReview` (none-user fast path). Unrestricted grants are confirmed with `SelfSubjectAccessReview` so a `RoleBinding` in `default` cannot impersonate cluster access
* Cache keys use API group + plural, not apiVersion (`Placement` v1alpha1 / v1beta1 share one entry)
* `eventFilter` now calls `canGetResource`, which answers @KevinFCormier "this is only used in tests" comment

backend/src/routes/eventsAccess.ts
```diff
@@ -0,0 +1,284 @@
...
+ * Namespaced resources are reviewed in the resource's namespace (cached per token+namespace).
+ * Cluster-scoped resources use a probe-namespace review only as a negative/named-binding cache;
+ * unrestricted grants from that probe are confirmed with SSAR so RoleBindings in `default`
+ * cannot impersonate cluster-scoped access.
+ */
+export function canGetResource(resource: AccessResource, token: string): Promise<boolean> {
+  return resolveKindGetAccess(resource, token).then((access) => {
+    // Probe-namespace SSRR cannot distinguish RoleBindings from ClusterRoleBindings.
+    // Confirm unrestricted cluster-scoped grants with SSAR to close the default-ns proxy hole.
+    if (!isNamespacedResource(resource) && access.type === 'allow-all') {
```
KevinFCormier: This blocks access for the 'allow-all' case but not the 'allow-names' case. This can be exploited.

Looking at `user1`, I added a RoleBinding `user1-read-acm39327-mc-02` in the default namespace pointing to a Role that seems to let the user read ManagedCluster acm39327-mc-02. But since these are Role/RoleBinding and not ClusterRole/ClusterRoleBinding, the user does not actually have that permission:

```
% oc get managedcluster acm39327-mc-02
NAME             HUB ACCEPTED   MANAGED CLUSTER URLS   JOINED   AVAILABLE   AGE
acm39327-mc-02   true                                           Unknown     6d19h
% oc get managedcluster acm39327-mc-02 --as user1
Error from server (Forbidden): managedclusters.cluster.open-cluster-management.io "acm39327-mc-02" is forbidden: User "user1" cannot get resource "managedclusters" in API group "cluster.open-cluster-management.io" at the cluster scope
```

But the UI shows it:

Image

Would it be simpler to bypass `resolveKindGetAccess` completely for  cluster-scoped resources? I don't think there are any cases where the information in the SSRR is useful for cluster-scoped resources.

Ginxo: the allow-names hole was real, that's an amazing catch from your side, and I agree bypassing resolveKindGetAccess entirely for cluster-scoped resources is not an option here.

That path would go straight to per-object SSAR for every cluster-scoped get, including the none-user / empty-rules case. That’s exactly what ACM-39327 was trying to fix. Full bypass would undo the main purpose of this change.

Instead I went with a middle ground:

* **Cluster-scoped kinds** (explicit catalog, not metadata.namespace) (not happy with this solution): SSRR in default is only a negative cache (empty/incomplete-empty -> deny-all, no SSAR).
* **Any non-deny-all cluster-scoped result** (allow-all, allow-names, incomplete, evaluationError with partial rules) -> SSAR get confirm, so a RoleBinding in default can’t impersonate cluster access (your user1 / acm39327-mc-02 case).
* **Namespaced resources**: SSRR in the resource’s namespace (same as before). Kubernetes treats listed rules as authoritative there.

Also fixed classifying ManagedCluster by `kind`, not by whether the payload happens to carry `metadata.namespace`.

Tests cover the Kevin scenarios + the evaluationError path -> all passing locally.

Please check the proposal at https://github.com/stolostron/console/pull/6648/changes/fe811a867585c6aacba5f308c246ef8edae29d59

Honestly, I have the impression we’re trying to use a sledgehammer to crack a nut. SSRR was never meant to drive authz decisions, and we’re layering rules on top of it because the original O(N) SSAR path doesn’t scale. This compromise keeps the performance win for restricted users while SSAR remains the source of truth whenever SSRR would actually grant access on cluster-scoped kinds.

Ginxo: FYI performance test report against latest changes https://redhat.atlassian.net/browse/ACM-39327?focusedCommentId=18071023

KevinFCormier: Ah, I completely discounted the "negative path"... if the SSRR shows no access for a cluster-scoped resource, and it is not incomplete, then we are sure the user has no access. Thanks for fixing.

KevinFCormier (approved):

backend/src/routes/eventsAccess.ts
```diff
@@ -0,0 +1,322 @@
...
+  'Infrastructure',
+  'ManagedCluster',
+  'ManagedClusterSet',
+  'ManagedClusterSetBinding',
+  'MultiClusterEngine',
+  'Namespace',
+  'Placement',
+  'PlacementDecision',
+  'Search',
+  'StorageClass',
```
KevinFCormier: @Ginxo Looking at the CodeRabbit review for the main-branch PR, I see it noticed that some of these are NOT cluster-scoped, and I noticed some additional ones.

- ClusterCurator
- DiscoveredCluster
- DiscoveryConfig
- ManagedClusterSetBinding
- Placement
- PlacementDecision
- Search

Also, ClusterExtension is missing from the list.

I think this will be error-prone to maintain because it is separate from the list of watches. Can I suggest that you:
- Add a `clusterScoped` optional boolean to the `IWatchOptions` interface
- Go through the current definitions and mark the correct ones as cluster-scoped
- Calculate the list of cluster-scoped kinds from `definitions`... eg. `const CLUSTER_SCOPED_KINDS = definitionsfilter((definition) => definition.clusterScoped).map((definition) => definition.kind)`

KevinFCormier (changes requested):
I discovered a new issue while reviewing the `main` PR. Please see inline comments.

Ginxo:
@KevinFCormier your cluster scope feedback is in on both PRs.

eventsCache.ts / eventsAccess.ts are the same on main and release-2.13. Only events.ts differs by branch (main keeps compression/meta/SSE optimizations).

Changes: 
* clusterScoped on watch definitions (no drifting hardcoded list)
* namespaced kinds no longer treated as cluster-scoped
* cluster-scoped grants still SSAR-confirmed
* SSAR keys include API group
* failed SSRR no longer cached as incomplete.

backend/src/routes/events.ts
```diff
@@ -221,8 +239,13 @@ const definitions: IWatchOptions[] = [
   },
 ]
 
+configureClusterScopedKinds(
+  definitions.filter((definition) => definition.clusterScoped).map((definition) => definition.kind)
+)
+
```
KevinFCormier, line 245: This seems overly complicated. I don't follow why in `eventsAccess.ts` you have the clusterScopedKinds defined as a `let`, and then you are depending on the side effect of loading this file to call the `configureClusterScopedKinds` function. And in the tests, you are testing with a subset of cluster-scoped kinds. Wouldn't it be more straightforward to do something like this?
```suggestion
export const CLUSTER_SCOPED_KINDS = new Set(definitions.filter((definition) => definition.clusterScoped).map((definition) => definition.kind))
)
```

If you're trying to avoid a circular dependency between events.ts and eventsAccess.ts, maybe we just need to move the definitions to their own file?

Ginxo: what do you think about https://github.com/stolostron/console/pull/6648/commits/0a66bdb8a07e5bfff7433dbc225b09837dcb61fe
as soon as we agree on this I will forward/backport it to the rest of versions (2.14-5.1)

KevinFCormier (approved):

PR #6658, opened 2026-08-10
FCN-641: Cluster creation function

nitin-dhevar:
# 📝 Summary
FCN-641: Cluster creation function

**Ticket Summary (Title):**  

Implementing the actual cluster-creation logic behind that callback: parsing the YAML into discrete Kubernetes resources, provisioning any prerequisite secrets, applying the resources to the ACM hub in the correct order, handling partial-failure rollback, and surfacing success/error state back to the wizard.

**Ticket Link:**  

https://redhat.atlassian.net/browse/FCN-641

**Type of Change:**  

- [ ] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General

#### If Feature

#### If Bugfix

---

## Summary by CodeRabbit

* **New Features**
  * Added ROSA HCP cluster creation through the managed-cluster wizard.
  * Added form-based YAML generation and validation for ROSA HCP resources.
  * Added support for networking, access, roles, OIDC, machine pools, encryption, and credentials.
  * Automatically derives availability zones and role references from selected infrastructure.
  * Successful creation opens the new cluster’s details page.
* **Bug Fixes**
  * Added clear submission errors, cleanup after failed creation, and error reset behavior.
* **Tests**
  * Added coverage for resource generation, validation, creation, rollback, and wizard workflows.


frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/RosaHCPWrapper.tsx
```diff
@@ -246,9 +243,19 @@ export const RosaHCPWrapper = () => {
         wizardData={wizardData}
         resourceGenerator={resourceGenerator}
         title=""
-        onCancel={() => console.log('CANCELLED')}
+        product="acm"
+        config={{ hiddenSteps: [STEP_IDS.CLUSTER_WIDE_PROXY, STEP_IDS.CLUSTER_UPDATES] }}
+        onSubmitError={submitError}
+        onBackToReviewStep={() => setSubmitError(false)}
+        onCancel={() => navigate(NavigationPath.managedClusters)}
```
davidaznaur, line 250: for now the only thing so far that I can see straight away. If we navigate like this away, I assume we unmount wizard. The state of the wizard is not preserved right?

KevinFCormier (commented):
Hey @nitin-dhevar @davidaznaur ,

I ran this through an AI code review and it picked up on a few things. I have filtered this down just to what I think would be good to fix. I am fine to merge as-is though and you can address these in follow-up bugs as you see fit.

1. create-rosa-hcp-cluster.ts ~L88-93 — Rollback can delete a pre-existing namespace.
The Fixing namespace issue during rollback commit (3d2d0c58d) now unconditionally pushes the namespace to createdResources right after ignoreAlreadyExists(createProject(...)). When createProject returns 409 (namespace already existed), a later failure will rollback() → delete a namespace this invocation did not create, destroying whatever else lives in it. CodeRabbit's thread on this is marked resolved but the code still pushes unconditionally.
Suggestion: have the namespace step report whether it actually created the namespace (e.g. ignoreAlreadyExists returns a boolean), and push to createdResources only on real creation. Add a regression test: namespace 409 → secret 500 → assert no namespace delete.

2. create-rosa-hcp-cluster.ts L106-108 — Not retry-safe: creds Secret create rethrows on 409.
${clusterName}-creds is created with a plain createResource(...).promise (no conflict handling). If a prior failed attempt left the Secret behind (rollback is best-effort via .catch(noop)), a resubmit throws already exists and stops before reconciling dependents. The PR description itself notes retries fail with "already exists".
Suggestion: define an idempotent policy for the creds Secret and per-cluster resources (replace/reconcile rather than blindly ignore, since a stale Secret carries stale credentials). Add a test for "retained creds Secret after failed attempt → successful retry".

3. i18n — user-facing error strings are not translated.
throw new Error('Unable to determine the cluster name from the generated resources.') (L79-80) and new Error('Failed to create the ROSA HCP cluster.') (L124) surface directly to the user via setSubmitError(err.message) → onSubmitError in RosaHCPWrapper.tsx (https://github.com/stolostron/console/blob/FCN-641-cluster-creation/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/components/rosahcp/RosaHCPWrapper.tsx). The wrapper translates its own fallback (t('An unexpected error occurred.')) but not these. ESLint won't catch it because they're in a .ts lib, not JSX.
Suggestion: pass a t function into createRosaHcpCluster, or map known failures to translated messages in the wrapper's catch instead of echoing err.message.

4. Late 422 when generated fields are absent (enrichFormValues.ts / template L34-61).
deriveAvailabilityZones and deriveRolesRef can return undefined; the .hbs then omits availabilityZones / rolesRef, and the ROSA HCP admission webhook rejects both with 422. With rollback now in place this is recoverable, but it's a poor UX (failure only after submit).
Suggestion: require non-empty AZs and a complete rolesRef in validateYaml (or pre-submit), returning a form validation error rather than shipping an incomplete manifest.

5. buildCredentialsSecret — Buffer.from(...).toString('ascii'). Use 'utf-8' to avoid corrupting any non-ASCII bytes in decoded credentials.

PR #6671, opened 2026-08-10
ACM-40502 fix(backend): use GET /api with body drain instead of GET /apis [release-2.14]

Randy424:
## Summary

Backport of the memory leak fix to `release-2.14`.

`isAuthenticated()` in `backend/src/lib/token.ts` was calling `GET /apis` on every authenticated request and never consuming the response body on the success path. This left sockets stuck outside the keepAlive pool, causing unbounded native (external) memory growth in `console-chart-console-v2` pods.

Reproduced at 5,899 MB peak RSS (+5,512 MB above baseline) with 615 stuck sockets under 150 concurrent connections. V8 heap oscillated normally throughout — the leak is entirely in native/external memory outside the GC's reach.

## Changes

- `isAuthenticated()` now uses `GET /api` (~200 bytes) instead of `GET /apis` (35+ KB, scales with CRDs)
- Response body is drained via `response.body?.on('error', () => undefined).resume()` so the socket returns to the keepAlive pool immediately
- Return type changed from `Promise<Response>` to `Promise<number>` — callers no longer hold a Response reference
- `authenticated.ts` updated to consume numeric status
- All route test auth mocks updated: `.get('/apis')` → `.get('/api')`

## Related

- ACM-40502 (this backport)
- ACM-38826 (parent — release-2.15, 2.16, 2.17 backports)
- Customer case: 04433290

## Test plan

- [ ] All 84 affected backend route tests pass locally
- [ ] `ci/prow/check` and `ci/prow/unit-tests-sonarcloud` pass in prow

### DIFF
--- a/backend/src/lib/authenticated.ts
+++ b/backend/src/lib/authenticated.ts
@@ -7,9 +7,8 @@ export function authenticated(req: Http2ServerRequest, res: Http2ServerResponse)
   const token = getToken(req)
   if (!token) return unauthorized(req, res)
   isAuthenticated(token)
-    .then((response) => {
-      res.writeHead(response.status).end()
-      void response.blob()
+    .then((status) => {
+      res.writeHead(status).end()
     })
     .catch(catchInternalServerError(res))
 }
--- a/backend/src/lib/token.ts
+++ b/backend/src/lib/token.ts
@@ -26,26 +26,30 @@ export function getToken(req: Http2ServerRequest): string | undefined {
   return token
 }
 
-export async function isAuthenticated(token: string) {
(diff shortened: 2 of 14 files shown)


KevinFCormier (commented):
@Randy424 Why is this PR missing the changes to `liveness.ts` (and corresponding `liveness.test.ts` updates)?

https://github.com/stolostron/console/pull/6602/changes#diff-c10b15c7d46ec88d05122d5c044a829ff46cac97fdd454d9adb7bd52692e8334

Randy424:
@KevinFCormier...Seems to also be the case for 2.14, 2.15, and 2.16. I don't recall why yet, but am looking into it...

PR #6760, opened 2026-08-24
ACM-38197-guard-against-flapping-resources

jeswanke:
# 📝 Summary

A policy resource is prevented from flapping and this alert is shown in console:

CleanShot 2026-08-24 at 13 49 35@2x

thresh hold parameter
export const FLAP_THRESHOLD = 5 // **N**: modifications that trigger flapping detection
export const FLAP_WINDOW_MS = 5 * 1000 // **M:** sliding window for counting modifications
export const FLAP_COOLDOWN_MS = 60 * 1000 // **P**: min interval between browser updates while flapping
export const FLAP_THROTTLE_KINDS = ['Policy'] // kinds subject to flapping detection

where:
If the same resource kind/name/namespace is modified more then **N** times in **M** seconds, only one event is sent per **P** minutes

to test:
TEST_THROTTLING=true npm run plugins

**Ticket Summary (Title):**  
ACM-38197 Guard Against Watched Kube Resource Flapping--Especially Policies

**Ticket Link:**  
https://redhat.atlassian.net/browse/ACM-38197

**Type of Change:**  

- [x] 🐞 Bug Fix  
- [ ] ✨ Feature  
- [ ] 🔧 Refactor
- [ ] 💸 Tech Debt
- [ ] 🧪 Test-related  
- [ ] 📄 Docs

---

### General
#### If Feature

#### If Bugfix

---

## Summary by CodeRabbit

- **New Features**
  - Added a warning clock indicator and tooltip for policies whose remediation actions are temporarily throttled.
  - Throttling status is reflected consistently for policies affected through propagated policy relationships.
  - Added support for configuring policy flapping and throttling timing thresholds through environment settings.

- **Bug Fixes**
  - Improved flapping detection so remediation throttling remains active through the appropriate cooldown period.
  - Prevented throttled updates from producing unnecessary cache or event activity.
  - Restored normal policy status and caching after throttling recovery.


frontend/src/components/LoadData.tsx
```diff
@@ -704,7 +718,35 @@ export function LoadData(props: { children?: ReactNode }) {
 
   const children = useMemo(() => <Fragment>{props.children}</Fragment>, [props.children])
 
-  return children
+  return (
```
KevinFCormier: This component cannot be used to display content, because it is rendered as a context provider that will be present in other parts of the OpenShift console that are not ACM.

Image

backend/src/routes/events.ts
```diff
@@ -35,10 +35,179 @@ export interface SettingsEvent {
...
+async function notifyFlapping(entry: FlapTrackerEntry): Promise<void> {
+  const message = formatFlappingMessage(entry.kind, entry.namespace, entry.name)
+  logger.warn({ msg: message, kind: entry.kind, namespace: entry.namespace, name: entry.name })
+  if (entry.flappingEventID) {
+    ServerSideEvents.removeEvent(entry.flappingEventID)
+  }
+  entry.flappingEventID = await ServerSideEvents.pushEvent({
+    data: {
+      type: 'FLAPPING',
+      message,
```
KevinFCormier: Rather than sending an untranslated message to the frontend, we should just include all the information needed here for the frontend to construct a translated message. We already have kind, namespace, and name... will just need some of the flapping parameters.

backend/src/routes/events.ts
```diff
@@ -35,10 +35,179 @@ export interface SettingsEvent {
...
+
+export function resourceFlapKey(
+  resource: Pick<IResource, 'kind'> & { metadata?: { namespace?: string; name?: string } }
+) {
+  return `${resource.kind}/${resource.metadata?.namespace ?? ''}/${resource.metadata?.name ?? ''}`
+}
+
+export function formatFlappingMessage(kind: string, namespace: string, name: string): string {
+  const timesPerMinute = Math.max(1, Math.round(60_000 / FLAP_COOLDOWN_MS))
+  return `${kind} ${namespace} ${name} is flapping, verify this resource is configured correctly. Until corrected, this resource will not update in the UI more then ${timesPerMinute} times per minute`
```
KevinFCormier: I don't think "flapping" will be universally understood. How about:

${kind} ${name} in namespace {$namespace} has been modified more than ${updated} times in the last ${threshold} minutes. Verify this resource is configured correctly. Updates are being limited to ${timesPerMinute} times per minute.

backend/src/routes/events.ts
```diff
@@ -35,10 +35,179 @@ export interface SettingsEvent {
...
+    const resource: IResource = {
+      kind: 'Policy',
+      apiVersion: 'policy.open-cluster-management.io/v1',
+      metadata: {
+        name: 'test-flapping-policy',
+        namespace: 'default',
+        uid: 'test-flapping-policy-uid',
+        resourceVersion: String(revision),
+      },
+    }
```
KevinFCormier, line 167: This test policy breaks the UI because it is incomplete. The UI expects spec.disabled to present.

KevinFCormier (changes requested):
Neat idea. I'm not sure how much impact it will have on the backend, but it will be good to have customers be aware when policies are misconfigured.

The flapTracker is another cache that can grow unbounded as new policies are created. Do you think we should clear the entry when a policy is deleted? I guess it depends if there are types of flapping where resources are continually created and deleted.

KevinFCormier (changes requested):
Hey @jeswanke this is getting closer, but I have some concerns and I've converted the Jira Bug to be a Story, because I think we'll need QE validation on this.

- I don't think the alerts look good in the current placement with their rounded corners right up against square borders.
- If there are multiple resources flapping, which is quite likely...
  - The alerts could fill the screen
  - The user has to dismiss each of them individually
  - The exact same message except for resource name is repeated over and over 
- There are plural issues in the string. Would need to use nested strings like the hostCount/infraEnvCount to have multiple plurals.
- Alert dismissal should probably be stored in local storage

So I think the alerts need some desigin work. Might need a consultation with @imjoyjean to assess whether there is a good way to display alerts like this within ACM pages, or if the console backend should just generate an OCP alert. Or maybe it would make more sense to only do this for policies and mark the policies themselves?

Image

KevinFCormier (changes requested):
@jeswanke I think this is a better design.

We might need a bit of tuning though:

- New policies are immediately detected as flapping. Probably because the status gets computed. We don't want this warning to show up under the normal course of events.
- I created 2 conflicting policies, `kevin-bar` and `kevin-baz` that fight over a ConfigMap. (See sample YAML below.) These are not consistently shown as flapping. Might need a bit more research into how often these would be checked and whether that is adjustable. But I think we should at least work well with the defaults.

```
apiVersion: policy.open-cluster-management.io/v1
kind: Policy
metadata:
  name: kevin-baz
  namespace: default
  annotations:
    policy.open-cluster-management.io/standards: NIST SP 800-53
    policy.open-cluster-management.io/categories: CM Configuration Management
    policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
spec:
  remediationAction: enforce    # inform = audit only; enforce = auto-remediate
  disabled: false
  policy-templates:
    - objectDefinition:
        apiVersion: policy.open-cluster-management.io/v1
        kind: ConfigurationPolicy
        metadata:
          name: kevin-baz
        spec:
          remediationAction: enforce
          severity: low
          namespaceSelector:
            include:
              - default
          object-templates:
            - complianceType: musthave
              objectDefinition:
                apiVersion: v1
                kind: ConfigMap
                metadata:
                  name: kevin-foo
                  namespace: default
                data:
                  foo: baz
```

backend/src/routes/events.ts
```diff
@@ -35,10 +35,226 @@ export interface SettingsEvent {
...
+    kind: resource.kind,
+    namespace: resource.metadata?.namespace ?? '',
+    name: resource.metadata?.name ?? '',
+  }
+}
+
+export function formatFlappingMessage(kind: string, namespace: string, name: string): string {
+  const windowMinutes = Math.max(1, Math.round(FLAP_WINDOW_MS / 60_000))
+  const timesPerMinute = Math.max(1, Math.round(60_000 / FLAP_COOLDOWN_MS))
+  return `${kind} ${name} in namespace ${namespace} has been modified more than ${FLAP_THRESHOLD} times in the last ${windowMinutes} minutes. Verify this resource is configured correctly. Updates are being limited to ${timesPerMinute} times per minute.`
```
JustinKuli, line 129: I have a nitpick with the wording here: "Updates are being limited" might make users think the `evaluationInterval` (or something similar) has been automatically set on these Policies in order to slow them down - but really it's just that *updates in the UI* are being limited, right?

Maybe something like `The state of the ${kind} will only be visually updated here ${timesPerMinute} times per minute` would be more clear?

JustinKuli (commented):
This is a nice idea and overall I think it's looking pretty good!

There are a few situations where this kind of thing will occur: as mentioned, if two Policies incompatibly manage the same resource, they will fight with each other; or similarly if another controller on the cluster manages the same object as a policy, they may conflict and both repeatedly update it. There used to be cases where this would just happen with one Policy - for example if some default fields were dropped by the API server, the Policy would keep noticing them as missing and say it was updating them - but I think/hope we've fixed all of those bugs by now. Just FYI.

jeswanke:
@KevinFCormier , ok, logic is more robust:
1. accounts for a settling time on new policies
2. resetting throttle is now done wth a poll
3. when if happens it adds this to the policy in the policy list
CleanShot 2026-09-16 at 09 57 37@2x
4. and now there is a card in the overview that summarizes the affected policy with more explanation:
CleanShot 2026-09-16 at 09 57 55@2x

backend/src/resources/resource.ts
```diff
@@ -43,6 +43,7 @@ interface OwnerReference {
 export interface IResource {
   kind: string
   apiVersion: string
+  throttled?: boolean
```
KevinFCormier, line 46: @jeswanke I don't like this approach. I don't think we should augment the resource with metadata _about_ the resource.

This might be somewhat related to an issue I'm seeing. If I enable both fernanod and montoya, or both kevin-foo-bar and kevin-foo-baz, I see they get detected as flapping.

But, if I disable one of them, it is immediately removed from the **Policy conflicts** list, but the other one remains indefinitely. I think this is because it depends on the resource being updated for it to be considered not flapping? But it might never be updated again. There needs to be an expiration on throttling.

frontend/public/locales/en/translation.json
```diff
@@ -2626,6 +2627,7 @@
   "policy.modal.warning.pruneParameter": "Some policies have the Prune parameter set.",
   "policy.modal.warning.pruneParameter.deleteMessage": "Deleting this policy might delete some related objects on the managed cluster(s).",
   "policy.modal.warning.pruneParameter.disableMessage": "Disabling this policy might delete some related objects on the managed cluster(s).",
+  "policy.overview.conflicts.tooltip": "The policies in this list are changing their compliances too rapidly. This can happen if:\n\n• A user policy may have a defect.\n\n• Two or more policies are trying to enforce different values for the same resource.\n\nUntil this is resolved, these policy updates will be throttled to once a minute.",
```
KevinFCormier: A few things here:
- The first bullet should say "A user policy has a defect." (So that I can read each bullet with "This can happen if" prepended to it)
- Formatting of messages should use a <Trans> component and tags that will be replaced
- Might want to check with @JustinKuli if the wording of the last sentence sounds specific enough to the UI

jeswanke: ok:
"A user policy has a defect."
"Until this is resolved, these policies **will only be visually updated here** once per minute." which is @JustinKuli suggestion.

backend/src/resources/resource.ts
KevinFCormier: @jeswanke any thoughts on this? I guess having it with the data is very convenient in the frontend.

Today I haven't been able to reproduce the issue I had with a policy never recovering from being throttled. It does still take awhile, but it does recover.

frontend/src/routes/Governance/overview/Overview.tsx
```diff
@@ -396,6 +409,92 @@ function ClustersCard() {
   )
 }
 
+const policyConflictsPopoverBody = (t: TFunction) => (
+  <span style={{ whiteSpace: 'pre-wrap' }}>
+    <Trans t={t} i18nKey="policy.overview.conflicts.tooltip" />
+  </span>
+)
```
KevinFCormier: My point about using the `<Trans>` component was that I don't think we should use pre-wrap and have newlines and bullet characters in the translation value.

KevinFCormier:
@jeswanke please check Sonar issues.

jeswanke:
@KevinFCormier ok Sonar is happy

PR #6849, opened 2026-09-17
ACM-45223 - ACM-44885: Fix non-admin SSE OOM under large inventory for Go

Ginxo:
# 📝 Summary

Backports [ACM-44885](https://redhat.atlassian.net/browse/ACM-44885) / [#6638](https://github.com/stolostron/console/pull/6638) to the **Go** `GET /events` path as part of [ACM-42568](https://redhat.atlassian.net/browse/ACM-42568) / [ACM-45223](https://redhat.atlassian.net/browse/ACM-45223).

**Depends on:** [#6842](https://github.com/stolostron/console/pull/6842) (`go/migration`). This PR should merge into `go/migration`, not `main` directly.

**Type of Change:** 🐞 Bug Fix

---

## Problem

Restricted (non-admin) users could drive the console backend into high memory use and CPU saturation on large inventories. After cluster-scoped `list` SSAR was denied, the Go hub fell through to **O(N) per-object SelfSubjectAccessReviews** (`list` namespaced → `get`) for every cached resource in the SSE snapshot/live stream.

Node fixed the same class of bug in #6638 with two ideas:

1. **RBAC short-circuit** — one `SelfSubjectRulesReview` per token+namespace, then local `deny-all` / `allow-all` / `allow-names` decisions (with SSAR confirmation for cluster-scoped kinds).
2. **Filter before inflate** — do not decompress cached resource blobs until RBAC allows the event.

Only **(1)** is ported here. That is the change that bounds non-admin inventory filtering and prevents the OOM.

---

## What changed (Go)

- **`SSARAccess.canSee`**: cluster `list` SSAR → `SelfSubjectRulesReview` → local kind decision; SSAR `get` only for incomplete fallback or cluster-scoped confirmation (RoleBinding in `default` cannot impersonate `ManagedCluster` access).
- **Caches**: hashed token keys, SSRR per namespace, kind-access per `namespace+group+resource`, per-token SSAR cap (2000), 60s TTL.
- **`WatchSpec.ClusterScoped`**: same cluster-scoped kinds as Node `CLUSTER_SCOPED_KINDS` (`ManagedCluster`, `Namespace`, `StorageClass`, …). Scope is **not** inferred from `metadata.namespace`.
- **Tests**: Go parity with Node `eventsAccess.test.ts` (e.g. 500 `ManagedCluster` → 1 SSRR, 0 SSAR `get`).
- **Cleanup**: remove orphaned Node TS from the #6638 merge (`backend/src/routes/events*.ts` and tests) — not executed by the Go backend.
- **Docs**: `backend/AGENTS.md`, `docs/ARCHITECTURE.md`.

---

## Why no zlib resource cache or per-client inflate

The Node backend kept SSE resources in memory as **dictionary-compressed zlib blobs** (`deflateResource` / `inflateEvent`). Each connected client could force **decompression to full JSON** before RBAC filtering (the pre-#6638 OOM path). #6638 added lightweight `meta` and **filter-before-inflate** so denied events never expanded.

The Go backend has a **different architecture**; those Node mechanisms are not needed for parity:

| Node (#6638) | Go (this PR) |
|--------------|--------------|
| In-process `resourceCache` of zlib `Buffer`s | **client-go informer cache** — one `unstructured` copy per resource, shared across clients |
| RBAC needs `meta` or inflate to read identity | Identity from `Event.Object` / `GVR` without a decompress step |
| `inflateEvent` per client before send | **`writeFiltered` → `Allow` → `json.Marshal` only if allowed** — denied events are never serialized to the wire |
| zlib on stored resources | **gzip/deflate on the SSE HTTP stream** (`internal/events/hub/encode.go`) — compression in transit, not in the event store |

So:

- There is **no inflate step** because nothing is stored as a compressed blob between cache and client.
- There is **no per-resource zlib cache** because informers already hold the canonical object; duplicating Node’s dictionary+zlib layer would add CPU and complexity without addressing the real Go gap (RBAC O(N)).
- The **security and performance invariant** we preserve is: *do not do expensive per-object authorization work for users who cannot list cluster-scoped resources* — achieved here with SSRR, not with compression.

---

## Test plan

- [x] `go test ./...` (hub SSRR tests including 500-cluster scale case)
- [x] golangci-lint on backend
- [ ] Manual on hub: kubeadmin vs `none` user on Inventory — SSE completes, backend RSS/CPU stay bounded (no `MOCK_CLUSTERS` in Go; regression gate is unit tests above)

---

### General

#### If Bugfix

- [x] Root cause and fix summary are documented in the ticket (for future reference / errata)
- [ ] Fix tested thoroughly and resolves the issue (pending hub validation)
- [x] Test(s) added to prevent regression

---

### 🗒️ Notes for Reviewers

- Compare behavior with Node `eventsAccess.ts` from #6638; cluster-scoped SSAR confirmation after non-`deny-all` SSRR is intentional.
- `DELETED` events remain unfiltered (bug-compat with Node; documented in `backend/AGENTS.md`).

[ACM-44885]: https://redhat.atlassian.net/browse/ACM-44885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACM-42568]: https://redhat.atlassian.net/browse/ACM-42568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACM-45223]: https://redhat.atlassian.net/browse/ACM-45223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ginxo:
closed in favor of https://github.com/Ginxo/console/pull/73

